PR #6, opened 2022-10-02
Iswap state

michaelbest55:
Hi there!
I discussed with Rick that we should keep track of the iSWAP state in the star class before calling methods that need the iSWAP to be parked.

A couple of questions/comments:
- I guess the `need_iswap_parked` wrapper should not be where it is, would it make sense to add it to STAR_helpers.py module?
- I wasn't sure if I should be using the `request_iswap_in_parking_position` or the `park_iswap function`.
- When setting up the environment on a Linux machine, it's not possible to install pywin==32, I think you would want a `platform_system` in setup.py?

Please let me know what you think, and thanks for setting this project up! I would love to be a part of it if necessary

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR.py
@@ -6,10 +6,11 @@ This file defines interfaces for all supported Hamilton liquid handling robots.
 from abc import ABCMeta, abstractmethod
 import datetime
 import enum
+import functools
 import logging
 import re
 import typing
-from typing import Union, List, Optional
+from typing import Callable, Union, List, Optional
 
 from pylabrobot import utils
 from pylabrobot.liquid_handling.resources import (
@@ -385,6 +386,7 @@ class STAR(HamiltonLiquidHandler):
     self.dev: Optional[usb.core.Device] = None
     self._tip_types: dict[str, int] = {}
     self.num_channels = num_channels
+    self._iswap_parked: bool = False
 
     self.read_endpoint: Optional[usb.core.Endpoint] = None
     self.write_endpoint: Optional[usb.core.Endpoint] = None
@@ -461,6 +463,7 @@ class STAR(HamiltonLiquidHandler):
       self.initialize_iswap()
 
     self.park_iswap()
+    self._iswap_parked = True
 
   def stop(self):
     if self.dev is None:
@@ -468,6 +471,27 @@ class STAR(HamiltonLiquidHandler):
     logging.warning("Closing connection to USB device.")
     usb.util.dispose_resources(self.dev)
     self.dev = None
+
+
+  def need_iswap_parked(self, func: Callable):
+    """Ensure that the iSWAP is in parked position before running command.
+
+    If the iSWAP is not parked, it get's parked before running the command.
+    """
+
+    @functools.wraps(func)
+    def wrapper(self, *args, **kwargs):
+      if self._iswap_parked:
+        result = func(*args, **kwargs)
+        self._iswap_parked = False
+      else:
+        self.park_iswap()
+        self._iswap_parked = True
+        result = func(*args, **kwargs)
+        self._iswap_parked = False
+
+      return result
+    return wrapper
 
   # ============== Tip Types ==============
 
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR.py
@@ -44,6 +44,26 @@ except ImportError:
   USE_USB = False
 
 
+def need_iswap_parked(method: Callable):
+  """Ensure that the iSWAP is in parked position before running command.
+
+  If the iSWAP is not parked, it get's parked before running the command.
+  """
+
+  @functools.wraps(method)
+  def wrapper(self, *args, **kwargs):
+    if self.iswap_parked:
+      result = method(self, *args, **kwargs)
+      self.iswap_parked = False
+    else:
+      self.park_iswap()
+      self.iswap_parked = True
+      result = method(self, *args, **kwargs)
+      self.iswap_parked = False
+
+    return result
+  return wrapper
+
 class HamiltonLiquidHandler(LiquidHandlerBackend, metaclass=ABCMeta):
   """
   Abstract base class for Hamilton liquid handling robot backends.
@@ -391,6 +411,14 @@ class STAR(HamiltonLiquidHandler):
     self.read_endpoint: Optional[usb.core.Endpoint] = None
     self.write_endpoint: Optional[usb.core.Endpoint] = None
 
+  @property
+  def iswap_parked(self):
+    return self._iswap_parked
+
+  @iswap_parked.setter
+  def iswap_parked(self, parked: bool):
+    self._iswap_parked = parked
+
   def setup(self):
     """ setup
 
@@ -471,27 +499,6 @@ class STAR(HamiltonLiquidHandler):
     logging.warning("Closing connection to USB device.")
     usb.util.dispose_resources(self.dev)
     self.dev = None
-
-
-  def need_iswap_parked(self, func: Callable):
-    """Ensure that the iSWAP is in parked position before running command.
-
-    If the iSWAP is not parked, it get's parked before running the command.
-    """
-
-    @functools.wraps(func)
-    def wrapper(self, *args, **kwargs):
-      if self._iswap_parked:
-        result = func(*args, **kwargs)
-        self._iswap_parked = False
-      else:
-        self.park_iswap()
-        self._iswap_parked = True
-        result = func(*args, **kwargs)
-        self._iswap_parked = False
-
-      return result
-    return wrapper
 
   # ============== Tip Types ==============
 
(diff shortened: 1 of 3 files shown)


rickwierenga:
Hi Michael, thanks for your contribution!

> I guess the need_iswap_parked wrapper should not be where it is, would it make sense to add it to STAR_helpers.py module?

I think we could move it to be a method of STAR, like [`need_setup_finished`](https://github.com/pylabrobot/pylabrobot/blob/1b0eaeea94edd970fed1b207b88d1c57d3054fc6/pyhamilton/liquid_handling/liquid_handler.py#L353).

> I wasn't sure if I should be using the request_iswap_in_parking_position or the park_iswap function.

We should only have to use `request_iswap_in_parking_position` once, at setup. After that it should be sufficient to keep track of iSWAP state in the memory. `park_iswap` should be used by the decorator, if necessary. See my comments in the code review.

> When setting up the environment on a Linux machine, it's not possible to install pywin==32, I think you would want a platform_system in setup.py?

What do we need pywin for?

pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -43,6 +44,26 @@
...
+      self.iswap_parked = False
+    else:
+      self.park_iswap()
+      self.iswap_parked = True
+      result = method(self, *args, **kwargs)
+      self.iswap_parked = False
+
+    return result
+  return wrapper
+
```
rickwierenga, line 66: As mentioned in my comment we should move this to be a member of STAR.

pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -43,6 +44,26 @@
...
+  """Ensure that the iSWAP is in parked position before running command.
+
+  If the iSWAP is not parked, it get's parked before running the command.
+  """
+
+  @functools.wraps(method)
+  def wrapper(self, *args, **kwargs):
+    if self.iswap_parked:
+      result = method(self, *args, **kwargs)
+      self.iswap_parked = False
```
rickwierenga, line 57: Why would iswap become not parked after finishing a method that needs it parked?

Imo, `self.iswap_parked = False` should be set in the `get_plate`, `put_plate` and `move_plate_to_position` methods, which send the firmware commands to use the iSWAP.

pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -43,6 +44,26 @@
...
+    else:
+      self.park_iswap()
+      self.iswap_parked = True
```
rickwierenga, line 60: Similarly, `self.iswap_parked = True` should be set in `park_iswap`.

(Right now if the user would call `park_iswap()` before calling a `@need_iswap_parked` method, we would send the park command twice.)

pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -43,6 +44,26 @@
...
+      result = method(self, *args, **kwargs)
+      self.iswap_parked = False
```
rickwierenga, line 62: idem

pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -385,10 +406,19 @@ def __init__(self, num_channels: int = 8, read_timeout: int = 5, **kwargs):
...
     self.read_endpoint: Optional[usb.core.Endpoint] = None
     self.write_endpoint: Optional[usb.core.Endpoint] = None
 
+  @property
+  def iswap_parked(self):
+    return self._iswap_parked
+
+  @iswap_parked.setter
+  def iswap_parked(self, parked: bool):
+    self._iswap_parked = parked
```
rickwierenga, line 420: I don't really see value in having a setter over just writing to `_iswap_parked` directly, since we only write from within self. In fact, this gives the impression that the setter is publicly available, which is not the case (the state should only be handled indirectly, by STAR itself).

pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -461,6 +491,7 @@ def setup(self):
       self.initialize_iswap()
 
     self.park_iswap()
+    self._iswap_parked = True
```
rickwierenga, line 494: idem

pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -385,10 +406,19 @@ def __init__(self, num_channels: int = 8, read_timeout: int = 5, **kwargs):
...
     self._tip_types: dict[str, int] = {}
     self.num_channels = num_channels
+    self._iswap_parked: bool = False
 
     self.read_endpoint: Optional[usb.core.Endpoint] = None
     self.write_endpoint: Optional[usb.core.Endpoint] = None
 
+  @property
+  def iswap_parked(self):
+    return self._iswap_parked
```
rickwierenga, line 416: Nice!

Can we add a type indicator for the return type? (`-> bool`?)

rickwierenga (changes requested):
Some comments about a few minor changes to make the code more readable and stable.

michaelbest55:
> Hi Michael, thanks for your contribution!
> 
> > I guess the need_iswap_parked wrapper should not be where it is, would it make sense to add it to STAR_helpers.py module?
> 
> I think we could move it to be a method of STAR, like [`need_setup_finished`](https://github.com/pylabrobot/pylabrobot/blob/1b0eaeea94edd970fed1b207b88d1c57d3054fc6/pyhamilton/liquid_handling/liquid_handler.py#L353).
> 
> > I wasn't sure if I should be using the request_iswap_in_parking_position or the park_iswap function.
> 
> We should only have to use `request_iswap_in_parking_position` once, at setup. After that it should be sufficient to keep track of iSWAP state in the memory. `park_iswap` should be used by the decorator, if necessary. See my comments in the code review.
> 
> > When setting up the environment on a Linux machine, it's not possible to install pywin==32, I think you would want a platform_system in setup.py?
> 
> What do we need pywin for?

Pywin is set in the requirements file.
If I run `pip install -e '.[all]'` as described in the docs for development, I get the error
```
ERROR: Could not find a version that satisfies the requirement pywin32 (from pyhamilton->PyLabRobot==0.1.0) (from versions: none)
ERROR: No matching distribution found for pywin32 (from pyhamilton->PyLabRobot==0.1.0)
```

pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -385,10 +386,32 @@ def __init__(self, num_channels: int = 8, read_timeout: int = 5, **kwargs):
...
+    If the iSWAP is not parked, it get's parked before running the command.
+    """
+
+    @functools.wraps(method)
+    def wrapper(self, *args, **kwargs):
+      if self.iswap_parked:
+        result = method(self, *args, **kwargs) # pylint: disable=not-callable
+      else:
+        self.park_iswap()
+        result = method(self, *args, **kwargs) # pylint: disable=not-callable
```
rickwierenga, line 410: This can be further refactored as

```py
if not self.iswap_parked:
  self.park_iswap()

result = method(self, *args, **kwargs) # pylint: disable=not-callable
```

pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -385,10 +386,32 @@ def __init__(self, num_channels: int = 8, read_timeout: int = 5, **kwargs):
     self.dev: Optional[usb.core.Device] = None
     self._tip_types: dict[str, int] = {}
     self.num_channels = num_channels
+    self._iswap_parked: bool = False
 
     self.read_endpoint: Optional[usb.core.Endpoint] = None
     self.write_endpoint: Optional[usb.core.Endpoint] = None
 
+  @property
+  def iswap_parked(self)->bool:
```
rickwierenga, line 395: This is very nitpicky, and not even included in our current linting rules, but please add spaces around the arrow for consistent code style with the rest of the repo :)

pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -385,10 +386,32 @@ def __init__(self, num_channels: int = 8, read_timeout: int = 5, **kwargs):
     self.dev: Optional[usb.core.Device] = None
     self._tip_types: dict[str, int] = {}
     self.num_channels = num_channels
+    self._iswap_parked: bool = False
```
rickwierenga, line 389: Maybe start as `None` because it is underdetermined at initialization?

rickwierenga (changes requested):

rickwierenga:
> If I run pip install -e '.[all]' as described in the docs for development, I get the error

True. What do you think about adding a `dev` target that sets up all universal dependencies, and have the platform specific dependencies be optional, even for development? This would be a good second PR! (in general it is good to have self contained and single-topic PRs)

pylabrobot/liquid_handling/backends/hamilton/STAR.py
michaelbest55: addressed

PR #7, opened 2022-10-05
Docs installation update for Mac

yachty66:


rickwierenga:
The issue here was most likely using system Python.

Thanks for the help working through this!

reference https://forums.pylabrobot.org/t/how-to-use-hamilton-star-with-mac/503

PR #9, opened 2022-10-07
example dev setup

michaelbest55:
So here is what was suggested here: https://github.com/PyLabRobot/pylabrobot/pull/6#issuecomment-1264691856

I actually think a better way of doing this is to add `"sys_platform == "win32"'` to the install_requires in the pyhamilton library instead. Let me know what you think

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -64,7 +64,7 @@ Alternatively, you can install PyLabRobot from source. This is particularly usef
 ```bash
 git clone https://github.com/pylabrobot/pylabrobot.git
 cd pylabrobot
-pip install -e '.[all]'
+pip install -e '.[dev]'
 ```
 
 See [CONTRIBUTING.md](https://github.com/PyLabRobot/pylabrobot/blob/main/CONTRIBUTING.md) for specific instructions on testing, documentation and development.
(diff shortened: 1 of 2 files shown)


rickwierenga:
PyHamilton already has `pythonnet` and `pywin32` in install_requires. What are you proposing to add?

Since we do not have tests for PyHamilton in PyLabRobot (yet?), I don't think it should be required in the dev target, Windows or not.

Also, please add the new target to CONTRIBUTING.md. I think maybe it should replace `all` in the 'from source' section.

Happy to discuss.

michaelbest55:
I've removed the venus extras from the dev target. I've also replaced the mention of the all target in the installation docs for dev work.

Regarding pyhamilton, I'm suggesting the following https://github.com/dgretton/pyhamilton/blob/f23d4dd4fc4010db0e80bc56cfee44b6ae8f7119/setup.py#L18
should become
```
    install_requires=['requests', 'pythonnet', 'pywin32; sys_platform == "win32"'', 'pyserial']
```
This allows you to specify that you only want to install pywin if you are on a windows platform (which I would argue is what we would want).

rickwierenga:
Thanks! This looks great.

As for PyHamilton, since it only works on Windows, we don't care (in fact we prefer it) if it has Windows-only requirements.

PR #11, opened 2022-11-19
Fix tip length for high volume no filter tips

alex-spaero:
Hi all, thanks for this package! 

After testing, it seems like the current tip height for non-filtered high-volume tips was copied from the standard tips and is not correct, as it should be 95.1 (the same as the filtered high-volume tips). This PR updates the value to the correct length.


rickwierenga:
Good catch, thanks!

PR #12, opened 2022-11-19
Increase command timeout for STAR

alex-spaero:
Testing on the latest commit, it seems like the timeout for the STAR commands is too small. Right now, it will timeout during setup for example, or any other long running command like serial aspirations. The increased value here is enough to allow `setup` to complete successfully, but there should probably be a better heuristic here for commands that might take even longer than 100 s (which is definitely a possibility with low flow rates, high volumes, multiple moves etc).

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR.py
@@ -309,7 +309,7 @@ class HamiltonLiquidHandler(LiquidHandlerBackend, metaclass=ABCMeta):
     self,
     module: str,
     command: str,
-    timeout: int = 16,
+    timeout: int = 100,
     fmt: typing.Optional[str]=None,
     wait = True,
     **kwargs


rickwierenga:
Thanks for the PR! You are right that 16 seconds may not be enough in some cases.

However, I think 100 sec is too long as a base case, and would prefer we move to a dynamic version, as you mentioned. I know this is not an easy thing to implement, but I think for now we can be somewhat dynamic by using the `timeout` parameter in the calls to `send_command` where necessary.

Happy to discuss!

alex-spaero:
What is a timeout you find more reasonable?

There should probably be a way to set the timeout now if you know you will hit the default. Right now, there is no way to set the timeout via the LiquidHandling interface without some extra work adding a propagating an extra kwargs to every method with a `send_command` call in the backend. 

An easier solution might be to have the timeout be set as a parameter to the `__init__` for the `STAR` backend. Right now there is a `read_timeout` parameter, but it seems like this only affects the timeout per read attempt, not the overall timeout before an exception is raised. How would you feel about :
1. Removing the `timeout` parameter from `send_command` and instead having a `read_attempts` parameter, which will cause a fixed number of `read_timeout` interval attempts at reading.
2. Replacing the `read_timeout` in `STAR.__init__` and `timeout` in `send_command` with a single `timeout` in the `STAR.__init__` which is used for every command sent.

rickwierenga:
> 1. Removing the timeout parameter from send_command and instead having a read_attempts parameter, which will cause a fixed number of read_timeout interval attempts at reading.

I think that we should keep it as a parameter for flexibility, and as an improvement it should be propagated through `backend_kwargs`. I can do that extra work soon.

> 2. Replacing the read_timeout in STAR.__init__ and timeout in send_command with a single timeout in the STAR.__init__ which is used for every command sent.

I think we should have separate parameters, for the individual read attempts and the cumulative time to find the right command. (currently `read_timeout` applies to both, and exits at whichever point it expires first, which I agree is unclear.) I think the best way to represent the cumulative time is as a `max_read_attempts: int` param at initialization. I'm thinking this would also be the parameter that is exposed via `backend_kwargs` at the command level.

This way, we maximize flexibility of the API, and it best prepares us for the point where we add dynamic timeouts.

alex-spaero:
As a note, it takes the STAR I am working with about 62 seconds to execute the `DI` command as part of the `setup` procedure when there are tips attached that need to be dropped. This is with  8-channel pipettes, MPH and iSWAP arm.

alex-spaero:
Closing since the timeouts should be covered by https://github.com/PyLabRobot/pylabrobot/commit/2882cc2267ac59544701a8f15e085263af5ac724

rickwierenga:
Great!

PR #14, opened 2023-01-11
Add lid checker

wilsonh1:
Raises an error if aspirating or dispensing from a plate with a lid. Updated tests to account for the new check.


rickwierenga:
Looks great!

PR #16, opened 2023-01-26
Add missing await to super setup for STAR

alex-spaero:
The asyncio refactor looks great!

There is a missing await in STAR.setup, which is preventing initialization. This PR fixes that.


rickwierenga:
Great, thank you!

PR #17, opened 2023-01-26
Always eject tips in trash during STAR setup

alex-spaero:
Currently, on `setup` the STAR will eject tips in the waste position for the STARlet (which is in the middle of the deck). This PR updates the setup to use the `xw` parameter from the extended machine configuration to avoid hard-coding the trash location. This will have the robot discard tips in the correct spot for both STAR and STARlet robots.


rickwierenga:
Thanks for this too :)

PR #18, opened 2023-01-26
Fix error handling for asyncio

alex-spaero:
Currently, if an HamiltonFirmwareError is raised by the `parse_response` in `_continuously_read` (e.g. when the robot fails to pick up tips), the caller hangs, since the future it is awaiting is never set (the handling thread crashes).

This PR updates the error handling in ` _continuously_read` to set the exception for the `fut` so that it is raised the awaiting caller.


rickwierenga:
Good catch!

PR #19, opened 2023-01-31
Add missing `te` parameter to discard_tips (TR)

alex-spaero:
I ran into an issue with the robot hanging when running `LiquidHandler.return_tips` with channels other than the first few. Here is a simple example for reproducing that issue:

```python3
from pylabrobot.liquid_handling.backends import STAR
from pylabrobot.resources.hamilton import STARDeck
    TIP_CAR_480_A00,
    STF_L
)

backend = STAR()
lh = LiquidHandler(backend=backend, deck=STARDeck())

tip_car = TIP_CAR_480_A00(name='tip carrier')
tip_car[0] = STF_L(name='tips_01')
lh.deck.assign_child_resource(tip_car, rails=3)

tiprack = lh.get_resource("tips_01")
await lh.pick_up_tips(tiprack["A1":"D1"], use_channels=[3,4,5])
await lh.return_tips()
```

The last call to `return_tips` will result in the robot hanging and eventually timing out trying to read a response. On reinitialization, the robot sends an error message for the TR command (err01/31, faulty parameter). I think the issue is the the TR command should have a `te` parameter analogous to the `te` param in ` initialize_pipetting_channels`. When I add the param, the command works as expected.

Weirdly enough, `return_tips` works if I'm not specifying `use_channels`. Not sure why it does not require `te` when only using the first n channels.

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR.py
@@ -811,6 +811,7 @@ class STAR(HamiltonLiquidHandler):
         begin_tip_deposit_process=int((max_z + max_total_tip_length)*10),
         end_tip_deposit_process=int((max_z + max_tip_length)*10),
         minimum_traverse_height_at_beginning_of_a_command=2450,
+        z_position_at_end_of_a_command=2450,
         discarding_method=drop_method
       )
     except HamiltonFirmwareError as e:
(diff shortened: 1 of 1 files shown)


rickwierenga:
Thanks!

In our log files (>17GB), `te` is always 2450 for command `TR`. If you have seen any other values, please let me know so we can reverse engineer the formula. Otherwise, could you please pass 2450 to the call to `discard_tip` in `drop_tips`?

alex-spaero:
Done!

PR #20, opened 2023-02-03
Initial Tecan integration

wilsonh1:
Adds preliminary support for Tecan labware and liquid classes. Backend includes methods for limited aspirating and dispensing. Dimensions and fixed constants need further verification.


pylabrobot/liquid_handling/backends/tecan/EVO.py
```diff
@@ -0,0 +1,734 @@
...
+    """
+
+    x_positions, y_positions, z_positions = self._ops_to_fw_positions(ops, use_channels)
+
+    tecan_liquid_classes = [
+      get_liquid_class(
+        target_volume=op.volume,
+        liquid_class=op.liquid_class,
+        tip_type=op.tip.tip_type
+      ) if hasattr(op.tip, 'tip_type') else None for op in ops]
```
rickwierenga, line 207: `if isinstance(op.tip, TecanTip)`

pylabrobot/resources/tecan/tip_racks.py
```diff
@@ -0,0 +1,1522 @@
...
+    self.z_dispense = z_dispense
+    self.z_max = z_max
+    self.area = area
+
+
+def DiTi_100ul_Te_MO(name: str) -> TecanTipRack:
+  return TecanTipRack(
+    name=name,
+    size_x=128.0,
+    size_y=85.80000000000001,
```
rickwierenga, line 42: These are most likely floating point errors, so maybe round to 4 decimal places when making resources?

rickwierenga (changes requested):
Great work!

Many comments are very minor. Most important is maybe introducing `TecanResource` as a base class for Tecan labware, but as always I'm happy to discuss.

Do you want to keep working in this PR?

pylabrobot/liquid_handling/liquid_classes/abstract.py
```diff
@@ -43,9 +43,11 @@ def from_str(s: str) -> "LiquidClass":
       s = "Blood"
     elif s == "Glycerin80":
       s = "Glycerin 80%"
-    elif s == "SysFlWater":
+    elif s in {"Water", "SysFlWater"}:
       s = "Water" # TODO: ?
     elif s == "Plasma":
       s = "Plasma 100%"
+    elif s == "DMSO":
+      s = "DMSO"
```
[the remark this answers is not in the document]
wilsonh1: The abstract class seems general enough to leave as is for now. In particular, it already covers all of the default Tecan liquid classes.

wilsonh1:
Seems like the type checking failure is due to a new release of mypy.

pylabrobot/liquid_handling/liquid_classes/abstract.py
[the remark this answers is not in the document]
rickwierenga: How would we differentiate the different systems in the future?

rickwierenga:
https://github.com/PyLabRobot/pylabrobot/commit/6835d1b9c5324dcd2c343c12bec658543d637464 should fix that if you sync the fork

pylabrobot/liquid_handling/liquid_classes/abstract.py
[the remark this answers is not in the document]
wilsonh1: I moved some of the Tecan parsing to a new `from_str` method in `liquid_classes/tecan.py`. The old method could be moved to `liquid_classes/hamilton.py`. `abstract.py` could then be a system-agnostic list of generic liquids.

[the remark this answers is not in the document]
rickwierenga: Good idea.

One thing to consider is that `from_str` is currently a member of `TecanLiquidClass`, but does not really have anything to do with that class (and blurs the distinction with `LiquidClass`), so why not make it a top level method?

pylabrobot/liquid_handling/backends/tecan/EVO.py
```diff
@@ -0,0 +1,740 @@
...
+      resp: The response string to parse.
+
+    Returns:
+      A dictionary containing the parsed values.
+    """
+
+    data: List[int] = [int(x) for x in resp[3:-1].split(",")]
+    return {
+      "module": resp[1:3],
+      "ret": [0], # data[0] TODO: get error return code
```
rickwierenga, line 91: This was the byte thing, right?

pylabrobot/liquid_handling/liquid_classes/tecan.py
```diff
@@ -0,0 +1,2795 @@
...
+    """ Parses a Tecan liquid class name and creates a LiquidClass object. """
+
+    m = re.match(r"(\w+) free dispense$", s)
+    if m is None:
+      return None
+
+    lc = m.group(1)
+    if lc in {"Ethanol", "Serum"}:
+      lc += " 100%"
+    return LiquidClass(lc)
```
rickwierenga, line 132: See other comment above

rickwierenga (changes requested):

pylabrobot/liquid_handling/backends/tecan/EVO.py
wilsonh1: Yup. I'll capture an example error in Wireshark when I use the machine again.

luisvillaautomata:
This is amazing Wilson!

If you need help filling in commands, please let me know.

pylabrobot/liquid_handling/backends/tecan/EVO.py
```diff
@@ -226,55 +295,56 @@ async def aspirate(
     # moves such that first channel is over first location
     x, _ = self._first_valid(x_positions)
     y, yi = self._first_valid(y_positions)
+    print(y, yi)
```
rickwierenga, line 308: Think this is a debugging artifact :)

rickwierenga:
Great work! Is there any part of the setup that is not part of this code yet?

rickwierenga:
Can we include those in the deck initialization?

Are there any other things that should be done before [basic.ipynb](https://docs.pylabrobot.org/basic.html) runs as a standalone notebook?

wilsonh1:
A wash station is assigned on its own now. For now, the backend assumes that it is assigned to rail 1.

Setting fixed tips is a bit trickier. `lh` currently requires something like:
```
# standard fixed tips
tips = {channel: standard_fixed_tip() for channel in range(8)}
lh.update_head_state(tips)
```

rickwierenga:
Perhaps we can add a method like `request_initial_machine_state` to the backend spec? I'm not the biggest fan of requiring the code snippet above in every Tecan protocol. Happy to discuss.

wilsonh1:
I don't think the Tecan can report which channels have fixed tips mounted. It seems like this is configured and stored entirely by EVOware. A new method would still require the user to manually set tips.

rickwierenga:
All right. I think we should be ready to merge then.

PR #25, opened 2023-04-02
Add self-assignment and multiple-parent checks to Resource.assign_child_resource

naikymen:
Motivation: resources can be assigned to themseves and have multiple parents.

image

Which can be fixed by this PR.

### DIFF
--- a/pylabrobot/resources/resource.py
+++ b/pylabrobot/resources/resource.py
@@ -128,7 +128,9 @@ class Resource:
     """ Get the size of this resource in the z-direction. """
     return self._size_z
 
-  def assign_child_resource(self, resource: Resource, location: Optional[Coordinate]):
+  def assign_child_resource(self,
+                            resource: Resource, location: Optional[Coordinate],
+                            reassign: bool = True):
     """ Assign a child resource to this resource.
 
     Will use :meth:`~Resource.resource_assigned_callback` to notify the parent of the assignment,
@@ -137,7 +139,7 @@ class Resource:
     """
 
     # Check for unsupported resource assignment operations
-    self.check_assignment(resource)
+    self.check_assignment(resource, reassign)
 
     resource.parent = self
     resource.location = location
@@ -151,7 +153,7 @@ class Resource:
 
     self.children.append(resource)
 
-  def check_assignment(self, resource: Resource):
+  def check_assignment(self, resource: Resource, reassign: bool = True):
     """Check if the resource assignment produces unsupported or dangerous conflicts.
     """
     msgs = []
@@ -160,8 +162,13 @@ class Resource:
     if resource is self:
       msgs.append(f"Will not assign resource '{self.name}' to itself.")
 
-    # Check for multiple parents
-    if resource.parent is not None:
+    # Check for reassignment to the same parent
+    if resource.parent is not None and (resource.parent is self):
+      msgs.append(f"Will not reassign resource '{resource.name}' " +
+                  f"to the same parent: '{resource.parent.name}'.")
+
+    # Check for pre-existing parent
+    if resource.parent is not None and not reassign:
       msgs.append(f"Will not assign resource '{resource.name}' " +
                   f"that already has a parent: '{resource.parent.name}'.")
 
--- a/pylabrobot/resources/resource.py
+++ b/pylabrobot/resources/resource.py
@@ -156,7 +156,7 @@ class Resource:
       msgs.append(f"Will not assign resource '{self.name}' to itself.")
 
     # Check for multiple parents
-    if resource.parent:
+    if resource.parent is not None:
       msgs.append(f"Will not assign resource '{resource.name}' that already has a parent: '{resource.parent.name}'.")
 
     # TODO: write other checks, perhaps recursive or location checks.
--- a/pylabrobot/resources/resource.py
+++ b/pylabrobot/resources/resource.py
@@ -128,7 +128,7 @@ class Resource:
     """ Get the size of this resource in the z-direction. """
     return self._size_z
 
-  def assign_child_resource(self, resource: Resource, location: Optional[Coordinate]):
+  def assign_child_resource(self, resource: Resource):
     """ Assign a child resource to this resource.
 
     Will use :meth:`~Resource.resource_assigned_callback` to notify the parent of the assignment,
@@ -147,8 +147,10 @@ class Resource:
       raise e
 
     self.children.append(resource)
-    
-  def check_assignment(self, resource: Resource, location: Optional[Coordinate]):
+  
+  def check_assignment(self, resource: Resource):
+    """Check if the resource assignment produces unsupported or dangerous conflicts.
+    """
     msgs = []
 
     # Check for self assignment
@@ -157,7 +159,8 @@ class Resource:
 
     # Check for multiple parents
     if resource.parent is not None:
-      msgs.append(f"Will not assign resource '{resource.name}' that already has a parent: '{resource.parent.name}'.")
+      msgs.append(f"Will not assign resource '{resource.name}' " +
+                  f"that already has a parent: '{resource.parent.name}'.")
 
     # TODO: write other checks, perhaps recursive or location checks.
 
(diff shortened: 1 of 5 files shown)


pylabrobot/resources/resource.py
```diff
@@ -136,16 +136,34 @@ def assign_child_resource(self, resource: Resource, location: Optional[Coordinat
...
+    
+  def check_assignment(self, resource: Resource, location: Optional[Coordinate]):
+    msgs = []
+
+    # Check for self assignment
+    if resource is self:
+      msgs.append(f"Will not assign resource '{self.name}' to itself.")
+
+    # Check for multiple parents
+    if resource.parent:
```
rickwierenga, line 159: `if resource.parent is not None` is cleaner.

pylabrobot/resources/resource.py
```diff
@@ -136,16 +136,34 @@ def assign_child_resource(self, resource: Resource, location: Optional[Coordinat
...
       self.resource_assigned_callback(resource) # call callbacks first.
     except Exception as e:
       resource.parent = None
       resource.location = None
       raise e
 
     self.children.append(resource)
+    
+  def check_assignment(self, resource: Resource, location: Optional[Coordinate]):
+    msgs = []
```
rickwierenga, line 152: These errors are mutually exclusive (one would be raised before the other, so no two conditions should be true at the same time), so it may be better to raise the errors directly rather than putting them in a list.

pylabrobot/resources/resource.py
```diff
@@ -136,16 +136,34 @@ def assign_child_resource(self, resource: Resource, location: Optional[Coordinat
     which case you will be responsible for handling any checking, if necessary.
     """
 
-    resource.parent = self
-    resource.location = location
     try:
+      self.check_assignment(resource, location)
```
rickwierenga, line 140: `check_assignment` is sufficiently short so that we can put it above this try block. Happy to discuss.

rickwierenga (changes requested):
Thanks for the PR! Good finds, these error messages definitely make things clearer for the user.

I just had a few small comments about the code.

pylabrobot/resources/resource.py
naikymen: Yes :)

pylabrobot/resources/resource.py
naikymen: Hmm both can be true if the resource has a parent, and the user is trying to assign it it itself.

I made a list just to get every "wrong thing" into the message, thinking that more checks might be added in the future.

pylabrobot/resources/resource.py
naikymen: Sure, I only wanted it above this, to catch the errors before modifying the objects.

```
      resource.parent = self
      resource.location = location
```

I put it into the try clause just to reuse it, not a strong preference.

Im not confident about the place. Perhaps you can choose a better one?

Perhaps inside the callback? Which seems to do a name check already.

naikymen:
That was a quick response :)

I pushed an additional fix for the second proposed check.

pylabrobot/resources/resource.py
rickwierenga: I think it's fine to put it above the assignment of `location` and `parent`, as the first code in this method. That way we do not need set the values to `None` if an error occurs.

The reason I assigned them before calling the callback (and unassign them on error), is that some callbacks will want to use this information do determine validity. That is not the case here.

pylabrobot/resources/resource.py
rickwierenga: Right, let's keep the list for now.

rickwierenga:
Unit tests are failing because `assign_child_resource` used to override an existing parent. I think it's good to keep that behavior, but to make it optional using an `reassign: bool` parameter. Based on this parameter, an error would be raised if the resource already has a parent. What do you think is best: defaulting to true or false?

naikymen:
> Unit tests are failing because `assign_child_resource` used to override an existing parent. I think it's good to keep that behavior, but to make it optional using an `reassign: bool` parameter. Based on this parameter, an error would be raised if the resource already has a parent. What do you think is best: defaulting to true or false?

A re-assign behaviour with an existing parent makes sense to me too.

I'm not sure I understand completely though. Where should the reassign parameter be added?

rickwierenga:
To the `assign_child_resource` method, with a default value (true or false) so the user is not required to specify it. I'm not sure what default value I like best.

naikymen:
I see, hmm.

I added the check because I wanted to know if an instance of a resource could be lazily reused to add multiple similar resources to a deck.

That was why I would expect to see an error in my setup, people tend to be lazy.

On the other hand, I'm also not expecting stuff to move labware around (which might happen in the future). In the movable labware case the user will likely know that they're doing by reassignment.

Adding parameters and new functions complicates things. I think that `reassign=True` as default is good and simple to understand.

Perhaps a better check would be against re-assignment to the same parent, which would align with my original motivation, but allow reassignments.

naikymen:
However, that last case is already covered by a downstream exception "Error: Resource 'plate_01' already assigned to deck"

The only "trouble" with that implementation is that the resource's parent and location are reset to None. This is the only point in favour of writing a check for this case before the try clause.

naikymen:
Added the `reassign` parameter to `assign_child_resource` with default `True` (replicates the default behaviour). Stuff should not break.

Added a check on reassigning to the same parent, which is harmless by may be due to misuse of PLR. Not sure if this can break other things.

pylabrobot/resources/resource.py
naikymen: Done :ok:

pylabrobot/resources/resource.py
```diff
@@ -147,6 +153,31 @@ def assign_child_resource(self, resource: Resource, location: Optional[Coordinat
...
+    msgs = []
+
+    # Check for self assignment
+    if resource is self:
+      msgs.append(f"Will not assign resource '{self.name}' to itself.")
+
+    # Check for reassignment to the same parent
+    if resource.parent is not None and (resource.parent is self):
+      msgs.append(f"Will not reassign resource '{resource.name}' " +
+                  f"to the same parent: '{resource.parent.name}'.")
```
rickwierenga, line 168: Perhaps it's better to make this a warning (using [the logger](https://github.com/PyLabRobot/pylabrobot/blob/86ff89651bc41de093a9840613ba7a78136b55fe/pylabrobot/__init__.py#L7))? Something like `f"Resource '{resource.name}' already assigned to '{resource.parent.name}'"`. I'm somewhat hesitant to raising errors when something is technically OK.

It would crash if `reassign` is False.

rickwierenga:
> On the other hand, I'm also not expecting stuff to move labware around (which might happen in the future). In the movable labware case the user will likely know that they're doing by reassignment.

Yes. It's probably just when defining deck layouts. Beyond that, I think this should be PLR's concern, for regular users at least.

> The only "trouble" with that implementation is that the resource's parent and location are reset to None. This is the only point in favour of writing a check for this case before the try clause.

Right. I have briefly considered adding two callbacks, perhaps `will_assign_resource_callback` and `did_assign_resource_callback`. I think this is a good motivation for finally implementing that.

> Added the reassign parameter to assign_child_resource with default True (replicates the default behaviour). Other stuff should not break.

Nice!

> Added a check on reassigning to the same parent, which is harmless by may be due to misuse of PLR. Not sure if this can break other things.

Added a comment in the code.

pylabrobot/resources/resource.py
naikymen: Done :)

Note that now the downstream check (in `Deck._check_name_exists`)  is raised instead.

image

The logger shows: 

`2023-04-02 19:53:24,904 - pylabrobot.resources.resource - WARNING - Resource 'plate_01' already assigned to 'basic deck 1'`

naikymen: Output with reassign=False:

image

rickwierenga: Great!

rickwierenga (approved):

rickwierenga:
Thanks again!

PR #27, opened 2023-05-02
pylabrobot.resources.hamilton do not exist

isaacguerreir:
I was getting a error while trying to run this code, at it seems from the `pylabrobot` folder inside `site-packages` that pylabrobot only have two folder `utils` and `liquid_handling`.

From there I found a `resources` folder inside inside `liquid_handling`.

It's possible to reproduce this error by:

```
mkdir your_project
cd your_project
python -m virtualenv env
source env/bin/activate

pip install pylabrobot[all]
```

create a script using `from pylabrobot.resources.hamilton import STARLetDeck`

and it will generate a Module not found error.

### DIFF
--- a/docs/using-the-simulator.ipynb
+++ b/docs/using-the-simulator.ipynb
@@ -50,7 +50,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from pylabrobot.resources.hamilton import STARLetDeck"
+    "from pylabrobot.liquid_handling.resources.hamilton import STARLetDeck"
    ]
   },
   {


rickwierenga:
Thanks for the PR!

This is actually because the version on PyPI is outdated. If you install it from source, it should work. I will update the PyPI version shortly.

isaacguerreir:
Thanks, Rick!

I will close this issue so, let me know if would be interesting to have a CD pipeline for generating new pypi deployments for tagged or merged PR on main branch.

rickwierenga:
[Already have that](https://github.com/PyLabRobot/pylabrobot/blob/main/.github/workflows/python-publish.yml), I just haven't created a release on GitHub yet :)

PR #36, opened 2023-09-07
calibrations, PumpArrays, and the backend for Agrow Dose Pumps

maraxen:
tests for loading calibrations and agrow dose pumps in progress


pylabrobot/pumps/agrowpumps/agrowdosepump.py
```diff
@@ -0,0 +1,130 @@
...
+      keep_alive(self): This method sends a Modbus request every 25 seconds to keep the
+      connection alive.
+      """
+      while True:
+        time.sleep(25)
+
+        self.modbus.read_holding_registers(0, 1, unit=self.unit)  # type: ignore[union-attr]
+
+    thread = threading.Thread(target=keep_alive, daemon=True)
+    thread.start()
```
rickwierenga, line 70: I think we should maintain a reference to this thread on self so we can terminate it on `stop`.

pylabrobot/pumps/calibration.py
```diff
@@ -0,0 +1,155 @@
...
+
+  def __init__(self, calibration: Optional[List[Union[float, int]]] = None):
+    """
+    Args:
+        calibration: calibration of the pump in pump-specific volume per time/revolution units.
+    Raises:
+        ValueError: if a value in the calibration is outside expected parameters.
+    """
+    if calibration is not None and any(value <= 0 for value in calibration):
+      raise ValueError("A value in the calibration is is outside expected parameters.")
```
rickwierenga, line 24: Perhaps specify what the expected parameters are in the doc string / error message?

pylabrobot/pumps/pump.py
```diff
@@ -1,27 +1,71 @@
...
+  Attributes:
+      backend: The backend that the pump is controlled through.
+      calibration: The calibration of the pump.
+
+      """
 
-  def __init__(self, backend: PumpBackend):
+  def __init__(self, backend: PumpBackend, calibration: Optional[PumpCalibration] = None):
     self.backend: PumpBackend = backend
+    self.calibration = calibration
```
rickwierenga, line 19: Perhaps it's worth getting the single calibration from the object know, so we don't have to index it below for each use. It may also be good to raise an error if more than one item exists in the calibration list.

pylabrobot/pumps/pump.py
```diff
@@ -1,27 +1,71 @@
...
+    the pump to be calibrated at the input speed.
+    Args:
+      speed: speed in rpm/pump-specific units.
+      volume: volume to pump.
+    """
+    if self.calibration is None:
+      raise TypeError(
+        "Pump is not calibrated. Volume based pumping and related functions unavailable.")
+    duration = volume / self.calibration[0]
+    await self.run_for_duration(speed=speed, duration=duration)
```
rickwierenga, line 67: I think running for a certain number of revolutions may be more precise, since pumps could have acceleration / deceleration built in. I know the agrow pumps do not have this level of control. Do you think this should be a backend level command? Do you have other suggestions?

pylabrobot/pumps/backend.py
```diff
@@ -19,8 +20,47 @@ def run_continuously(self, speed: float):
     If speed is 0, the pump will be halted.
 
     Args:
-      speed: speed in rpm.
+      speed: speed in rpm/pump-specific units.
```
rickwierenga, line 23: I'm not a huge fan of having a parameter on the front end / abstract backend be something that's machine specific. What's the unit for the agrow pumps again?

rickwierenga (changes requested):
Great work!

I added a few thoughts to discuss in the code. Also, I think it's worth writing some tests / having copilot write some tests before we merge it into main.

pylabrobot/pumps/backend.py
maraxen: mL/second. In all liquid pumping I've worked with that has been the standard unit

pylabrobot/pumps/agrowpumps/agrowdosepump.py
maraxen: I made this change in the most recent commit

pylabrobot/pumps/calibration.py
maraxen: made this change!

pylabrobot/pumps/pump.py
maraxen: adjusted this to call the first item

pylabrobot/pumps/pump.py
maraxen: I agree, but I think a lot of pumps are not like the Cole Palmer. It would be strongly preferable if it was volume per rotation, as then you won't need the calibration to be speed specific

pylabrobot/pumps/backend.py
maraxen: for most pump implementations that don't have some kind of built-in tachometer/tracking of the number of rotations it will be volume per unit time, which is our implementation assumes unit volume per second

pylabrobot/pumps/pump.py
maraxen: I think we could have it on the backend as an override method.

rickwierenga: I agree that volume / rotation is more general.

How would you envision implementing it as an override method? The frontend would need to check whether the backend supports this operation, which is not a pattern used elsewhere in plr. I'm open to introducing it here, but would prefer to think about alternatives first.

* We could considering making `pump_volume` a backend-only method, like `run_continuously`. This would involve moving `run_for_duration` out of the frontend, which could get messy.
* Another option is having a parameter `method` (there's probably a better name) with options `"duration"` or `"rotations"`, which determines which backend command we call. What do you think?

pylabrobot/machine.py
```diff
@@ -56,11 +58,18 @@ async def stop(self):
 
   def __del__(self):
     if self.setup_finished:
-      self.stop()
+      self.backend.stop()
```
rickwierenga, line 61: I think we want to call `self.stop` here in case a frontend has a state (actually `Machine` already does: `_setup_finished`)

pylabrobot/pumps/pump.py
rickwierenga: Also, if the unit of calibration is volume/rotation, won't we need to having something like this:

```py
num_rotations = volume / self.calibration
duration = num_rotations / speed
```

pylabrobot/pumps/calibration_tests.py
```diff
@@ -0,0 +1,54 @@
...
+    calibration = PumpCalibration.load_calibration(self.json_list_path)
+    self.assertEqual(calibration[0], 1)
+    self.assertEqual(calibration[1], 1)
+    calibration = PumpCalibration.load_calibration(self.json_dict_path)
+    self.assertEqual(calibration[0], 1.0)
+    self.assertEqual(calibration[1], 1.0)
+    self.assertRaises(TypeError, PumpCalibration.load_calibration, self.json_null_path)
+    self.assertRaises(NotImplementedError,
+                      PumpCalibration.load_calibration,
+                      self.txt_path)
```
rickwierenga, line 54: I think these could be split up in separate tests for loading from a list, dictionary, csv, etc. This makes it easier to see where the issue is when a tests fails (based on which tests pass/fail), and allow testing of individual methods (tests should run very quickly here so not really an issue, but it's good practice to split it up).

pylabrobot/pumps/backend.py
rickwierenga: Ok. I will add a `tube_volume` attribute to Masterflex, so we can do the volume/time -> rotations/time (rpm) conversion on that backend.

pylabrobot/pumps/agrowpumps/agrowdosepump.py
```diff
@@ -0,0 +1,185 @@
...
+        Args:
+            speed: rate at which to run pump.
+            use_channels: pump array channels to run
+
+        Raises:
+            ValueError: Pump address out of range
+            ValueError: Pump speed out of range
+    """
+    assert self.modbus is not None, "Modbus connection not established"
+    assert self.pump_index_to_address is not None, "Pump address mapping not established"
```
rickwierenga, line 120: Could these checks be put in a decorator?

pylabrobot/pumps/agrowpumps/agrowdosepump.py
```diff
@@ -0,0 +1,185 @@
...
+      address = self.pump_index_to_address[pump]
+      self.modbus.write_register(address, 0, unit=self.unit)
+
+  async def stop(self):
+    """ Close the connection to the pump array. """
+    await self.halt()
+    assert self.modbus is not None, "Modbus connection not established"
+    self.modbus.close()
+    assert self.keep_alive_thread is not None, "Keep alive thread not established"
+    self.keep_alive_thread.join()
```
rickwierenga, line 145: I think we shouldn't assert that the modbus or thread exist before stopping. We just want to close them when they exist, but don't worry about the state where they are None (it could have failed to start), so I think we should say

```py
if self.modbus is not None:
  self.modbus.close()
if self.keep_alive_thread is not None:
  self.keep_alive_thread.join()
```

Note that the keep_alive_thread will not join, because it never quits. Instead, I would suggest adding a `_stopping_thread` (or similar) attribute, which you set to `True` right before the join, and rewriting `while True` to `while not _stopping_thread`. You can set it to `False` after the join succeeds.

pylabrobot/pumps/agrowpumps/agrowdosepump.py
```diff
@@ -0,0 +1,185 @@
...
+    self.connected = connected
+
+  async def connect(self):
+    self.connected = True
+
+  def read_holding_registers(self, *args, **kwargs):
+    assert self.connected, "Modbus connection not established"
+    for arg in kwargs.items():
+      if arg == "unit":
+        continue
```
rickwierenga, line 167: What does this do?

pylabrobot/pumps/agrowpumps/agrowdosepump.py
```diff
@@ -0,0 +1,185 @@
...
+      raise ValueError("unit must be specified")
+    if args[0] in range(100, 107):
+      return
+    else:
+      raise ValueError("address out of range")
+
+  def close(self):
+    assert self.connected, "Modbus connection not established"
+    self.connected = False
+    del self
```
rickwierenga, line 185: Why should closing the connection destroy the object? I think more likely than not it's fine to have Python destroy the object when it's no longer referenced.

pylabrobot/pumps/pump_tests.py
```diff
@@ -0,0 +1,224 @@
...
+                              volume=volume)
+            self.assertRaises(ValueError, pump_array.pump_volume,
+                              speed=speed,
+                              use_channels=use_channels,
+                              volume=test_volume)
+
+  async def test_stop(self):
+    async with PumpArray(backend=self.agrow_backend,
+                         calibration=self.null_calibration) as pump_array:
+      await pump_array.stop()
```
rickwierenga, line 218: Isn't the context manager supposed to stop it after the context ends?

rickwierenga (changes requested):

pylabrobot/pumps/pump.py
maraxen: I like the option of keeping it on the front end only but checking the backend for the method. I think `calibration_unit` could be the parameter. Would we want to have an attribute that is set on the backend (`backend.calibration_unit`) or do you think another approach would be better?

pylabrobot/pumps/backend.py
maraxen: Okay, that sounds good

pylabrobot/pumps/calibration_tests.py
maraxen: I ran into this problem when I was debugging. I think I want to keep a full set of tests like this for load_calibration, but only for the ones where there is some transformation of the input prior to the loading into the sub-function. I'll also add the dictionary and lists to SetUp so they don't have to be written out here.

pylabrobot/pumps/agrowpumps/agrowdosepump.py
maraxen: good thinking!

pylabrobot/pumps/agrowpumps/agrowdosepump.py
maraxen: okay, I'll make these changes. I think I only added these asserts to make mypy happy.

pylabrobot/pumps/pump_tests.py
maraxen: good catch!

pylabrobot/pumps/agrowpumps/agrowdosepump.py
maraxen: I'll delete line 185

pylabrobot/pumps/calibration.py
```diff
@@ -0,0 +1,181 @@
...
+    self.calibration = calibration
+
+  def __getitem__(self, item) -> Union[float, int]:
+    if self.calibration is None:
+      raise TypeError(
+        "Pump is not calibrated. Volume based pumping and related functions unavailable.")
+    return self.calibration[item]  # type: ignore
+
+  def __bool__(self):
+    return self.calibration is not None
```
rickwierenga, line 36: I don't think this is still needed, right?

Also, I think the calibration attribute should never be None on this class - it is cleaner to have the PumpCalibration variable/attribute elsewhere be None than to initialize it to a 'useless' (in that context) object.

pylabrobot/pumps/calibration.py
```diff
@@ -0,0 +1,181 @@
...
+    Returns:
+      PumpCalibration
+
+    Raises:
+      ValueError: if the calibration dictionary is not formatted correctly.
+    """
+
+    if not any(key == 0 for key in calibration.keys()):
+      if any(key == 1 for key in calibration.keys()):
+        calibration = {key - 1: value for key, value in calibration.items()}
```
rickwierenga, line 129: This allows for an ambiguous use of the API. I think we should have the user input a list in a singular format for clarity.

pylabrobot/pumps/calibration.py
```diff
@@ -0,0 +1,181 @@
...
+      else:
+        raise ValueError("Calibration dictionary keys must start at 0 or 1.")
+    if len(calibration) == 1:
+      calibration = {0: list(calibration.values())[0]}
```
rickwierenga, line 133: Given the code above, this wouldn't change the list. When the user inputs a singular format (always starting with a 0, or a 1) this should no longer be needed.

pylabrobot/pumps/calibration.py
```diff
@@ -0,0 +1,181 @@
...
+    if any(key < 0 for key in calibration.keys()) or any(key >= len(calibration) for key in
+                                                         calibration.keys()):
+      raise ValueError("Calibration dictionary keys must be non-negative and less than the "
+                       "length of the dictionary.")
+    if any(value <= 0 for value in calibration.values()):
+      raise ValueError("Calibration dictionary values must be positive.")
+    if sorted(calibration.keys()) != list(range(len(calibration))):
+      raise ValueError("Calibration dictionary keys must be a contiguous range of integers.")
```
rickwierenga, line 141: This is equivalent to the test above (l.134, `if any(key < 0...`) but a lot cleaner IMO. Do you want to keep this one and remove the other one?

pylabrobot/pumps/pump_tests.py
```diff
@@ -0,0 +1,235 @@
...
+    self.agrow_backend = AgrowPumpArray(port="simulated", unit=1, keep_alive_thread_enabled=False)
+    self.valid_durations = [int(0), int(100), float(0), float(100)]
+    self.invalid_durations = [int(-1), float(-1)]
+    self.mock_backend = Mock(spec=PumpArrayBackend)
+    self.null_calibration = PumpCalibration()
+    self.test_calibration = PumpCalibration.load_calibration(1, num_items=6)
+
+  async def test_setup(self):
+    """ Test that the AgrowPumpArray class can be initialized.  """
+    pump_array: PumpArray
```
rickwierenga, line 81: Is this still needed?

pylabrobot/pumps/calibration_tests.py
```diff
@@ -0,0 +1,73 @@
...
+    self.assertEqual(calibration[1], 2.0)
+
+  def test_load_from_value(self):
+    calibration = PumpCalibration.load_calibration(self.test_value, 2)
+    self.assertEqual(calibration[0], 1.0)
+    self.assertEqual(calibration[1], 1.0)
+
+  def test_uncalibrated(self):
+    calibration = PumpCalibration.uncalibrated()
+    self.assertIsNone(calibration.calibration)
```
rickwierenga, line 73: This test should no longer be needed if PumpCalibration always contains a calibration. (pump.py and pumparray.py already use it like that.)

pylabrobot/pumps/pump_tests.py
```diff
@@ -0,0 +1,235 @@
...
+      self.assertEqual(pump.calibration, PumpCalibration.load_calibration())
+      self.assertEqual(pump.backend, self.masterflex_backend)
+      self.assertEqual(pump.backend.com_port, "simulated")
+
+  async def test_run_revolutions(self):
+    """
+    Test that the Pump class can run for a specified number of revolutions.
+    """
+    async with Pump(backend=self.masterflex_backend, calibration=self.test_calibration) as pump:
+      await pump.run_revolutions(num_revolutions=1)
```
rickwierenga: What does this test?

pylabrobot/pumps/pumparray.py
```diff
@@ -0,0 +1,142 @@
...
+    elif calibration_units == "revolutions":
+      num_rotations = [channel_volume / self.calibration[channel] for channel, channel_volume in
+                       zip(use_channels, volume)]
+      tasks = [asyncio.create_task(
+        self.run_revolutions(num_revolutions=num_rotation,
+                             use_channels=channel))
+          for num_rotation, channel in zip(num_rotations, use_channels)]
+    else:
+      raise ValueError("Calibration units must be 'duration' or 'revolutions'.")
+    await asyncio.gather(*tasks)
```
rickwierenga, line 138: Nice!

pylabrobot/pumps/pumparray.py
```diff
@@ -0,0 +1,142 @@
...
+
+    Args:
+      speed: speed in rpm/pump-specific units. use_channels: pump array channels to run using
+        0-index. volume: volume to pump.
+    calibration_units: units of calibration. Volume per seconds ("duration") or volume per
+      revolution ("revolutions").
+
+    Raises:
+      TypeError: if the pump is not calibrated.
+    """
```
rickwierenga, line 106: Since calibration is specified on the class, I think `calibration_units` should be as well. Perhaps it is even better to store the calibration units on the PumpCalibration object since they'd always be used together.

rickwierenga (changes requested):

pylabrobot/pumps/pump_tests.py
```diff
@@ -0,0 +1,220 @@
...
+          with self.assertRaises(ValueError):
+            await pump_array.run_continuously(speed=speed,
+                                              use_channels=use_channels)
+      for use_channels in self.agrow_faulty_channels:
+        for test_speed in self.agrow_valid_speeds:
+          speed = [test_speed] * len(use_channels)
+          with self.assertRaises(ValueError):
+            await pump_array.run_continuously(speed=speed,
+                                              use_channels=use_channels)
+      await asyncio.sleep(25)
```
rickwierenga, line 120: what is this for?

rickwierenga:
Some changes:

- Fixed tests
  - `self.assertRaises(Exception` does not work for async functions, switch to `with self.assertRaises`
  - `pump_array.pump_volume` no longer takes calibration_units
- Use `await self.pump_array.setup()` in `async def asyncSetUp(self)` so that we don't have to indent all tests for the context manager
- Placed the testing values for pumps & calibration in the test case instead of setup. This makes it easier to see what is going on
- Removed the loops in tests, because we can reasonably expect permutations to work & loops make it hard to see which exact value is failing
- Introduced `NotCalibratedError` instead of `TypeError` for non-calibration
- Split the tests for AgrowPumps and PumpArray so each test their respective layer: the front end should test the backend is called correctly, the backend should test that the right commands are sent
- Move `any(channel not in range(0, self.num_channels) for channel in use_channels)` check to PumpArray, this is a general check that should exist on the front end
- read_holding_registers is not an async function
- no longer use dict reader for the csv: this was treating the first line as the header which was wrong
- switch everything over to 0-indexing, some were left at 1
- `threading.Thread` takes a callable as a target
- switch thread sleep frequency to 1/10s so we can quit faster. Still sending healthchecks every 25s
- Simply mock the modbus instead of a separate `AgrowPumpArrayTester` object: less duplication and more confidence in tests
- removed leftover empty calibrations, mostly in tests

I left two `TODO(marielle)`. It would be nice if you could look at those.

PR #37, opened 2023-09-12
Tecan dev branch

b-mahjour:
Ongoing efforts to optimize connectivity with TECAN robot *hopefully* without breaking abstraction


rickwierenga:
Looking good! Is this still a WIP or can I start reviewing it?

rickwierenga:
This PR should work fine! Whatever is most convenient for you.

Also, please let me know if I can be helpful in any way!

b-mahjour:
Hey Rick,

I've gotten around to this again, and I've made some updates to be integrate from my end. 

I'd like to programmatically control moving plates around using the liquid_handling_server.py interface. Right now there's only liquid handling commands. Do you think it would be hard to write a function that allows me to call the API to move a place from one spot to another? Let me know how I should approach this.

Thanks,
Babak

rickwierenga:
Awesome! Happy to hear you're back to working on this and great to see this progress!

> async def rinse_tips(self):

I'm guessing this method is probably specific to your protocol/setup and not a general Tecan capability. The role of backends should be facilitating communication between the general frontend (LiquidHandler) and specific devices. Higher level methods should exist in a user-defined protocol/script. This is to keep backends maximally simple and general, which makes adding new devices as easy as possible. I think rinse_tips may be a protocol-level function, rather than a backend-level function. For this reason, I'd like to not merge this into main.

This is based on an assumption, I could be wrong, and definitely happy to discuss!

> I'd like to programmatically control moving plates around using the liquid_handling_server.py interface. Right now there's only liquid handling commands. Do you think it would be hard to write a function that allows me to call the API to move a place from one spot to another? Let me know how I should approach this.

This should not be hard. In general, the server just takes in the parameters LH would use in JSON format. For the `move_resource` function, there are two options:

- move to coordinate: JSON should contain a destination coordinate
- move to resource: JSON should contain the `name` of the destination resource

In both cases, the source/target resource name should be included.

I hope the other methods can serve as an example, but please let me know if I can be helpful!

b-mahjour:
>I think rinse_tips may be a protocol-level function, rather than a backend-level function.

Sure, but it's not clear where I'd define such a custom function. Since I am interfacing through the API, would it be written there? Can I have such fine-tuned control to write a custom function from this level? Let me know what you're thinking and the best way to go about it.

```
@lh_api.route("/rinse", methods=["POST"])
async def rinse_tips():
  # add custom function logic here maybe?
  return add_and_run_task(Task(current_app.lh.rinse_tips()))
```

And no need to worry about pushing anything to main, this branch is intended to showcase the changes I had to make to get it working for a synthesis project in our lab, perhaps to serve as a reference for another's project in a future. But if there are valuable updates you'd like to incorporate into the main branch, by all means.

I'll take a crack at writing the move_resource API tomorrow when I am in lab

b-mahjour:
@rickwierenga 

Do you think you could take a look at my attempt to write the ROMA API? I'm passing the data correctly but I am getting this error: "Operation is not supported by resource TecanPlateCarrier(name=plate_carrier_1, location=(238, 061, 000), size_x=149.0, size_y=295.0, size_z=6.0, category=tecan_plate_carrier)."

resource_plate = "reaction_setup"
to = "carrier-plate_carrier_2-spot-0"


An issue I'm also encountering is that every time I update the API, I have to reboot the server and rerun setup. Is there a way such that I don't have to rerun setup each time? It takes 5 minutes or so each time.

b-mahjour:
```
The line that's raising the error is:

if par.roma_x is None or par.roma_y is None or par.roma_z_safe is None \
      or par.roma_z_travel is None or par.roma_z_end is None:
      raise ValueError(f"Operation is not supported by resource {par}.")
```

b-mahjour:
would be nice if there were some 'reset' functionalities to move the LiHA/RoMA back into a place in the event of a 'not fetched' or some other type of workflow error

b-mahjour:
Actually, the setup does not initialize the z location of the RoMA. Currently, there is no method to reset the RoMA if the workflow crashes in its protracted position. Is there an analogous serial command I can use like we use for the x position?

```
  async def position_initialization_x(self):
    """ Reinitializes X-axis of the arm. """

    await self.backend.send_command(module=self.module, command="PIX")

```

b-mahjour:
image

b-mahjour:
Actually it is a bit confusing as I assume this line should reset the arm:
`    self._roma_connected = await self.setup_arm(EVO.ROMA)`

b-mahjour:
One of these lines are leading to a collision error:

```
    await self.roma.set_vector_coordinate_position(1, x, y, z["end"], 900, None, 1, 1)
    await self.roma.set_vector_coordinate_position(2, x, y, z["travel"], 900, None, 1, 2)
    await self.roma.set_vector_coordinate_position(3, x, y, z["safe"], 900, None, 1, 3)
    await self.roma.set_vector_coordinate_position(4, xt, yt, zt["safe"], 900, None, 1, 4)
    await self.roma.set_vector_coordinate_position(5, xt, yt, zt["travel"], 900, None, 1, 3)
    await self.roma.set_vector_coordinate_position(6, xt, yt, zt["end"], 900, None, 1, 0)
```

Any thoughts on how to debug? Right now the arm will grip the plate, but then it fails. 

The error is ('Invalid command (collision)', C1, 2)

b-mahjour:
Some additional notes as I continue to try and work this out:

We've swapped the locations of the PnP and RoMA arms – this is leading to some fairly violent crashes as the RoMA or LiHA attempt to get into position. It seems that I will have to implement control over the PnP and move that before any other operation occurs. Is there an easy interface for this?

b-mahjour:
>I see that the last commit message is "everything is working now". Do you still need help with this?

I think we should be all good except for whether we want to add 'rinse_tips' into main or not (and some messy print statements I've added in the process of debugging; can clean up later). The rest should be ok for 'general use'.  

As for the question of whether 'rinse_tips' is a general method, I would think yes for the Tecan – basically a valve rotates allowing water to be dispensed from a centralized water reservoir (as opposed to for instance aspirating a wash solvent and dispensing into a waste location). In any case I can move this functionality elsewhere for now and clean up the branch.


>Do you really need to use the server to communicate with the robot? With PLR you can just connect the robot directly to your laptop (any OS) and run commands from there during development, and test on the server later. I really like using Jupyter notebooks for this purpose.

Hmm I think I had to reset the jupyter notebook each time I updated the code, and had to run setup again anyways. I could be misremembering though, I'll try it out and see.

rickwierenga:
> I think we should be all good except for whether we want to add 'rinse_tips' into main or not (and some messy print statements I've added in the process of debugging; can clean up later). The rest should be ok for 'general use'.

Awesome!

There are a bunch of 'magic numbers' in the rinse_tips methods. Will `self.liha.position_absolute_all_axis(45, 1031, 90)` be the same on all robots? Is there an ordering to the 5 'blocks' of code (l.325, l.331, etc.) that we can abstract away into nice general methods?

My idea is that if all backends simply implement 4 atomic methods (tip pickup/drop, aspirate, dispense) then everything else can be higher level and shared across all robots. Backends should be maximally simple. This makes adding new backends maximally easy. Of course, there should be room to support specific utilities for each robot on the backends, but as you understand I'm hesitant to reimplementing 'general'/high-level functions on backends.

I feel like ultimately this is going to have to be your call, but I'm hugely in favor of erring on the side of simplicity when in doubt.

> Hmm I think I had to reset the jupyter notebook each time I updated the code, and had to run setup again anyways. I could be misremembering though, I'll try it out and see.

There is an [autoload](https://ipython.org/ipython-doc/3/config/extensions/autoreload.html) extension, which works most of the time, but can also be horribly confusing. What I do is just write the functions using `send_command` (on hamilton) and sort of hack it until i get a method that I paste into a file. Interactive development is a huge + of PLR and I want you to be able to use this!

b-mahjour:
```
    plate.unassign()
    if isinstance(to, Coordinate):
      to_location -= self.deck.location # passed as an absolute location, but stored as relative
      self.deck.assign_child_resource(plate, location=to_location)
    elif isinstance(to, CarrierSite): # .zero() resources
      to.assign_child_resource(plate, location=Coordinate.zero())
    elif isinstance(to, (ResourceStack, PlateReader)): # manage its own resources
      to.assign_child_resource(plate)
    else:
      to.assign_child_resource(plate, location=to_location)
```

Not sure why, but this code reassigning the location of the plate in the internal rep. of the deck after a plate transfer does not run (end of   async def move_plate part of liquid_handler.py). I tried removing the superfluous "return" in move_resource to no avail.

rickwierenga:
I think your dev branch may be lagging behind main a little bit. Before I dive in, would you mind if I merged main? I should have edit rights.

b-mahjour:
ah yeah go ahead I just merged feel free to change what you'd like

b-mahjour:
```
if isinstance(to, Coordinate):
      to_location -= self.deck.location # passed as an absolute location, but stored as relative
      self.deck.assign_child_resource(plate, location=to_location)

```

This line is causing the crash, I'm not sure why it's not reassigning correctly. These are the coordinates sent via the "Move" object i.e. the coordinates in which the plate is moved to.

>Location (948.400, 072.500, 063.000) is already occupied by resource 'plate_carrier_3'.

rickwierenga:
> I'm wondering what would be the best parameter to change to have it drop the plate a bit closer to the reactor surface. Right now it kind of drops it a bit too high. It seems there are several ways I could go about it; what would be the best practice?

While you can obviously adjust the resource location/size to achieve the same results, I would recommend using parameters in the `Move` operation.

LiquidHandler provides the standardized `pickup_distance_from_top` parameter for this exact reason, which EVO does not implement yet. This should be easy to do by passing `move.pickup_distance_from_top` as a parameter to the `EVO._roma_positions` method calls in `EVO.move_resource` and adjusting the `z_positions["end"]`. Could you let me know if that works?

I'm working with Vilhelm from DTU, who is also using PLR with a Tecan EVO, on updating the Tecan resource definitions to the latest PLR format ([link](https://forums.pylabrobot.org/t/pylabrobot-tecan-error-in-adding-labware-to-carrier/2987). It would be great to merge your changes into main soon so that you can profit from each other's work!

> Similarly, I need the arms to widen a bit further after it drops it so it doesn't clip the plate on its way back up. Any ideas here?

You should be able to adjust the value in `position_absolute_g` (l.522 for me, first after `# pick up resource`) which is currently 900. This parameter specifies the distance of each gripper arm to the center of the gripper arms in units of 0.1mm. If you're gripping on the long side of the plate in landscape mode, you'll probably want to use `int(plate.get_size_x() + margin)*10` where maybe `margin = 5` (mm) or similar. Please let me know if that works.

b-mahjour:
I'll test out your suggestions later at some point when I get a chance, but for now I cleaned up the branch so it could be merged into main hopefully

rickwierenga:
Apologies for the delay. I'm still working on updating the Tecan definitions to be compatible with the latest version of PLR.

PR #39, opened 2023-09-30
Add missing __init__.py

tfehlmann:
Without the `__init__.py` the liquid_classes folder doesn't get installed with pip install.


rickwierenga:
Thanks!

PR #40, opened 2023-10-06
include the full width of the last rail in available width calculations

ben-ray:
this tiny change allows a user to place labware on the rightmost rail of STAR and STARlet robots


rickwierenga:
Thanks!

The linting issue appeared after upgrading pylint from 2.17.4 to 3.0.1. I'll merge this and look at what the exact problem is soon.

PR #42, opened 2023-10-13
implement CO-RE gripper

ben-ray:


rickwierenga:
Thanks again!

A few comments which should be quick to resolve:

* I don't think commands like `get_core` and `put_core` should be part of an abstract liquid handler backend, because they are extremely Hamilton specific. Instead, I think `move_resource` in Hamilton should have a flag `use_arm` that takes `"iswap"` or `"core"`, and then calls the corresponding method on itself. Currently it always uses iswap (`pick_up_resource`, `move_picked_up_resource` and `release_picked_up_resource`, which could be renamed to `iswap_*`).

* It seems that the channels used are hard coded. Could you make those parameters?

* Could you remove the `build/lib` folder?

ben-ray:
your wish is my command! I'm not sure I implemented these changes in the most developer-friendly way, but I do think it is now abstracted as you'd prefer. this makes a lot of sense from the perspective of supporting many different robots.

rickwierenga:
Could you enable the option "Allow edits from maintainers" on this PR so that I can push to your branch? I made some small refactoring changes.

ben-ray:
It looks like github organization accounts [don't have access to this option](https://github.com/orgs/community/discussions/5634), so I just gave you write access instead - should be equivalent?

rickwierenga:
That works! I think "duplication is cheaper than the wrong abstraction" here.

I made iswap the default arm for movements, to be consistent with earlier versions.

Lastly, could you get a firmware string to write a unit test for this method?

ben-ray:
Very nice changes - much better this way.

I made the CO-RE gripper the default arm for movements because in practice, we use it for >90% of all movements inside the deck. It is much faster, more reliable, and less error-prone than the iSWAP gripper. From a pure function standpoint, making the CO-RE gripper default will match how (I think) most hamilton experts will be using their robots.

Here are the firmware strings for your newest Move() function using the CO-RE gripper with pickup_distance_from_top=13:
```
Get CoRe gripper tool:        C0ZTid0020xs07975xd0ya1250yb1070pa07pb08tp2350tz2250th2450tt14
Get plate using CoRe gripper: C0ZPid0021xs04150xd0yj1145yv0050zj1876zy0500yo0890yg0830yw15th2840te2750
Put plate using CoRe gripper  C0ZRid0022xs04150xd0yj2105zj1876zi000zy0500yo0890th2840te2750
Discard CoRe gripper tool:    C0ZSid0023xs07975xd0ya1250yb1070tp2150tz2050th2450te2450
```

Let me know if I should make the tests too - unsure how granular these should be when you already have pretty great coverage on the iSWAP tests

ben-ray:
I just noticed something interesting about the pick_up_tips function. Typically during a process, I won't know exactly what tiprack I am picking tips from. There is a good chance I will need to pick up tips from more than one different tiprack (if I am on tip #94, but on my next pick_up_tips command I need to pick up 8 more tips, so I should pick up 1 tip from the first rack and the other 8 tips from the next rack). But as the pick_up_tips implementation currently stands, there's no way I can handle tip pickup from an arbitrary number of different tipracks - I can only pick up tips from a single rack resource.

In pyhamilton, this is solved by inputting [(rack, index), (rack, index), ... ] but in pylabrobot, this is set up as rack[index, ... ]

What do you think is the ideal abstraction for adding this functionality?

rickwierenga:
Thanks for the strings, I'll add tests.

> What do you think is the ideal abstraction for adding this functionality?

This should already be possible! When you index a tip rack, you get back a list of `TipSpot` objects (not `Tip`, because that wouldn't work with `drop_tip`). You should be able to pass a list with `TipSpot`s from multiple tip racks: `lh.pick_up_tips(tip_rack_a["A1"] + tip_rack_b["A1"])`.

For your application, perhaps it'd be neat to use a generator:

```python
def get_tips(n):
  index = 0
  max_tips = 96 * num_tip_racks
  spots = tip_rack_a.get_all_items() + tip_rack_b.get_all_items() + ...
  while index < max_tips:
    yield spots[index:index+n]
    index += n
  raise StopIteration()
```

rickwierenga:
What happens if the grippers are mounted before `setup` is called?

ben-ray:
good catch! currently the STARlet will just throw the co-re grippers in the trash (same behavior if it detects pipettes during setup(). I am unsure if there is a way to detect the difference between a pipette tip and a CO-RE gripper...

rickwierenga:
Do you think anything should be changed before merging?

ben-ray:
I think this is great to go. Thank you very much on all the help & feedback!

ben-ray:
a historic occasion for PR #42 - thanks for making it happen!

rickwierenga:
Thanks for the work!

PR #46, opened 2023-11-23
bottom type can be a number only

DonSqualo:
also getattr doesn't work for me

### DIFF
--- a/pylabrobot/resources/well.py
+++ b/pylabrobot/resources/well.py
@@ -55,7 +55,7 @@ class Well(Container):
   def serialize(self):
     return {
       **super().serialize(),
-      "bottom_type": self.bottom_type.value,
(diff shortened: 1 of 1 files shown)


rickwierenga:
In which case would bottom_type not be of type WellBottomType? I think that's the underlying issue.

DonSqualo:
My generated json save file contains entries that say bottom_type: 1 
If I try importing them without this fix I get an error
image

rickwierenga:
This was actually an issue with the deserializer/initializer. Fixed with https://github.com/PyLabRobot/pylabrobot/commit/6dbff8d534663c7c4a20fe1e22b5d1357fd8d369. (the raw value in the JSON doesn't really matter, but made it a string for clarity)

PR #47, opened 2023-11-24
Allow sending of direct firmware commands to STAR

DonSqualo:


rickwierenga:
Great suggestion.

Two tiny stylistic details:
- could you use 2 spaces for indentation?
- I usually find it cleaner to raise/return in the if portion of an if statement as opposed to else portion, because this way you don't need `else` and prevent very deep indentation levels.

DonSqualo:
done

PR #48, opened 2023-11-28
Clarified usage of low level functions

DonSqualo:


rickwierenga (changes requested):
could you fix the indentation where necessary?

PR #49, opened 2023-12-19
Machine: Add return values to need_setup_finished

vkorotkovs:
need_setup_finished from Machine.py was missing return values. All functions wrapped would return None


rickwierenga:
Thanks!

PR #50, opened 2023-12-19
Powder Dispenser: Initial commit

vkorotkovs:
Initial commit to include front end class for a powder dispenser, abstract class for a backend and a powder resource. The front end class is missing deck related methods that the liquid handlers have. Though, the powder dispensers usually have simpler decks and manipulations.  Could may be transfer methods like _assert_resources_exist to base class?

### DIFF
--- a/pylabrobot/powder_dispensing/__init__.py
+++ b/pylabrobot/powder_dispensing/__init__.py
@@ -1,0 +1,1 @@
+from .powder_dispenser import PowderDispenser
--- a/pylabrobot/powder_dispensing/backend.py
+++ b/pylabrobot/powder_dispensing/backend.py
@@ -1,0 +1,60 @@
+from __future__ import annotations
+from abc import ABCMeta, abstractmethod
+from typing import List
+
+from pylabrobot.machine import MachineBackend
+from pylabrobot.resources import Resource, Powder
+from typing import Optional, Dict, Any
+
+
+class PowderDispenserBackend(MachineBackend, metaclass=ABCMeta):
+  """
+  An abstract class for a powder dispenser backend.
+  """
+
+  @abstractmethod
+  async def setup(self) -> None:
+
+    """Set up the powder dispenser."""
+
+  @abstractmethod
+  async def stop(self) -> None:
+    """Close all connections to the powder dispenser and make sure setup() can be called again."""
+
+  @abstractmethod
+  async def dispense(
+    self,
+    dispense_parameters: List[PowderDispense],
+    **backend_kwargs
+  ) -> List[DispenseResults]:
+    """Dispense powders with set of dispense parameters."""
+
+
+class PowderDispense:
+  """
+  A class for input parameters for powder dispensing.
+  """
+
+  def __init__(
+    self,
+    resource: Resource,
+    powder: Powder,
+    amount: float,
+    params: Optional[Dict[str, Any]] = None,
+    **kwargs
+  ) -> None:
+    self.resource = resource
+    self.powder = powder
+    self.amount = amount
+    self.params = params
+    self.kwargs = kwargs
+
+
+class DispenseResults(dict):
+  """
+  A class for results of powder dispensing, behaving like a dictionary
+  but ensuring the presence of 'actual_amount'.
+  """
+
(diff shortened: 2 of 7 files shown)


vkorotkovs:
> Initial commit to include front end class for a powder dispenser, abstract class for a backend and a powder resource. The front end class is missing deck related methods that the liquid handlers have. Though, the powder dispensers usually have simpler decks and manipulations. Could may be transfer methods like _assert_resources_exist to base class?

Some linting and testing issues. Will work on fixing.

rickwierenga:
Looking good!

What is the rationale behind making `params` on PowderDispense a dict? I think it would be more readable if the parameters were actual attributes.

Under the assumption that this is meant for machine-specific parameters, in the past we have used the `backend_kwargs` parameter for that. With other parts of PLR, any parameters not recognized by the frontend (the `**backend_kwargs` of the function header) would be passed to the backend (as actual parameters, that are defined and typed by the backend). The operation object contains only parameters generally expected to be supported.

The reason for this is that it is hard to generally match anything sent to the front end via `backend_kwargs` to a specific operation we construct. For example, we might have 1) a `special_parameter` that is a list, with one value per operation and 2) `another_special_parameter` that is used for all operations. How to tell which is which?

In this case, `dispense_parameters` is almost a workaround for sending parameters to operations, with `backend_kwargs` still existing for the general case. While I like the idea of having parameters be part of the `PowderDispense` class, I also think it introduces more complexity in the API:

```py
dispenser.dispense(
  resources=[plate["A1"]],
  powders=[Powder("salt")],
  amounts=[0.05],
  special=[42], # special is a backend_kwarg, for the operations
  another_special=21) # another_special is another backend_kwarg, applied to all operations
```

vs.

```py
dispenser.dispense(
  resources=[plate["A1"]],
  powders=[Powder("salt")],
  amounts=[0.05],
  dispense_parameters=[{"special": 42}])
```

I think the first case is more uniform and preferred. We could think of a way to pass `special` to the operations (detect it's a list?), and `another_special` as a regular backend_kwarg. This would introduce some complexity on the `PowderDispense`, because it has to store this dictionary.

Happy to discuss!

> Could may be transfer methods like _assert_resources_exist to base class?

Great idea! Feel free to move that to `Deck`.

vkorotkovs:
Thanks for the great feedback!

Could also pass all parameters as part of `backend_kwargs ` and let the backend implementation do the assertion of list lengths. Since the name and type of parameters would be still up to each backend implementation. This way you don't have to worry about adapting to the whole spectrum of potential backends but still have a cleaner API. One could also add store them in a `Powder` class for convenience.

rickwierenga:
> Could also pass all parameters as part of backend_kwargs  and let the backend implementation do the assertion of list lengths

Yes, exactly.

vkorotkovs:
> > Could also pass all parameters as part of `backend_kwargs`  and let the backend implementation do the assertion of list lengths
> 
> Yes, exactly.

Added the code changes also added assertion for `resources` length. Changed the amount description to mg instead of kg.

rickwierenga:
Great!

vkorotkovs:
Are there any other items that you would like to see in this PR? Shall I squash the two commits?

rickwierenga:
Up to you. Do you want to keep working in this PR or open a new one for the concrete backend implementation? I'm not sure I should merge this into main quite yet, but I'd be happy to merge it into a power_dispenser branch in this repo if you'd like.

There's no need to squash commits.

vkorotkovs:
I think a separate backend PR would make most sense. I'm not so sure how useful a separate branch in this repo would be regarding maintenance.

I will continue working on the backend and do a separate PR from the fork when available. This way I can maintain the forked main to avoid any conflicts in the future.

What requirements do you need or want to fulfill before merging into main?

rickwierenga:
Sounds good!

> What requirements do you need or want to fulfill before merging into main?

I think going forward we should aim to only merge things into main when they are ready for people to use. While here the foundation is laid for developers to implement power_dispensing backends, the end-users are unable to use these classes yet (for the simple reason that no concrete backend exists).

It's not really a big deal at this point and if you're more comfortable using main then merging this maximizes total productivity, so I'll do that.

PR #52, opened 2023-12-22
Add pytest filter to suppress expected warning

ecalifornica:
This change cleans up the test output by preventing the expected warning from being printed. No alterations were made to the test logic or functionality of the code being tested.


rickwierenga:
Thanks for this one too!

PR #53, opened 2023-12-22
Update liquidhandler argument names

ecalifornica:
Looks like the argument name was changed in https://github.com/PyLabRobot/pylabrobot/pull/35 but the callers were not updated? This was causing test warnings from the LiquidHandler class, also updated SerializingBackend for consistency.


rickwierenga:
Thanks a lot!

PR #55, opened 2023-12-25
Enable adding labware in any order

BioCam:
1. Fixed a small logic bug: inserting a carrier in between two carriers (exactly one carrier apart in the x-dimension) incorrectly threw up a ValueError if carriers are of the same size (specialised case but important for moving a central carrier in and out of the deck in an interactive manner during an automation run). 
2. Changing `if nested-logic AND nested-logic` condition to represent `nested-logic` with any() to enhance readability.


rickwierenga:
Good catch! Appreciate the Christmas PR :)

PR #56, opened 2023-12-30
Add liquid handler operation callbacks

ecalifornica:
A quick pass at adding callback registration to `LiquidHandler`, generally following your suggestions @rickwierenga. Let me know what you think about this approach. Would you want to handle state tracker transactions in the callback?

### DIFF
--- a/pylabrobot/liquid_handling/liquid_handler.py
+++ b/pylabrobot/liquid_handling/liquid_handler.py
@@ -62,6 +62,25 @@ class LiquidHandler(MachineFrontend):
   defined in `pyhamilton.liquid_handling.backends`) to communicate with the liquid handler.
   """
 
+  ALLOWED_CALLBACKS = {
+    "aspirate",
+    "aspirate_plate",
+    "discard_tips",
+    "dispense",
+    "dispense_plate",
+    "drop_tips",
+    "drop_tips96",
+    "move_lid",
+    "move_plate",
+    "move_resource",
+    "pick_up_tips",
+    "pick_up_tips96",
+    "return_tips",
+    "return_tips96",
+    "stamp",
+    "transfer",
+  }
+
   def __init__(self, backend: LiquidHandlerBackend, deck: Deck):
     """ Initialize a LiquidHandler.
 
@@ -1401,18 +1420,27 @@ class LiquidHandler(MachineFrontend):
 
   def register_callback(self, method_name: str, callback: OperationCallback):
     """Registers a callback for a specific method."""
-    if self._callbacks.get(method_name):
+    if method_name in self._callbacks:
       error_message = f"Callback already registered for: {method_name}"
+      raise RuntimeError(error_message)
+    if method_name not in self.ALLOWED_CALLBACKS:
+      error_message = f"Callback not allowed: {method_name}"
       raise RuntimeError(error_message)
     self._callbacks[method_name] = callback
 
-  def _trigger_callback(self, method_name: str, *args, **kwargs):
-    """Triggers the callback associated with a method, if any."""
-    error = kwargs.pop("error", None)
+  def _trigger_callback(self, method_name: str, *args, error: Optional[Exception] = None, **kwargs):
+    """Triggers the callback associated with a method, if any.
+
+    NB: If an error exists it will be passed to the callback instead of being raised.
+    """
     if callback := self._callbacks.get(method_name):
       callback(*args, error=error, **kwargs)
     elif error:
       raise error
+
+  @property
+  def callbacks(self):
+    return self._callbacks
 
   @classmethod
   def deserialize(cls, data: dict) -> LiquidHandler:
--- a/pylabrobot/liquid_handling/liquid_handler.py
+++ b/pylabrobot/liquid_handling/liquid_handler.py
@@ -9,7 +9,7 @@ import logging
 import numbers
 import threading
 import time
-from typing import Any, Callable, Dict, Union, Optional, List, Sequence, Set, Tuple
+from typing import Any, Callable, Dict, Union, Optional, List, Sequence, Set, Tuple, Protocol
 import warnings
 
 from pylabrobot.machine import MachineFrontend, need_setup_finished
@@ -74,6 +74,7 @@ class LiquidHandler(MachineFrontend):
 
     self.backend: LiquidHandlerBackend = backend
     self._picked_up_tips96: Optional[TipRack] = None # TODO: replace with tracker.
+    self._callbacks: Dict[str, OperationCallback] = {}
 
     self.deck = deck
     self.deck.resource_assigned_callback_callback = self.resource_assigned_callback
@@ -1383,6 +1384,21 @@ class LiquidHandler(MachineFrontend):
     with open(path, "w", encoding="utf-8") as f:
       json.dump(self.serialize(), f, indent=2)
 
+  def register_callback(self, method_name: str, callback: OperationCallback):
+    """Registers a callback for a specific method."""
+    if self._callbacks.get(method_name):
+      error_message = f"Callback already registered for: {method_name}"
+      raise RuntimeError(error_message)
+    self._callbacks[method_name] = callback
+
+  def _trigger_callback(self, method_name: str, *args, **kwargs):
+    """Triggers the callback associated with a method, if any."""
+    error = kwargs.pop("error", None)
+    if callback := self._callbacks.get(method_name):
+      callback(*args, error=error, **kwargs)
+    elif error:
+      raise error
+
   @classmethod
   def deserialize(cls, data: dict) -> LiquidHandler:
     """ Deserialize a liquid handler from a dictionary.
@@ -1405,3 +1421,8 @@ class LiquidHandler(MachineFrontend):
 
     with open(path, "r", encoding="utf-8") as f:
       return cls.deserialize(json.load(f))
+
+
+class OperationCallback(Protocol):
+  def __call__(self, handler: "LiquidHandler", *args: Any, **kwargs: Any) -> None:
+    ...  # pragma: no cover
--- a/pylabrobot/liquid_handling/liquid_handler.py
+++ b/pylabrobot/liquid_handling/liquid_handler.py
@@ -678,12 +678,19 @@ class LiquidHandler(MachineFrontend):
 
     try:
       await self.backend.aspirate(ops=aspirations, use_channels=use_channels, **backend_kwargs)
-    except:
+    except Exception as error:  # pylint: disable=broad-exception-caught
       for op in aspirations:
         if not op.resource.tracker.is_disabled:
           op.resource.tracker.rollback()
           op.tip.tracker.rollback()
-      raise
+      self._trigger_callback(
+        "aspirate",
+        liquid_handler=self,
+        operations=aspirations,
+        use_channels=use_channels,
+        error=error,
+        **backend_kwargs,
+      )
     else:
       for op in aspirations:
         if not op.resource.tracker.is_disabled:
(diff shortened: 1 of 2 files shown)


ecalifornica:
This test succeeds intermittently for me locally. My brief investigation suggests that it's failing regularly, but that the tests sometimes finish before the async code raises? I haven't had time to look into it yet.

https://github.com/PyLabRobot/pylabrobot/actions/runs/7363549130/job/20043111792#step:6:397

pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -1383,6 +1399,21 @@ def save(self, path: str):
     with open(path, "w", encoding="utf-8") as f:
       json.dump(self.serialize(), f, indent=2)
 
+  def register_callback(self, method_name: str, callback: OperationCallback):
+    """Registers a callback for a specific method."""
```
rickwierenga, line 1403: I like that you decided to use strings to identify callbacks, simplifies things a lot.

I think this method should check that the method_name is allowed (`method_name in {'aspirate', 'dispense', etc.}`.

pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -1383,6 +1399,21 @@ def save(self, path: str):
...
+    if self._callbacks.get(method_name):
+      error_message = f"Callback already registered for: {method_name}"
+      raise RuntimeError(error_message)
+    self._callbacks[method_name] = callback
+
+  def _trigger_callback(self, method_name: str, *args, **kwargs):
+    """Triggers the callback associated with a method, if any."""
```
rickwierenga, line 1410: Could you document that if a callback method exists, the error will be passed there instead of being raised?

pylabrobot/liquid_handling/liquid_handler_tests.py
```diff
@@ -569,3 +569,42 @@ async def test_save_state(self):
...
+
+  def test_duplicate_register_callback(self):
+    self.lh.register_callback("test_duplicate", self.callback)
+    with pytest.raises(RuntimeError):
+      self.lh.register_callback("test_duplicate", self.callback)
+
+  def test_trigger_callback_without_error(self):
+    self.lh.register_callback("test_operation_without_error", self.callback)
+    self.lh.trigger_callback("test_operation_without_error", self.lh)
+    self.callback.assert_called_once()
```
rickwierenga, line 602: Very nice!

pylabrobot/liquid_handling/liquid_handler_tests.py
```diff
@@ -569,3 +569,42 @@ async def test_save_state(self):
...
+
+  def test_trigger_callback_not_found_with_error(self):
+    with pytest.raises(RuntimeError):
+      self.lh.trigger_callback(
+        "test_callback_not_registered_with_error",
+        liquid_handler=self.lh,
+        error=RuntimeError("test"),
+      )
```
rickwierenga, line 610: I think we should also test passing an error when a handler does exist.

pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -1383,6 +1399,21 @@ def save(self, path: str):
...
+    error = kwargs.pop("error", None)
+    if callback := self._callbacks.get(method_name):
+      callback(*args, error=error, **kwargs)
+    elif error:
```
rickwierenga, line 1414: I generally try to explicitly use `is None` in PLR. Perhaps it's a personal quirk, but I like to be consistent :)

rickwierenga (commented):
Great work! This feature will be very useful.

Just added a couple of suggestions and questions. If any to them sound bad to you, I'd be very happy to discuss!

rickwierenga:
> This test succeeds intermittently for me locally. My brief investigation suggests that it's failing regularly, but that the tests sometimes finish before the async code raises? I haven't had time to look into it yet.

This is likely a threading issue in the simulator backend, that sometimes occurs. I'll have to fix that soon, so for now feel free to ignore it...

pylabrobot/liquid_handling/liquid_handler_tests.py
```diff
@@ -569,3 +569,42 @@ async def test_save_state(self):
     self.assertEqual(well_a2.tracker.liquids, [(None, 10)])
 
     set_volume_tracking(enabled=False)
+
+
+class LiquidHandlerForTesting(LiquidHandler):
+  @property
+  def callbacks(self):
+    return self._callbacks
```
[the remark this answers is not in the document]
ecalifornica: I don't see why not, added it.

pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -1383,6 +1399,21 @@ def save(self, path: str):
...
+  def register_callback(self, method_name: str, callback: OperationCallback):
+    """Registers a callback for a specific method."""
+    if self._callbacks.get(method_name):
+      error_message = f"Callback already registered for: {method_name}"
+      raise RuntimeError(error_message)
+    self._callbacks[method_name] = callback
+
+  def _trigger_callback(self, method_name: str, *args, **kwargs):
+    """Triggers the callback associated with a method, if any."""
+    error = kwargs.pop("error", None)
```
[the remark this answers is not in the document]
ecalifornica: Yep!

pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -1383,6 +1399,21 @@ def save(self, path: str):
     with open(path, "w", encoding="utf-8") as f:
       json.dump(self.serialize(), f, indent=2)
 
+  def register_callback(self, method_name: str, callback: OperationCallback):
+    """Registers a callback for a specific method."""
+    if self._callbacks.get(method_name):
```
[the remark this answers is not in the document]
ecalifornica: Agreed, not sure why I wrote it the other way.

pylabrobot/liquid_handling/liquid_handler.py
ecalifornica: Let me know if you like the implementation, and if the list of allowed method names is correct.

pylabrobot/liquid_handling/liquid_handler.py
ecalifornica: Added a note in the docstring, is that what you meant? Or would you like me to update the documentation?

pylabrobot/liquid_handling/liquid_handler_tests.py
ecalifornica: Agreed, not sure how I missed this.

pylabrobot/liquid_handling/liquid_handler.py
ecalifornica: I think `is None` is a totally Pythonic. It is late here, I'm missing something obvious :p

pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -1405,3 +1436,8 @@ def load(cls, path: str) -> LiquidHandler:
 
     with open(path, "r", encoding="utf-8") as f:
       return cls.deserialize(json.load(f))
+
+
+class OperationCallback(Protocol):
+  def __call__(self, handler: "LiquidHandler", *args: Any, **kwargs: Any) -> None:
+    ...  # pragma: no cover
```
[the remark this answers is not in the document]
ecalifornica: I opted for `Protocol` because callbacks feel like their definition might expand over time. Yes, in my understanding they're similar to Swift, or Go's interfaces. `Callable` is simpler and more explicit, happy to change it.

ecalifornica:
Thanks for the great feedback!

Once you've approved the list of allowed methods I'll add the trigger to those. Where in the docs do you think a page on using callbacks should live?

rickwierenga:
I think 'convenience' methods like `return_tips` and `move_plate`, that are essentially just thin wrappers around other methods like `drop_tips` and `move_resource`, should not have callbacks: the callback function will be called using the 'slightly-lower-level' method, so it will have get those kwargs, and so if the user wanted to retry they'd need to call this 'slightly-lower-level' method. Additionally, when logging/collecting analytics I doubt people are interested in which exact method called this, since it really doesn't matter and they could refactor their code to *not* use these methods anytime.

Do you think there's an argument for keeping those callbacks?

pylabrobot/liquid_handling/liquid_handler.py
rickwierenga: The ALLOWED_CALLBACKS constant is great!

pylabrobot/liquid_handling/liquid_handler.py
rickwierenga: Perhaps it's an issue on my side, but I still see `elif error:` :)

pylabrobot/liquid_handling/liquid_handler_tests.py
```diff
@@ -569,3 +569,56 @@ async def test_save_state(self):
...
+    self.lh.register_callback("test_operation_without_error", self.callback)
+    self.lh.trigger_callback("test_operation_without_error", self.lh)
+    self.callback.assert_called_once()
+
+  def test_trigger_callback_with_error_raised(self):
+    with pytest.raises(RuntimeError):
+      self.lh.trigger_callback(
+        "test_operation",
+        error=RuntimeError("test")
+      )
```
rickwierenga, line 612: Shouldn't this test first register a callback, then assert no error is raised when triggering it? Currently it's equivalent to `test_trigger_callback_not_found_with_error`.

pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -1427,7 +1427,7 @@ def _trigger_callback(self, method_name: str, *args, error: Optional[Exception]
     NB: If an error exists it will be passed to the callback instead of being raised.
     """
     if callback := self._callbacks.get(method_name):
-      callback(*args, error=error, **kwargs)
+      callback(self, *args, error=error, **kwargs)
```
ecalifornica, line 1430: I forgot to pass the liquid handler to the callback.

pylabrobot/liquid_handling/liquid_handler_tests.py
ecalifornica: Whoops!

pylabrobot/liquid_handling/liquid_handler.py
ecalifornica: I'm not confident that my update here is what you're looking for, let me know.

ecalifornica:
> I think 'convenience' methods like `return_tips` and `move_plate`, that are essentially just thin wrappers around other methods like `drop_tips` and `move_resource`, should not have callbacks: the callback function will be called using the 'slightly-lower-level' method, so it will have get those kwargs, and so if the user wanted to retry they'd need to call this 'slightly-lower-level' method. Additionally, when logging/collecting analytics I doubt people are interested in which exact method called this, since it really doesn't matter and they could refactor their code to _not_ use these methods anytime.
> 
> Do you think there's an argument for keeping those callbacks?

Makes sense, removed the wrapper methods, let me know how the list looks now.

pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -1435,7 +1428,7 @@ def _trigger_callback(self, method_name: str, *args, error: Optional[Exception]
     """
     if callback := self._callbacks.get(method_name):
       callback(*args, error=error, **kwargs)
-    elif error:
+    elif callback is None and error:
```
rickwierenga, line 1431: I'd say `elif error is not None` (callback should be established by the if statement above).

rickwierenga:
Looking good!

Now it should just be a matter of adding the callbacks to each method in `ALLOWED_CALLBACKS` :)

pylabrobot/liquid_handling/liquid_handler.py
ecalifornica: Agreed, I was attempting to shoehorn in `is None` :p 

> I generally try to explicitly use `is None` in PLR. Perhaps it's a personal quirk, but I like to be consistent :)

I'm still missing the obvious it seems, and am unable to understand what you're asking for here. Please rewrite this to include the `is None` as you see fit.

rickwierenga:
Sweet, edits were enabled. I should have checked before.

Great work on this PR! Good implementation and adds a lot of value!

ecalifornica:
Awesome! Thanks for the guidance :)

rickwierenga:
Thank *you*!

PR #58, opened 2024-01-04
Establishing Revvity resource folder (+ proposing readme files for each resource folder)

BioCam:
I have added a new manufacturer, Revvity (formerly PerkinElmer) into the resource folder with an initial definition for the Revvity_ProxiPlate_384Plus.

I also added a README.md file for this folder specifically that explains what resources from this manufacturer are currently defined in PLR, and gives the part number and website for accelerated ordering. I hope this helps navigate the resource codebase as PLR scales.

I also added an initial new MFXCarrier base class to start building the concept multiflex/multipurpose carriers.


pylabrobot/resources/carrier.py
```diff
@@ -200,6 +200,22 @@ def __init__(
...
+  """ Base class for multiflex carriers 
+  (i.e. carriers with mixed-use and/or specialized sites). """
+  def __init__(
+    self,
+    name: str,
+    size_x: float,
+    size_y: float,
+    size_z: float,
+    sites: Optional[List[CarrierSite]] = None,
+    category="plate_carrier",
```
rickwierenga, line 213: Should this be mfx_carrier or similar?

rickwierenga:
This is awesome!

pylabrobot/resources/revvity/plates.py
```diff
@@ -0,0 +1,43 @@
+""" Revvity plates """
+
+# pylint: disable=invalid-name
+
+from pylabrobot.resources.plate import Plate
+from pylabrobot.resources.well import Well, WellBottomType
+from pylabrobot.resources.itemized_resource import create_equally_spaced
+
+
+# TODO: update heights for volume calculations
```
rickwierenga, line 10: Is the current implementation reasonably accurate? Otherwise it might be best to throw a NotImplementedError imo

pylabrobot/resources/carrier.py
BioCam: Yes, you are absolutely right. Missed that change. Implemented it in the last update.

pylabrobot/resources/revvity/plates.py
BioCam: The plate definition is reasonably accurate, and I've used it quite a bit but the volume calculation is not yet. So I changed the volume calculation method to raise a NotImplementedError as you suggested. Thank you for checking.

rickwierenga: Makes sense!

rickwierenga:
Thanks again!

PR #59, opened 2024-01-13
New resource category: Corning-Axygen, definition of 24-DWP; new container volume calculation functions, enabling offsets in single well

BioCam:
I have added:

1. a **new resource folder to cover corning_axygen** as a manufacturer, which includes the definition and documentation for a 24-deep well plate (max container working volume = 10 ml)
2. Created a new set of **functions for calculating volumes of containers** which only require information available from the container definitions. At the moment this includes calculations for containers with 
     a. number of segments: 2, cross-section: square, bottom: V
     b. number of segments: 2, cross-section: round, bottom: V
     c. number of segments: 2, cross-section: round, bottom: U
These can be used to more dynamically and accurately calculate a resources liquid height (as implemented for the above 24-DWP definition).
3. Currently when one tries to pipette out of a container/well that is less than 2*9mm wide (y-dimension) a `ValueError(f"Resource is too small to space {n} channels evenly.")` is triggered. 
This is because the current `resource.py` code divides the `container size y` by `n+1` channels (for even spacing, I believe). Since 1000ul channels have to be further than 9mm apart from each other due to their size this makes sense for pipetting from large containers with multiple channels.
However, channel `offsets` should **enable pipetting from an offset position inside individual wells** too.
Adding a small condition to the offset enabled me to do this (currently, when using a single channel for pipetting). 

WARNING: the way that offsets are currently written, they appear to relate to the origin of the container (front-left) - not to the 2D center of the container!

PLR_container_offset_explainer

- For example, `offsets =[Coordinate(x=1, y=1, z=1)]` will pipette into the container 1mm from the origin. If the bottom is V or U shaped this will likely crash into the slanted side of the container wall.
- For example, `offsets =[Coordinate(x=-1, y=0, z=0)]` will crash into the plastic wall to the left of the container.
- For example, `offsets =[Coordinate(x=-5, y=3, z=0)]` will pipette into the container to the left of the one intended.


BioCam:
Modification to the pull request:
- removed all code regarding point 3. "enable pipetting from an offset position inside individual wells" because there were other dependencies of the code modified that are crucial to PLR function.
A simple solution to still using the offsets is to create them yourself as shown in the examples above (e.g. `offsets =[Coordinate(x=1, y=1, z=1)]`) instead of deducing them from the 2D_center.

- in regards to point 1. "new resource folder to cover corning_axygen", I had to modify the `Axy_24_DW_10ML` definition to equally_space a `Well` and not a `Container` because the `Plate()` base class requires wells. I initially modified this to take a `Container` so that simulator reflects the square cross-section of the physical plate (wells are automatically shown as round) but this doesn't change anything to the actual utility of the labware definition when used in the physical lab.

pylabrobot/resources/__init__.py
```diff
@@ -25,10 +25,13 @@
...
+from .corning_axygen import *
 from .revvity import *
 
 # liquid handling companies
 from .hamilton import HamiltonDeck, STARLetDeck, STARDeck
 from .ml_star import *
 from .opentrons import *
 from .tecan import *
+
+# labware made from 3rd parties that share their designs with PLR
```
rickwierenga, line 37: do you want this to be a part of this pr?

pylabrobot/resources/resource.py
```diff
@@ -285,6 +285,11 @@ def get_2d_center_offsets(self, n: int = 1) -> List[Coordinate]:
 
     dx = self.get_size_x() / 2
     dy = self.get_size_y() / (n+1)
+    # if n > 1:
+    #   dy = self.get_size_y() / (n+1)
+    # else:
+    #   dy = self.get_size_y()
```
rickwierenga, line 291: Is this still needed?

pylabrobot/resources/volume_functions.py
```diff
@@ -0,0 +1,151 @@
...
+  Returns:
+  float: The volume of the liquid in cubic millimeters.
+
+  The function calculates the volume based on whether the liquid height is within
+  the pyramid or extends into the cuboid.
+  """
+  if liquid_height > z_pyramid+z_cube:
+    print("""WARNING: Liquid overflow detected;
+    check your labware definiton and/or that you are using the right labware.""")
+    return 0.0
```
rickwierenga, line 34: Should this raise an error?

pylabrobot/resources/corning_axygen/plates.py
```diff
@@ -0,0 +1,57 @@
...
+      dy=5.8,
+      dz=2.24,
+      item_dx= 18,
+      item_dy=18,
+      size_x=17.0,
+      size_y=17.0,
+      size_z=42,
+      # bottom_type=WellBottomType.V,
+      # compute_volume_from_height=_compute_volume_from_height_Axy_24_DW_10ML,
+    ),
```
rickwierenga, line 46: Is this no longer needed?

pylabrobot/resources/volume_functions.py
```diff
@@ -0,0 +1,151 @@
...
+  """
+  Calculate the volume of liquid in a container consisting of an upside-down
+  square pyramid at the bottom and a cuboid on top. The container has the 
+  same x and y dimensions for both the pyramid and the cuboid.
+     
+  Parameters:
+  x (float): The base length of the square pyramid and cube in mm.
+  y (float): The base width of the square pyramid and cube in mm.
+  z_pyramid (float): The height of the square pyramid in mm.
+  z_cube (float): The height of the cube in mm.
```
rickwierenga: Would it make sense to rename z_cube to h_cube to be consistent with h_cone below?

pylabrobot/resources/volume_functions.py
```diff
@@ -0,0 +1,151 @@
...
+    check your labware definiton and/or that you are using the right labware.""")
+    return 0.0
+
+  # Calculating the full volume of the hemisphere
+  full_hemisphere_volume = (2/3) * math.pi * r**3
+
+  if liquid_height <= r:
+    # Liquid height is within the hemisphere
+    # Calculating the sub-volume of the hemisphere using spherical cap volume formula
+    h = r - (r - liquid_height)  # Height of the spherical cap
```
rickwierenga, line 141: isn't this just liquid_height?

pylabrobot/resources/corning_axygen/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+Company page: [Corning - Axygen® Brand Products](https://www.corning.com/emea/en/products/life-sciences/resources/brands/axygen-brand-products.html)
+
+> Corning acquired Axygen BioScience, Inc. and its subsidiaries in 2009. This acquisition included Axygen's broad portfolio of high-quality plastic consumables, liquid handling products, and bench-top laboratory equipment, which complemented and expanded Corning's offerings in the life sciences segment​.
+
+### Currently defined labware:
+
+| Description               | Image              |
+|--------------------|--------------------|
+| 'Axy_24_DW_10ML'<br>Part no.: P-DW-10ML-24-C-S<br>[manufacturer website](https://ecatalog.corning.com/life-sciences/b2b/UK/en/Genomics-&-Molecular-Biology/Automation-Consumables/Deep-Well-Plate/Axygen%C2%AE-Deep-Well-and-Assay-Plates/p/P-DW-10ML-24-C-S) | <img src="ims/axygen_Axy_24_DW_10ML.jpg" alt="Axy_24_DW_10ML" width="250"/> |
```
rickwierenga, line 12: Again, love the READMEs!

rickwierenga (commented):
This is really good. Thank you!

Just made a couple of small formatting changes and added some comments to the code.

It'd be good to consider whether the volume functions should throw errors if the height exceeds the container height. This would allow you to skip the check in the well volume function. (speaking of, the well volume function could be a `lambda h: other_vol_func(h, param1, param2` if you want)

rickwierenga:
> Currently when one tries to pipette out of a container/well that is less than 2*9mm wide (y-dimension) a ValueError(f"Resource is too small to space {n} channels evenly.") is triggered.
This is because the current resource.py code divides the container size y by n+1 channels (for even spacing, I believe). Since 1000ul channels have to be further than 9mm apart from each other due to their size this makes sense for pipetting from large containers with multiple channels.
However, channel offsets should enable pipetting from an offset position inside individual wells too.
Adding a small condition to the offset enabled me to do this (currently, when using a single channel for pipetting).

I think this is a hardware constraint on the STAR.

> removed all code regarding point 3. "enable pipetting from an offset position inside individual wells" because there were other dependencies of the code modified that are crucial to PLR function.
A simple solution to still using the offsets is to create them yourself as shown in the examples above (e.g. offsets =[Coordinate(x=1, y=1, z=1)]) instead of deducing them from the 2D_center.

`get_2d_center_offsets` returns a list of coordinates. You can add coordinates, like

```py
center = well.get_2d_center_offsets(n=1)[0]
my_offset = Coordinate(1, 1, 1)
true_offset = center + my_offset
```

Does that help?

> in regards to point 1. "new resource folder to cover corning_axygen", I had to modify the Axy_24_DW_10ML definition to equally_space a Well and not a Container because the Plate() base class requires wells. I initially modified this to take a Container so that simulator reflects the square cross-section of the physical plate (wells are automatically shown as round) but this doesn't change anything to the actual utility of the labware definition when used in the physical lab.

Let's add a parameter to Well to optionally make the square instead of circular. There is a part where we infer the max volume if the volume height func is missing by assuming the well is circular, which we should then check for.

On the simulator, we'll adjust rendering.

> WARNING: the way that offsets are currently written, they appear to relate to the origin of the container (front-left) - not to the 2D center of the container!

On STAR, I think the `_ops_to_fw_positions` method should add the offset per operation to the well center (line ~317). Only when you specify a single resource, LiquidHandler will infer you want to aspirate from the 2d center, which can be overridden (not adjusted) by specifying offsets. Is that not what happens? If it is confusing, I'm happy to discuss alternatives.

Btw, this graphic is amazing! With your permission, I'd like this one to go on the docs website as well!

pylabrobot/resources/volume_functions.py
BioCam: Yes, you are completely correct. This is an undesired redundancy and we should change this `if` statement to the simpler:

```
if liquid_height <= r:
    # Liquid height is within the hemisphere
    # Calculating the sub-volume of the hemisphere using spherical cap volume formula
    liquid_volume = (1/3) * math.pi * liquid_height**2 * (3*r - liquid_height)
```

Very nice catch! :)

pylabrobot/resources/volume_functions.py
BioCam: Yes, that is a very good idea. I have changed this now.

pylabrobot/resources/corning_axygen/plates.py
BioCam: I initially commented it out because these arguments didn't work with the `Container` class I used to achieve the square rendering in the simulator.
If we modify the `Well` class with a e.g. `cross_section` = `square` or `round` argument then it would make sense to uncomment these arguments again - just done that.

pylabrobot/resources/volume_functions.py
BioCam: Initially it did but the automated type hinting testing flagged it as an error because the function hints at a float as the return.
So I could either remove the type hinting, or write a warning message with a float return.
I liked the later more as I find type hinting more useful, and the chance that the `liquid_height` is actually larger than the container it is in is very small (only possible if the user by accident takes the wrong labware) and the consequences of that - in this case here - are not very dangerous to the machine.

pylabrobot/resources/resource.py
BioCam: No, you are right. Your answer showed nicely how this approach is not needed and will unlikely be needed in the future - I just removed these lines.

pylabrobot/resources/__init__.py
BioCam: Yes, just to get the DIY of labware and their PLR definitions sharing started :)

BioCam:
Yes:
```
center = well.get_2d_center_offsets(n=1)[0]
my_offset = Coordinate(1, 1, 1)
true_offset = center + my_offset
```
...this makes a lot of sense and is a very easy fix. I like it a lot. I also think that the offsets from origin I mentioned in my infographic is very logical. I do think though that we need more documentation to make this easily understandable to others.

"Let's add a parameter to Well to optionally make the square instead of circular."
"On the simulator, we'll adjust rendering."
That would be amazing, thank you! Please let me know whether I can help with that. To be honest I haven't looked too much into how the simulator backend works yet.

"Only when you specify a single resource, LiquidHandler will infer you want to aspirate from the 2d center, which can be overridden (not adjusted) by specifying offsets. Is that not what happens?"
I don't think that is what happens:
For example,
```
await lh.aspirate(Axy_24_DW_10ML['D3'], vols=[2], 
                  swap_speed = 50, homogenization_speed = 20,
                  dispensation_speed_during_emptying_tip = 25,
                 )
```
...will aspirate from the center of the well.

However,
```
await lh.aspirate(Axy_24_DW_10ML['D6'], vols=[2], 
                  swap_speed = 50, homogenization_speed = 20,
                  dispensation_speed_during_emptying_tip = 25,
                  offsets=[Coordinate(x=2.6, y=2.6, z=3]
                 )
```
...will aspirate at those coordinates relative to the well origin, not relative to the well center bottom.
But you showed at the top how an offset in regards to the well center bottom can be achieved quickly in your comment above.


"Btw, this graphic is amazing! With your permission, I'd like this one to go on the docs website as well!"

Thank you, and yes, of course. I would be happy to see the graphic on the documentation :)

pylabrobot/resources/volume_functions.py
rickwierenga: With a `raise`, you shouldn't need a `return`. I'm pretty sure the hinting is ok with that.

rickwierenga:
Thanks for the changes and responses!

> I do think though that we need more documentation to make this easily understandable to others.

On it. I will expand the resources tutorial based on all of your feedback (and with your amazing graphics!) very soon.

> That would be amazing, thank you! Please let me know whether I can help with that. To be honest I haven't looked too much into how the simulator backend works yet.

Added the CrossSectionType with https://github.com/PyLabRobot/pylabrobot/commit/a99e9f99c830335f99d8b8109ca28a82950a07a2 on main.

I'd be happy to update the simulator, but also wouldn't want to rob you of the opportunity to check that out. Just let me know what you want.

> I don't think that is what happens:

There is a check for `offset is None` on STAR, but that is overridden when the resource type is a well (or a tipspot) (L316). Could you please confirm you still observe this behavior after updating the types from Container to Well?

Also, added `get_2d_center_offset` (singular) for convenience https://github.com/PyLabRobot/pylabrobot/commit/27dc83a71fb767d2825d92d11d7aa1eb941b8ff4.

BioCam:
"Added the CrossSectionType with [a99e9f9](https://github.com/PyLabRobot/pylabrobot/commit/a99e9f99c830335f99d8b8109ca28a82950a07a2) on main."
- This is perfect, thank you.

"I'd be happy to update the simulator,"
- Thank you for offering to solve this for us. If you find some time, could you please do this. I think it will take you less time, and I have a couple other open-source items I am working on at the moment (I will still have a peek at your mods on GitHub afterwards to learn more about how the simulator works :eyes:). Also since the last update my simulator seems to be misbehaving, it doesn't update labware (after assigning them in the Notebook) and doesn't close when the kernel is restarted. Depending on whether this persists on my side I might raise an issue tomorrow.

rickwierenga:
You got it: https://github.com/PyLabRobot/pylabrobot/commit/7895acbef645afb5fb8fb99516fb855d5db1f25e

Please let me know if you're able to reproduce the issues with the simulator.

rickwierenga:
I think the last todo for this PR is deciding whether the volume functions should raise errors. Assuming the type checker supports it, I think doing so makes a lot of sense and enables a simpler use of these volume functions in wells.

BioCam:
"I think the last todo for this PR is deciding whether the volume functions should raise errors. Assuming the type checker supports it, I think doing so makes a lot of sense and enables a simpler use of these volume functions in wells."
- Yes, I agree. I have updated all volume functions to raise a ValueError to finish this PR.

"You got it: [7895acb](https://github.com/PyLabRobot/pylabrobot/commit/7895acbef645afb5fb8fb99516fb855d5db1f25e)"
- This is amazing, thank you!

"Please let me know if you're able to reproduce the issues with the simulator."
- I'll do some more testing on the simulator issues and will let you know how it goes but I think that is a different topic and we can close this PR. There is still something to be done with `offsets = plate.get_item('A1').get_2d_center_offsets(n=1)` because this will raise an error due to the well size y divided by n+1 (here 2) being smaller than 9mm. But I think that should be dealt with separately?

rickwierenga:
Great! Thanks for your work 🔥

Let's discuss `plate.get_item('A1').get_2d_center_offsets(n=1)` in a separate thread.

rickwierenga:
btw, 

> because this will raise an error due to the well size y divided by n+1 (here 2) being smaller than 9mm.

https://github.com/PyLabRobot/pylabrobot/commit/c634e8f4e6303e9695a542995d16318093066a51 destupified the implementation and fixed it

PR #57, opened 2024-01-02
Enabling use of TIP_50ul (_L & -P; Hamilton, conductive, with filter)

BioCam:
PLR has so far been missing the definitions of (standard, conductive) Hamilton 50ul tips with filter (cat. no.: 235948) as a `resource`.

This update integrates their definition in the form of...
- a new tip_creator: `fifty_ul_tip_with_filter()` - including the required 
    - `TipSize`: `FIFTY_UL=6` and 
    - `fitting_depth[FIFTY_UL] = 8`,
- tipracks `TIP_50ul_L` & `TIP_50ul_P`

Experimentally verified on a Hamilton `STARletDeck` with a `TIP_CAR_480_A00` carrier/template.

PR #63, opened 2024-01-24
Correcting trash z height for discarding tips

BioCam:
I have updated the height at which tips are discarded above the trash to allow tips to gently sit but not crash on the trash while being discarded.

Furthermore, made some small adjustments to definitions and corrected an existing autoload command.


rickwierenga:
Did you confirm this z height works better for tips of different lengths as well? (for future reference: https://forums.pylabrobot.org/t/pipette-crunching-during-tip-discarding-plr/)

BioCam:
Good point; I just tested it with `HTF_L` tips and for these tips the ejection height is now way too high.

All I changed is the suggested z-location on line 65 of hamilton_decks.py which works very well now for 50 ul tips (at least 20 mm).

How is the discard/ejection z height calculated for different tips?

rickwierenga:
https://github.com/PyLabRobot/pylabrobot/blob/main/pylabrobot/liquid_handling/backends/hamilton/STAR.py#L1351

`end_tip_deposit_process` is where it drops the tip, `begin_tip_deposit_process` is where it starts moving more slowly (between that and `minimum_traverse_height_at_beginning_of_a_command` it moves fast).

BioCam:
Testing:
- I have changed the trash z height back to its original value, and 50 ul tips are getting crunched again.
- Since I have never not yet used the 1000ul tips with PLR before, I tried them out as well. Even with the original value for the trash height they are being discarded at least 15 mm above the trash, i.e. without any change made in this PR.

rickwierenga:
Could you get two `C0TR...` logs for the 50/1000uL tip discards from VENUS? (I'd do it if I had access to a machine)

BioCam:
I can tomorrow.
But regarding PLR's discard_ti:

```
async def discard_tip(
    self,
    x_positions: List[int],
    y_positions: List[int],
    tip_pattern: List[bool],
    begin_tip_deposit_process: int = 0,
    end_tip_deposit_process: int = 0,
    minimum_traverse_height_at_beginning_of_a_command: int = 3600,
    z_position_at_end_of_a_command: int = 3600,
    discarding_method: TipDropMethod = TipDropMethod.DROP
  ):
    """ discard tip

    Args:
      x_positions: x positions [0.1mm]. Must be between 0 and 25000. Default 0.
      y_positions: y positions [0.1mm]. Must be between 0 and 6500. Default 0.
      tip_pattern: Tip pattern (channels involved). Must be between 0 and 1. Default 1.
      begin_tip_deposit_process: Begin of tip deposit process (Z- range) [0.1mm]. Must be between
          0 and 3600. Default 0.
      end_tip_deposit_process: End of tip deposit process (Z- range) [0.1mm]. Must be between 0
          and 3600.
      minimum_traverse_height_at_beginning_of_a_command: Minimum traverse height at beginning of a
          command 0.1mm] (refers to all channels independent of tip pattern parameter 'tm'). Must
          be between 0 and 3600.
      z-position_at_end_of_a_command: Z-Position at end of a command [0.1mm].
          Must be between 0 and 3600.
      discarding_method: Pick up method Pick up method. 0 = auto selection (see command TT
          parameter tu) 1 = pick up out of rack. 2 = pick up out of wash liquid (slowly). Must be
          between 0 and 2.

    If discarding is PLACE_SHIFT (0), tp/ tz = tip cone end height.
    Otherwise, tp/ tz = stop disk height.
    """

    assert all(0 <= xp <= 25000 for xp in x_positions), "x_positions must be between 0 and 25000"
    assert all(0 <= yp <= 6500 for yp in y_positions), "y_positions must be between 0 and 6500"
    assert 0 <= begin_tip_deposit_process <= 3600, \
      "begin_tip_deposit_process must be between 0 and 3600"
    assert 0 <= end_tip_deposit_process <= 3600, \
      "end_tip_deposit_process must be between 0 and 3600"
    assert 0 <= minimum_traverse_height_at_beginning_of_a_command <= 3600, \
      "minimum_traverse_height_at_beginning_of_a_command must be between 0 and 3600"
    assert 0 <= z_position_at_end_of_a_command <= 3600, \
      "z_position_at_end_of_a_command must be between 0 and 3600"

    return await self.send_command(
      module="C0",
      command="TR",
      tip_pattern=tip_pattern,
      fmt="kz### (n)vz### (n)",
      xp=[f"{x:05}" for x in x_positions],
      yp=[f"{y:04}" for y in y_positions],
      tm=tip_pattern,
      tp=begin_tip_deposit_process,
      tz=end_tip_deposit_process,
      th=minimum_traverse_height_at_beginning_of_a_command,
      te=z_position_at_end_of_a_command,
      ti=discarding_method.value,
    )
```

VS


```
async def drop_tips(
    self,
    ops: List[Drop],
    use_channels: List[int],
    drop_method: Optional[TipDropMethod] = None,
  ):
    """ Drop tips to a resource.

    Args:
      drop_method: The method to use for dropping tips. If None, the default method for dropping to
        tip spots is `DROP`, and everything else is `PLACE_SHIFT`. Note that `DROP` is only the
        default if *all* tips are being dropped to a tip spot.
    """

    if drop_method is None:
      if any(not isinstance(op.resource, TipSpot) for op in ops):
        drop_method = TipDropMethod.PLACE_SHIFT
      else:
        drop_method = TipDropMethod.DROP

    x_positions, y_positions, channels_involved = \
      self._ops_to_fw_positions(ops, use_channels)

    # get highest z position
    max_z = max(op.resource.get_absolute_location().z + \
                (op.offset.z if op.offset is not None else 0) for op in ops)
    max_total_tip_length = max(op.tip.total_tip_length for op in ops)
    max_tip_length = max((op.tip.total_tip_length-op.tip.fitting_depth) for op in ops)

    try:
      return await self.discard_tip(
        x_positions=x_positions,
        y_positions=y_positions,
        tip_pattern=channels_involved,
        begin_tip_deposit_process=int((max_z + max_total_tip_length)*10),
        end_tip_deposit_process=int((max_z + max_tip_length)*10),
        minimum_traverse_height_at_beginning_of_a_command=2450,
        z_position_at_end_of_a_command=2450,
        discarding_method=drop_method
      )
    except STARFirmwareError as e:
      tip_errors: List[int] = []
      for i in range(1, self.num_channels+1):
        channel_error = e.error_for_channel(i)
        if isinstance(channel_error, HamiltonNoTipError):
          tip_errors.append(i-1)

      if len(tip_errors) > 0:
        raise NoTipError(f"No tip present on channels {tip_errors}") from e

      raise e
```

How does `lh.discard_tips()` assign a new value to `begin_tip_deposit_process` and `end_tip_deposit_process`? 
(maybe it's just late here, and I missed it)

BioCam:
I see; then I don't know yet why `lh.discard_tips()` doesn't work for neither the 1000ul tips (discarded too high) nor the 50ul tips (discarded too low).

BioCam:
Here is the trc file log of the `C0TR` commands:

```go
// 1000ul tips
< 08:53:06.386 8AF#8000#00: C0TPid0227xp01179 01179 00000&yp2418 2328 0000&tm1 1 0&tt04tp2264tz2164th2450td1 // pick up 2x 1000ul tips
> 08:53:13.217 8AF#8000#00: C0TPid0227er00/00sx466 466 000 000 000 000 000 000sg490 492 000 000 000 000 000 000
< 08:53:13.232 8AF#8000#00: C0RTid0228
> 08:53:13.264 8AF#8000#00: C0RTid0228er00/00rt1 1 0 0 0 0 0 0
< 08:53:13.310 8AF#8000#00: C0RTid0229
> 08:53:13.326 8AF#8000#00: C0RTid0229er00/00rt1 1 0 0 0 0 0 0
< 08:53:13.369 8AF#8000#00: C0TRid0230xp08000&yp4050 3782 3514 3246 2978 2710 2442 2175tp1970tz1870th2450te2450tm1&ti0  // discard 1000ul tips
> 08:53:20.685 8AF#8000#00: C0TRid0230er00/00kz386 403 000 000 000 000 000 000vz390 406 000 000 000 000 000 000

// 50ul tips
< 08:53:20.685 8AF#8000#00: C0RTid0231
> 08:53:20.716 8AF#8000#00: C0RTid0231er00/00rt0 0 0 0 0 0 0 0
< 08:53:20.755 8AF#8000#00: C0TPid0232xp01179 01179 00000&yp1458 1368 0000&tm1 1 0&tt23tp2244tz2164th2450td1 // pick up 2x 50 ul tips
> 08:53:27.856 8AF#8000#00: C0TPid0232er00/00sx279 279 000 000 000 000 000 000sg304 302 000 000 000 000 000 000
< 08:53:27.856 8AF#8000#00: C0RTid0233
> 08:53:27.887 8AF#8000#00: C0RTid0233er00/00rt1 1 0 0 0 0 0 0
< 08:53:27.918 8AF#8000#00: C0RTid0234
> 08:53:27.950 8AF#8000#00: C0RTid0234er00/00rt1 1 0 0 0 0 0 0
< 08:53:27.981 8AF#8000#00: C0TRid0235xp08000&yp4050 3782 3514 3246 2978 2710 2442 2175tp1970tz1870th2450te2450tm1&ti0  // discard 50ul tips
> 08:53:35.694 8AF#8000#00: C0TRid0235er00/00kz386 403 000 000 000 000 000 000vz389 408 000 000 000 000 000 000
```

I included the tip pickup commands as well to check what is the difference between pick_up_tips (which I see as a drop_tips in reverse), and discard(trash).

A direct comparison:
1000ul pick up
```go
< 08:53:06.386 8AF#8000#00: C0TPid0227xp01179 01179 00000&yp2418 2328 0000&tm1 1 0&tt04
tp2264 // <------ Begin of tip picking up process (Z- range) [0.1mm]
tz2164
th2450
td1 // pick up 2x 1000ul tips
```
50ul pick up
```
< 08:53:20.755 8AF#8000#00: C0TPid0232xp01179 01179 00000&yp1458 1368 0000&tm1 1 0&tt23
tp2244 // <------ Begin of tip picking up process (Z- range) [0.1mm]
tz2164
th2450
td1 // pick up 2x 50 ul tips
```

A direct comparison:
1000ul discard tip
```go
< 08:53:13.369 8AF#8000#00: C0TRid0230xp08000&yp4050 3782 3514 3246 2978 2710 2442 2175
tp1970 // <------- Begin of tip deposit process (Z- discard range) [0.1mm]
tz1870
th2450
te2450
tm1
&
ti0   // <------- discarding method: 0 = place & shift vs 1 = drop (no shift) (default=1)
```

50ul discard tip 
```go
< 08:53:27.981 8AF#8000#00: C0TRid0235xp08000&yp4050 3782 3514 3246 2978 2710 2442 2175
tp1970 // <------- Begin of tip deposit process (Z- discard range) [0.1mm]
tz1870
th2450
te2450
tm1
&
ti0  // <------- discarding method: 0 = place & shift vs 1 = drop (no shift) (default=1)
```

BioCam:
1. discarding_method discrepancy
In comparison, PLR appears to default to a `discarding_method: TipDropMethod = TipDropMethod.DROP` (STAR.py:3422) which comes from tip_creators.py:31...
```python
class TipDropMethod(enum.Enum):
  """ Tip drop method """
  PLACE_SHIFT=0
  DROP=1
```
Which means PLR's `discard_tips(trash)` does default to the tip discard default as intended, but it doesn't seem VENUS actually uses the default itself, and instead performs a place & shift. This raises the question whether we should change the PLR default away from the fw command default because the PLR command is actually meant to imitate the VENUS command in this case?

I would definitely recommend we do this... The DROP command is very violent and shoots the tip at a rapid speed off. And one of the nice Hamilton features is the "soft" tip discard to avoid aerosol formation, i.e. the place & shift.

2. Same tp for differently-sized tips?
I guess this also explains why VENUS' `C0TR` commands also use the same tp/tip deposit process begin height for different tips?

3. PLR discard behaviour
Unfortunately I still cannot see why PLR's current `discard_tips(trash)` is shooting the tips away (a) from too high for 1000ul tips, and (b) from too low for 50ul tips.

rickwierenga:
Great thinking. If you send a firmware command (`STAR.send_raw_command`) with `td0`, does it work as intended? If so, let's use that for tip discards.

BioCam:
Yes, I just tried out...
```python
await lh.backend.send_raw_command('C0TRid0230xp08000&yp4050 3782 3514 3246 2978 2710 2442 2175tp1970tz1870th2450te2450tm1&ti0')
```
...for...
1. discarding 1 or 8 channels with 50ul tips attached, and
2. discarding 1 or 8 channels with 1000ul tips attached.

The exact same command works for all tips (kinda neat).
Also VENUS does not center the tips during discard based on the trash 2D_center. Instead it always uses the same discard positions, located towards the end of the trash (y+ side). I think this makes a lot of sense for safety considerations: just in case, put the tips as far away as possible (while still maintaining proper spacing) from the human operator.

Actionable result:
- It appears that no matter what tip is being used, the above discard command should work very well, and in a no-aerosol releasing fashion. I suggest we default the `lh.discard_tips()` command to this fw command as a quick and direct implementation of the VENUS command. By only changing the `lh.discard_tips()` we leave the `STAR.discard_tips()` method alone (because, as you mentioned, `STAR.drop_tips()` relies on `STAR.discard_tips()` and that works very well).

rickwierenga:
I believe this method of dropping tips probably does not work / is not desirable for dropping tips in tip racks. Since lh.discard_tips should not know about the specific backend it is sending commands to, I think this issue must be addressed on STAR.

As it turns out, in `drop_tips` (the high level STAR command) there already is a check for this (l.1342):

```py
if drop_method is None:
  if any(not isinstance(op.resource, TipSpot) for op in ops):
    drop_method = TipDropMethod.PLACE_SHIFT
  else:
    drop_method = TipDropMethod.DROP
```

Looking at STAR_tests.py, this appears to work correctly? (`ti1` for tip rack, `ti0` for discard) Now I actually remember implementing this >1y ago.

It is confusing to me why this would not work on your setup.

Just to confirm: is the firmware command STAR sends for you `ti1` for the discard / trash drop? If no, how is the command that is sent different from the command VENUS sends `C0TRid0230xp08000&yp4050 3782 3514 3246 2978 2710 2442 2175tp1970tz1870th2450te2450tm1&ti0`?

BioCam:
That is an interesting question.
How can you retrieve the fw string when executing `await lh.discard_tips()`? (it is not send to the code cell output and when stored in a variable the variable returns `None`.

rickwierenga:
It should be saved to a log file. (Alternatively, insert a `print(cmd)` statement in `send_command` in `liquid_handling/backends/hamilton/base.py:190`)

BioCam:
1. Current topic:

This is interesting:
The fw command of `lh.discard_tips()` is actually
```go
C0TRid0008xp08000 08000 00000&yp3514 2710 0000&tm1 1 0&tp2322tz2402th2450te2450ti0
```
For comparison, the functional command mentioned above:
```go 
'C0TRid0230xp08000&yp4050 3782 3514 3246 2978 2710 2442 2175tp1970tz1870th2450te2450tm1&ti0'
```
I am not sure why the first one actually shoots the tip off with vigorous force and the second actually executes the soft discard/place & shift.

2. log files
My log files are always completely empty (Zero bytes files size), except when an error occurs which is then recorded in the log file.
I thought that is known, and maybe desired?

rickwierenga:
### discards

The difference in `tp` is 2322 - 1970 = 352 whereas the difference in `tz` is 2402 - 1870 = 532. The difference in differences is 180. It doesn't ring a bill for me, but experience has shown that computing these deltas and keeping them in the back of my mind is very helpful.

In the documentation for tip discard `TR` it says

```
0 = place & shift          tp/ tz = tip cone end height 
1 = drop (no shift)        tp/ tz = stop disk height
```

It is not entirely clear to me what this means, but based on 1) the difference hinted at, and 2) your observations of fixed values for tp/tz with ti0, I would say that
- for ti0 it uses a 'plunger height',
- for 1, it will use the tip height.

This is opposite of how I would interpret the docs, but that's fine. The difference between tp/tz in ti0/1 is clear.

I think we could update the tp/tz calculations from

```py
max_z = max(op.resource.get_absolute_location().z + \
                (op.offset.z if op.offset is not None else 0) for op in ops)
max_total_tip_length = max(op.tip.total_tip_length for op in ops)
max_tip_length = max((op.tip.total_tip_length-op.tip.fitting_depth) for op in ops)

...

    begin_tip_deposit_process=int((max_z + max_total_tip_length)*10),
    end_tip_deposit_process=int((max_z + max_tip_length)*10),
```

to maybe something like

```py
max_z = max(op.resource.get_absolute_location().z + \
                (op.offset.z if op.offset is not None else 0) for op in ops)
if drop_method == TipDropMethod.PLACE_SHIFT:
  # magic values empirically found in https://github.com/PyLabRobot/pylabrobot/pull/63
  begin_tip_deposit_process  = int((max_z+59.9)*10)
  end_tip_deposit_process  = int((max_z+49.9)*10)
else:
  max_total_tip_length = max(op.tip.total_tip_length for op in ops)
  max_tip_length = max((op.tip.total_tip_length-op.tip.fitting_depth) for op in ops)
  begin_tip_deposit_process=int((max_z + max_total_tip_length)*10),
  end_tip_deposit_process=int((max_z + max_tip_length)*10),

...

    begin_tip_deposit_process= begin_tip_deposit_process,
    end_tip_deposit_process= end_tip_deposit_process,
```

Does that work?

This implementation is bad because there are two unknowns: the height of the trash and the magic 59.9 and 49.9, but probably the best we can do with the current information. When there is another place&shift drop on a non-trash resource, or a drop on the trash, we can pinpoint both values.

Also, which tip rack are you using? For both HTF_L and HT_L I get tp=2322 (same as you), tz=2242.

### logging

The ["level"](https://docs.python.org/3/library/logging.html#levels) of logging can be changed using `logger.setLevel(logging.DEBUG)`. I forgot that I removed this default setting, which has now returned https://github.com/PyLabRobot/pylabrobot/commit/f775675dc22bccfdca78902c305fdacac2f7b8a4.

For future reference, put at any point:

```py
logger = logging.getLogger("pylabrobot")
logger.setLevel(logging.DEBUG) # <-- choose a level
```

BioCam:
Magic indeed: these changes make everything work
commands generated by PLR discarding different numbers and differently-sized tips correctly with place & shift:
```go
#2x50ul: C0TRid0008xp08000 08000 00000&yp3514 2710 0000&tm1 1 0&tp1970tz1870th2450te2450ti0
#8x50ul: C0TRid0016xp08000 08000 08000 08000 08000 08000 08000 08000yp4050 3782 3514 3246 2978 2710 2442 2174tm1 1 1 1 1 1 1 1tp1970tz1870th2450te2450ti0

#2x1000ul: C0TRid0012xp08000 08000 00000&yp3514 2710 0000&tm1 1 0&tp1970tz1870th2450te2450ti0
#8x1000ul: C0TRid0014xp08000 08000 08000 08000 08000 08000 08000 08000yp4050 3782 3514 3246 2978 2710 2442 2174tm1 1 1 1 1 1 1 1tp1970tz1870th2450te2450ti0
```
I am not sure how to modify the unit tests though to update the expected behaviour in them?

rickwierenga:
I have assumed that the trash configuration on our STAR, and all other STARs, is in the same location, and updated the test accordingly. Glad this works now!

Thank you for the quick troubleshooting and detailed reporting!

ben-ray:
I noticed that these changes break define_tip_needle in STAR.py. The tg parameter of command COTT is sent value 6, which corresponds to 6 – CoRe 384 head tip (Hamilton) in the most updated firmware reference guide we are using. I suspect this breaks tip pickups for all STARs running newer firmware versions (8.0 and up). I haven't found a perfect fix yet...

rickwierenga:
You're 100% right. But I think that would be the result of https://github.com/PyLabRobot/pylabrobot/pull/57/files.

I just double checked in VENUS/the "Hamiton Co-Re Liquid Editor" and 50uL tips with filter use tip type 23; without filter it's 22. These default indexes for the tip type (TT) are what venus sends to the machine, so I am able to search our logs for corresponding definitions. Keep in mind that these indices are otherwise arbitrary. Here are two examples from our logs:

- 22: 50uL without filter `C0TTid0134tt22tf0tl0424tv00650tg2tu0`
- 23: 50uL with filter ` C0TTid0145tt23tf1tl0424tv00600tg2tu0`

It appears that tg (tip size) is set to 2 for both. Further, 1) the tip length is 42.4mm, not 50mm; 2) tip volume is 65.0/60.0uL, not 50uL as I initially merged:

```
def fifty_ul_tip_with_filter() -> HamiltonTip:
  """ 50 ul tip with a filter (Hamilton cat. no.: 235948) """
  return HamiltonTip(
    has_filter=True,
    total_tip_length=50,
    maximal_volume=50,
    tip_size=TipSize.FIFTY_UL,
    pickup_method=TipPickupMethod.OUT_OF_RACK
  )
```

Pushed an experimental commit to https://github.com/PyLabRobot/pylabrobot/tree/fix-50uL. @BioCam/@ben-ray could you let me know if this works for you? `git switch fix-50ul` (or use the gui to switch branches)

My bad - I should have caught this before merging.

PR #64, opened 2024-01-25
Exposing autoload carrier load, unload & barcode reading (v1)

BioCam:
I have written code to use the autoload on Hamilton STAR liquid handlers.

I used the iSWAP implementation as inspiration to make this function comprehensive and (in my hands) safe (an autoload or carrier left in the middle of the machine has the potential to damage the machine, users operate on their own responsibility):

1. Check whether autoload is installed during setup. (I couldn't find a specific cue of the machine or direct command that checks this (this could be because autoloads cannot be retrofitted into Hamilton liquid handlers, and therefore no need for checking would VENUS-normally arise). So instead I am using the `I0RO` fw command ("Request download date") to check when the autoload software has been downloaded. If the download has happened after 2015 the machine is considered to have an autoload installed. I will have to receive feedback from other users who do not have an autoload whether this truly recognises the absence of the autoload - similar to how the absence of an iSWAP had to be figured out by a user without an iSWAP :)
2. If an autoload is installed, a check whether the autoload is initialised will be performed.
3. If not initialised, the autoload will initialise.
4. Initialisation leaves the autoload hanging on the wrong end of the robot (i.e. crash danger). Hence, I implemented a "park_autoload" function. However, I couldn't find a direct fw command to simply move the autoload out of the way yet. So instead the fw command `C0CS` ("Check for presence of carriers on loading tray") acts as an intermediate solution because it ends with the autoload parked in a save position.

This enables loading of carriers:
```python
await STAR.load_carrier(tip_carrier_0)
```

This enables unloading of carriers:
```python
await STAR.unload_carrier(tip_carrier_0)
```

Furthermore, the fw command for loading is the main way of quickly identifying the barcodes of your labware.
This is activated using...
```python
await STAR.load_carrier(tip_carrier_0, barcode_reading=True)
```
This function can currently specify further:
```python
barcode_reading_direction: str = 'horizontal',
barcode_symbology: str = 'Code 128 (Subset B and C)',
no_container_per_carrier: int = 5,
park_autoload_after: bool = True
```

I have verified the ability to read barcodes from Hamilton tip carriers and Hamilton tips. 

Plate barcodes cannot be read yet.

After closing this PR I will write a forum post detailing the functionalities for easy adaptation by other users.

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR.py
@@ -1218,14 +1218,19 @@ class STAR(HamiltonLiquidHandler):
     tip_presences = await self.request_tip_presence()
     self._num_channels = len(tip_presences)
 
+    # Request machine information
+    conf = await self.request_machine_configuration()
     extended_conf = await self.request_extended_configuration()
+
     left_x_drive_configuration_byte_1 = bin(extended_conf["xl"])
     left_x_drive_configuration_byte_1 = left_x_drive_configuration_byte_1 + \
       "0" * (16 - len(left_x_drive_configuration_byte_1))
     left_x_drive_configuration_byte_1 = left_x_drive_configuration_byte_1[2:]
+    autoload_configuration_byte = bin(conf['kb']).split('b')[-1][-3]
+    # Identify installations
+    self.autoload_installed = autoload_configuration_byte == "1"
     self.core96_head_installed = left_x_drive_configuration_byte_1[2] == "1"
     self.iswap_installed = left_x_drive_configuration_byte_1[1] == "1"
-    self.autoload_installed = self.request_autoload_installed_after_2015()
 
     initialized = await self.request_instrument_initialization_status()
 
@@ -2635,16 +2640,6 @@ class STAR(HamiltonLiquidHandler):
     resp = await self.send_command(module="C0", command="QW", fmt="qw#")
     return resp is not None and resp["qw"] == 1
 
-  async def request_autoload_installed_after_2015(self) -> bool:
-    """
-    Check whether autoload has been installed after 2015.
-    
-    Convenient substitute for checking whether machine has an autoload.
-    """
-
-    resp = await self.send_command(module="I0", command="RO", fmt="ao####")
-    return resp is not None and resp["ao"] > 2015
-
   async def request_autoload_initialization_status(self) -> bool:
     """ Request autoload initialization status """
 
@@ -3114,7 +3109,8 @@ class STAR(HamiltonLiquidHandler):
     """ Request machine configuration """
 
     # TODO: parse res
-    return await self.send_command(module="C0", command="RM")
+    return await self.send_command(module="C0", command="RM",
+                                   fmt="kb**kp**")
 
   async def request_extended_configuration(self):
     """ Request extended configuration """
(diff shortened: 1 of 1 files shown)


pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -2627,6 +2634,22 @@ async def request_instrument_initialization_status(self) -> bool:
...
+  
+  async def request_autoload_installed_after_2015(self) -> bool:
+    """
+    Check whether autoload has been installed after 2015.
+    
+    Convenient substitute for checking whether machine has an autoload.
+    """
+
+    resp = await lh.backend.send_command(module="I0", command="RO", fmt="ao####")
+    return resp is not None and resp["ao"] > 2015
```
rickwierenga, line 2646: What is special about 2015?

pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -1225,6 +1225,7 @@ async def setup(self):
     left_x_drive_configuration_byte_1 = left_x_drive_configuration_byte_1[2:]
     self.core96_head_installed = left_x_drive_configuration_byte_1[2] == "1"
     self.iswap_installed = left_x_drive_configuration_byte_1[1] == "1"
+    self.autoload_installed = self.request_autoload_installed_after_2015()
```
rickwierenga, line 1228: Should we print a warning if installed before 2015?

rickwierenga (commented):
You're on fire!

pylabrobot/liquid_handling/backends/hamilton/STAR.py
BioCam: Yes, I think this is a great idea.

I have chosen 2015 because it seemed like long enough ago but not too long... if it correctly identifies the presence/absence of an autoload we can make it anything we want.

rickwierenga: Whether autoload is installed should actually be available in the third bit of the `kb` parameter of the response to 'Request machine configuration' (configuration data 1). There is an example for the similar (I assume) left_x_drive_configuration_byte_1 in STAR.setup. Could you see if that works?

BioCam: Ohh nice; I will try this out first thing tomorrow. Thank you for the tip.

BioCam: This works like a charm. 
I have the deleted the intermediate "check download date" solution and it is now replaced with the more robust `STAR.request_machine_configuration()` check for bit 3.

BioCam:
I have rewritten the original version to
1. Actually check whether an autoload is installed directly
2. Exposed a true `park_autoload` that identifies the number of tracks/positions (differs between STAR & STARlet) and when called sends the autoload to the last track where it is safely out of the way.

However, one of the unit tests suddenly fails and I am not yet sure why:

```python
=================================== FAILURES ===================================
___________________ LiquidHandlingApiGeneralTests.test_stop ____________________

self = <liquid_handling_api_tests.LiquidHandlingApiGeneralTests testMethod=test_stop>

    def test_stop(self):
      with self.app.test_client() as client:
        response = client.post(self.base_url + "/stop")
        self.assertEqual(response.status_code, 200)
>       self.assertIn(response.json.get("status"), {"running", "succeeded"})
E       AssertionError: 'error' not found in {'running', 'succeeded'}

pylabrobot/server/liquid_handling_api_tests.py:64: AssertionError
```

rickwierenga:
Awesome. I'll look at the code very soon.

Regarding the test: this is some kind of race condition that sometimes occurs. I have to debug that, but haven't found the time yet. It's fine to ignore for now.

PR #65, opened 2024-01-31
Expose simple move autoload to slot & simplify load_carrier

BioCam:
Mini-PR for autoload features recently merged:

1. Expose `move_autoload_to_slot()` - a simple, specific command to send the autoload to a specific location.
2. Simplify the `load_carrier()` function: currently, if `park_autoload_after = True` it will perform a full search of other carriers as a surrogate to move to the safe park position. I changed this to directly park the autoload, saving a lot of wait time.


rickwierenga:
Thanks!

PR #69, opened 2024-02-21
Probe z height using channel

BioCam:
Hi everyone,

I've written a handy function that allows one to probe the surface of any (conductive) object on a STAR(let) deck, and thereby map out the precise dimensions of the object in an automated way using PLR. This effectively makes the STAR act as a [Coordinate-Measurement Machine](https://en.wikipedia.org/wiki/Coordinate-measuring_machine) (CMM).

The first main aim of making this functionality available is to automate identification of precise z-values of different resources, and thereby help create more robust labware definitions for the PLR labware library.

A simple example of its use:
- create a list of x positions
- move each channel to a different y position
- loop over your x positions, and loop over your channels to call `lh.backend.probe_z_height_using_channel(chanel_index_as_int)` to collect z height readings

For a bare MFX_CAR_base this allows you to generate the following 3-dimensional map:
240219_MFX_CAR_base_STARlet_measurements_12min_32sec
...each axis shows the absolute x, y, z position in relationship to the STAR's deck origin (which is at the left-front-100mm below the physical deck).

For safety purposes, I recommend using Hamilton 50ul tips for mapping surfaces. These are relatively long and soft, acting as 'cushions' in case you try out faster detection speeds (not recommended). Small bends are tolerated well by the 50ul tips. Metal teaching needles might enable more sensitive measurements but I have not tested them due to concerns about damaging the channels. (Please let us know if you try them out.)

Sensor accuracy for z-height readings needs to be further tested but in my hands has been at least 0.2 mm, showcasing the STAR's incredible hardware.

This allows you to
- automatically identify the z-height of your labware, but extends further into
- mapping surface flatness for GD&T validation,
- will be useful for ad-hoc volume calculations of liquids in a well/container/tube on the deck (in synchrony with the optimised volume functions),
...and more (it might even be useful as a tool in more classic metrology settings)

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR.py
@@ -6775,10 +6775,10 @@ class STAR(HamiltonLiquidHandler):
     lowest_immers_pos_str = f"{lowest_immers_pos:05}"
     start_pos_lld_search_str = f"{start_pos_lld_search:05}"
     channel_speed_str = f"{channel_speed:05}"
-    channel_acc_str = f"{channel_acceleration:03}" 
+    channel_acc_str = f"{channel_acceleration:03}"
     detection_edge_str = f"{detection_edge:04}"
-    detection_drop_str = f"{detection_drop:04}" 
-    post_detection_dist_str = f"{post_detection_dist:04}" 
+    detection_drop_str = f"{detection_drop:04}"
+    post_detection_dist_str = f"{post_detection_dist:04}"
 
     await self.send_command(
       module=f"P{channel_idx}",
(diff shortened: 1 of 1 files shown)


rickwierenga:
pr 69 is unfathomably based

jkhales:
amazing

rickwierenga:
The channel_idx argument here is 1-indexed. I think since this function uses PLR units, and is not a python wrapper around a single firmware command, it would be more correct to use plr-style 0-indexing. Shall I make a PR?

BioCam:
> The channel_idx argument here is 1-indexed. I think since this function uses PLR units, and is not a python wrapper around a single firmware command, it would be more correct to use plr-style 0-indexing. Shall I make a PR?

No need, I added exactly that fix in my overhaul of the entire probing functions upgrade in PR260 :)

PR #71, opened 2024-03-01
MFX Carrier integration

BioCam:
Hi everyone,

The time has come to integrate MFX carriers into PLR and open ways to the combinatorial explosion of specialised PLR deck setups with them.

This simple PR adds the most universal MFX carrier, `MFX_CAR_L5_base`, to PLR.

This only includes the carrier itself.

Maintaining Hamilton's vocabulary, "MFX modules" can then be added in any combination that is needed by the user. 
However, we have not yet integrated any of these mfx_modules and are discussing which ones to integrate.
For a list of standard Hamilton MFX modules see their [catalogue number listing](https://www.hamiltoncompany.com/automated-liquid-handling/carriers) or look up the online catalogue that Hamilton shared in [this PyLabRobot forum post](https://forums.pylabrobot.org/t/library-downloads-links/1052).

In the meantime, PLR users can generate a `Resource()` instance of any specialised module that they want to use right now, and simply assign it to a `MFX_CAR_L5_base` site.

One thing to note: 
The `MFX_CAR_L5_base` is essentially just a carrier base with 10x 3mm(diameter) holes one each site.
4 of these holes represent a "carrier site".
However, modules placed on top of them can have varying dimensions.

To generate high utility this PR shares the `MFX_CAR_L5_base` with 5 carrier sites that are maximally sized in the x-y plane leaving 2mm in between each site.
As more people test it out, we may change the carrier sites to what the community deems most useful for **the** generic MFX carrier.

If your carrier sites should be placed differently you can still use the `.assign_child_resource()` method to generate specific "sites" bespoke to your needs.


rickwierenga:
nice!

PR #72, opened 2024-03-02
Small dispense_pip Docstring correction

BioCam:


rickwierenga:
is this true? the firmware docs say otherwise

BioCam:
To my knowledge yet; I can see no physical reason cLLD should not work and there are loads of videos surface dispensations into liquids which would not be possible if cLLD is not possible +dispensing_mode: (Type of dispensing mode 0 = Partial volume in jet mode 1 = Blow out in jet mode 2 = Partial volume at surface 3 = Blow out at surface 4 = Empty tip at fix position) indicates this capability exists.

I can check it empirically on Monday as well.

BioCam:
Here is [an example of surface_dispense](https://www.youtube.com/watch?v=Et2RAXpzFjM) (though I could only find it for the Zeus quickly) min2:10

rickwierenga:
Interesting. I'll just wait for you to empirically verify this :)

BioCam:
I can definitively say that Hamilton machines are capable of cLLD during dispensation.

This very feature might be a key factor to their robustness.

rickwierenga:
Thanks!

PR #73, opened 2024-03-07
[lids] fix move lid to plate

jkhales:
Previously we were passing the absolute location of the "to_location" when assigning the resource. But it expects the relative location.


rickwierenga:
Thanks!

PR #74, opened 2024-03-07
[STAR] make plate_width depend on grip direction

jkhales:


pylabrobot/liquid_handling/backends/hamilton/STAR_tests.py
```diff
@@ -606,14 +607,14 @@ async def test_iswap_plate_reader(self):
...
-      "C0PRid0004xs10430xd0yj3282yd0zj2063zd0th2840te2840go1300gr4ga0",
+      "C0PRid0004xs10430xd0yj3282yd0zj2063zd0th2840te2840go0890gr3ga0",
                 "xs#####xd#yj####yd#zj####zd#th####te####go####gr#ga#")
 
     await self.lh.move_plate(plate_reader.get_plate(), self.plt_car[0],
       pickup_distance_from_top=8.2, get_direction=GripDirection.LEFT,
       put_direction=GripDirection.FRONT)
     self._assert_command_sent_once(
-      "C0PPid0005xs10430xd0yj3282yd0zj2063zd0gr4th2840te2840gw4go1300gb1237gt20ga0gc1",
+      "C0PPid0005xs10430xd0yj3282yd0zj2063zd0gr3th2840te2840gw4go0890gb0827gt20ga0gc1",
```
jkhales, line 617: please be sure to sign off on this change. I'm unsure if this is OK.

rickwierenga: Would it be possible grab the plate form the front here so we don't have to change the test? Feel free to add a second test for left gripping if you like.

pylabrobot/liquid_handling/backends/hamilton/STAR_tests.py
```diff
@@ -601,19 +602,19 @@ async def test_iswap_plate_reader(self):
...
+      "C0PRid0004xs10430xd0yj3282yd0zj2063zd0th2840te2840go1300gr1ga0",
                 "xs#####xd#yj####yd#zj####zd#th####te####go####gr#ga#")
 
     await self.lh.move_plate(plate_reader.get_plate(), self.plt_car[0],
-      pickup_distance_from_top=8.2, get_direction=GripDirection.LEFT,
+      pickup_distance_from_top=8.2, get_direction=GripDirection.FRONT,
       put_direction=GripDirection.FRONT)
     self._assert_command_sent_once(
-      "C0PPid0005xs10430xd0yj3282yd0zj2063zd0gr4th2840te2840gw4go1300gb1237gt20ga0gc1",
+      "C0PPid0005xs10430xd0yj3282yd0zj2063zd0gr1th2840te2840gw4go1300gb1237gt20ga0gc1",
```
jkhales, line 617: @rickwierenga updated to the FRONT like you suggested, but I still had to make a small change to the `gr` value: 4 -> 1. I'm guessing this corresponds to the GripDirection enum so is unavoidable.

I'm more comfortable with this small change (previous one, the coords were pretty opaque).

pylabrobot/liquid_handling/backends/hamilton/STAR_tests.py
jkhales: oh good idea. Kind of worked, commented on the new commit explaining details

pylabrobot/liquid_handling/backends/hamilton/STAR_tests.py
rickwierenga: Oh f we were gripping from the left earlier (gr4, fig. 9 of c0 fw doc). These go/gb values correspond to a grip on the long side, and this firmware command is from when I fetched a plate from the plate reader (where the plate is rotated by 90 degrees, so the left on the machine is actually the front of the plate).

`get_size_x/y` take rotation into account so commands should be generated correctly.

What is missing here is that LH should update the resource's rotation on the `move_plate` command where currently only its location is updated.

jkhales: I've reverted the firmware commands and am trying something like:

image


image

But I can't seem to make it work regardless of the different permutation I'm putting int. 🤔

jkhales: update: I updated the firmware strings to the correct grip directions, everything else is the same

PR #76, opened 2024-03-10
[resource] don't allow assigned resource location to be undefined

jkhales:


rickwierenga:
Summary of changes: resource.location must be defined when it has a parent, but can be None when a resource is not assigned anywhere.

The `location` argument of HamiltonDeck and TecanDeck still accept `None`, to allow the user to specify `rails` instead. Perhaps in the future we can introduce `assign_at_rails` to make the types more explicit.

The `assign_child_resource` of `Plate` also accepts `None` because it can assign a default location for lids. In the future, we should refactor this by adding an `add_lid` method and having `assign_child_resource` require a specified location.

PR #77, opened 2024-03-12
MFX module integration with DWP & universal TIP module examples

BioCam:
Hi everyone,

Hamilton enables complete customisation of deck layouts using the MFX carriers.

These carriers are just an "empty" carrier plate with multiple holes in them onto which `MFX_modules` can be mounted by simply screwing them into the carrier.

Hamilton provides numerous  `MFX_modules` for different purposes. 

In this PR I exemplify what is possible by...
- updating the existing definition of the `MFX_CAR_L5_base`, and integrating 2 different types of `MFX_modules`:
    - the `MFX_TIP_module` to position a high- (aka 1000ul), standard- (aka 300ul), low volume (aka 10ul), 50 ul or 5ml tip rack (but not a 384 tip rack) on any position of the carrier (Hamilton cat. no.: 188160),
    - the `MFX_DWP_module` to position a Deep Well Plate / tube racks (MATRIX or MICRONICS) / NUNC reagent trough ( Hamilton cat. no.: 188042).

Examples of how to define them:

```python
    MFX_CAR_L5_base,
    HTF_L, ST_L, TIP_50ul_L, # TIPS 
    Cos_96_Rd, # PLATES
    MFX_DWP_module, MFX_TIP_module,
)

# Generate a MFX carrier instance
mfx_carrier_1 = MFX_CAR_L5_base(name='mfx_carrier_1')

# Generate MFX TIP module instance and assign it to a MFX carrier site
mfx_carrier_1[0] = mfx_tip_module_1 = MFX_TIP_module(name="mfx_tip_module_1")
# Generate a TIP rack instance and assign it to the MFX TIP module
mfx_tip_module_1[0] = tip_50ul_rack = TIP_50ul_L(name="tip_50ul_rack")

# Generate a MFX DWP module instance and assign it to THE SAME MFX carrier
mfx_carrier_1[1] = mfx_dwp_module_1 = MFX_DWP_module(name="mfx_dwp_module_1")
# Generate a plate instance and assign it to the MFX DWP module - this can be a DWP or any plate
mfx_dwp_module_1[0] = cos96_plate_1 = Cos_96_Rd(name='Cos96_plate_1')

# After building your custom-designed carrier assign it to your machine's deck and automate :)
await lh.deck.assign_child_resource(mfx_carrier_1, rails=42)
```

These two MFX_modules were verified on a STARlet.

---

After many discussion I had to come up with a way to integrate MFX_modules into PLR that is relatively simple but still allows quick set up of new and custom-built deck layouts.
I found that simply using the `.assign_child_resource()` method is not sufficient to achieve this goal as it requires explicit declaration of the location of a child_resource in relationship to its parent_resource which is a major handicap of reusing MFX_modules and generating new MFX_carrier layouts quickly.

Instead I created a new `Resource` class called `MFXModule` which is a mirror of the existing `Carrier` classes:
Similar to how a `PlateCarrier` has a `CarrierSite` predefined at a Coordinate relative to the `PlateCarrier` origin,
a `MFXModule` has one (or potentially) more `MFXSite`s at a Coordinate relative to the `MFXModule` origin.
This allows one person to define a new `MFXModule`'s dimensions and everyone using PLR can then use it, straight away, without any further definitions - just like specific `PlateCarriers` and `TipCarriers`.

---

For a more comprehensive list of Hamilton MFX modules that we will over time aim to integrate see [this VENUS four Operator Manual](https://download.hamiltonsupport.com/wl/?id=VQY0XAzTfcT7KY29A1YJtZVFwZ7GMOO2&path=Hamilton%20Guides%2FVENUS%20four%20Operators%20Manual.pdf&mode=grid) which includes a digital catalogue in the Appendix, specifically PDF page 219 ff, and was kindly shared by a Hamilton employee through the PyLabRobot forum.

Some examples:
- MFX_TIP_module (188160)
- MFX_NTR_96_module (191420)
- MFX_NTR_384_module (196371)
- MFX_NTR4_module (191425)
- MFX_MTP_module (188041)
- MFX_DWP_module (188042)
- MFX_reagent_through_module (188047)
- MFX_tube_cup_module (188048)
...and so many more

---

**The MFX_CAR_L5_base and MFX_modules in the broader scheme of the PLR Resource Infrastructure**

To visualise how this system fits into the existing PLR Resource infrastructure I created the following infographic:

PLR_Labware_Resource_Library_v2

Deck layouts are created from the top down:
- LVL0 & LVL1 are to generate specific carriers the way PLR was originally designed - additional carriers (ShakerCarrier & SampleCarrier) are still in development but were left for "the bigger picture" and to encourage discussion on this infrastructure.
- At the end of LVL1, carrier instances can be generated, each with a specific number of carrier sites associated to them.
- Depending on the type of carrier, the next step deviates:
    - TipCarriers & PlateCarriers can directly receive assignments of tip_racks or plates to their carrier_sites (LVL3).
    - MFXCarriers require MFX_modules to be screwed into them to make their carrier_sites functional. The specific MFX_module chosen defines that carrier_site's function, e.g. to house tips, plates, machines (shakers, temperature controllers, tilt_modules, ...), reservoirs, throughs, etc. (LVL2).
    -> to use these MFX_modules further labware has to be placed on them. E.g. a tip_rack on a MFX_TIP_module or a plate on a MFX_plate_module (LVL3).

Please add any information you notice missing, making this system as useful and easy as possible will require multiple iterations of feedback and redesign.


BioCam:
@rickwierenga, do you know why the type hinting fails in:

```python
class MFXSite(Resource):
  """ A single site within a MFX module. """

  def __init__(self, name: str, size_x: float, size_y: float, size_z: float, spot: int,
    category: str = "mfx_site", model: Optional[str] = None):
    super().__init__(name=name, size_x=size_x, size_y=size_y, size_z=size_z, category=category,
      model=model)
    self.resource: Optional[Resource] = None
    self.spot: int = spot

  def assign_child_resource(
    self,
    resource: Resource,
    location: Optional[Coordinate],
    reassign: bool = True
  ):
    self.resource = resource
    return super().assign_child_resource(resource, location=location)
```
specifically `return super().assign_child_resource(resource, location=location)`?

I have taken this directly from the CarrierSite definition without modification.

rickwierenga:
`location` must now be not `None` on assignment: https://github.com/PyLabRobot/pylabrobot/pull/76

pylabrobot/resources/ml_star/mfx_modules.py
```diff
@@ -0,0 +1,228 @@
...
+    self.spot: int = spot
+
+  def assign_child_resource(
+    self,
+    resource: Resource,
+    location: Optional[Coordinate],
+    reassign: bool = True
+  ):
+    self.resource = resource
+    return super().assign_child_resource(resource, location=location)
```
rickwierenga, line 31: I think we should raise an error in this case.

pylabrobot/resources/ml_star/mfx_modules.py
```diff
@@ -0,0 +1,228 @@
...
+  site_size_y: float) -> List[MFXSite]:
+  """ Create a list of MFX module sites with the same size. """
+
+  n = len(locations)
+  return create_mfx_module_sites(locations, [site_size_x] * n, [site_size_y] * n)
+
+
+class MFXModule(Resource):
+  """ Abstract base resource for MFX modules to be placed
+    on a MFX carrier (landscape/portrait, 4/5 positions).
```
rickwierenga, line 74: This is essentially duplicating `Carrier`. Can we subclass `Carrier` if it's important to have a separate `MFXCarrier`? It is possible the `Carrier` class provides everything you need, in which case I'd discourage adding a new class.

rickwierenga:
> I found that simply using the .assign_child_resource() method is not sufficient to achieve this goal as it requires explicit declaration of the location of a child_resource in relationship to its parent_resource which is a major handicap of reusing MFX_modules and generating new MFX_carrier layouts quickly.

You could make it so `assign_child_resource` accepts a `slot` location, much like HamiltonDeck and OpentronsDeck do. This slot is used to fetch the location. When given a Coordinate, this overriden method will check that the location is actually a slot.

pylabrobot/resources/ml_star/mfx_modules.py
```diff
@@ -0,0 +1,228 @@
...
+from typing import List, Optional, Union
+
+from pylabrobot.resources.resource import Resource
+from pylabrobot.resources.carrier import Coordinate
+
+logger = logging.getLogger("pylabrobot")
+
+
+class MFXSite(Resource):
+  """ A single site within a MFX module. """
```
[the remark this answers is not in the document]
BioCam: Yes, it is exactly the same and can therefore be replaced by `CarrierSite`.

I simply chose to name it `MFXSite` to avoid confusion about why a `CarrierSite` houses a `MFXModule` which houses a `CarrierSite`?

pylabrobot/resources/ml_star/mfx_modules.py
BioCam: You are right, it is essentially duplicating `Carrier`.

But while the software definition is essentially the same, the use is very different and emulates the way that one physically builds MFX_carriers: 
take empty carrier -> choose slot -> screw in MFX_module -> add labware to MFX_module

Continuing with `Carrier` this definition becomes... hard to understand:
take empty `Carrier` -> choose slot -> screw in "small" `Carrier` -> add labware to "small" Carrier's slot?

Hamilton also provides an avalanche of different MFX_modules. Exposing these to PLR will enable complete customisation of Hamilton deck layouts in PLR with no extra time commitment (as opposed to every person having to figure out how to define their MFX_modules in PLR from scratch).

rickwierenga:
I made these `MFXModule`s just mostly plain resources, just with the restriction of only taking one child resource. I think this models reality most closely in PLR.

Interestingly, these are actually saved as carriers (`tml` files) by Hamilton in `C:\Program Files (x86)\HAMILTON\LabWare\ML_STAR\MFXCreation` and indeed have one site. This is closer to your implementation before, but I think this new implementation is a lot simpler.

BioCam:
Hi @rickwierenga.
Thank you for the modification, that is indeed much cleaner.

In summary, you got around the issue of assigning a plate or a tip_rack to a specific position on the corresponding `MFXModule` by creating new class methods for each `mfx_module` type, i.e. 
- `MFX_TIP_module` -> `assign_child_resource()`-wrapper at relative location for a tip_rack is `assign_tip_rack()`
- `MFX_DWP_module` -> `assign_child_resource()`-wrapper at relative location for a plate is `assign_plate()`

In each case this means the assigned tip_rack/plate does not require specification of where it is placed exactly but instead is predefined - which is precisely what we want.

There are two points I would like to get some feedback on with this new implementation:
1. This now means that we have to write the same lengthy class definition for each new `mfx_module`. Since we only have to do this once I don't see a problem with this. But I want to mention that with each new `mfx_module` being added there will be quite a bit of repetitive class definition code and a very lengthy `mfx_modules.py`:
image
2. This implementation removes the ability to use indexing to assign a tip_rack or plate to its mfx_module, i.e. `mfx_carrier_1[1] = mfx_tip_module_1 = MFX_TIP_module(name="MFX_tip_module")` is no longer possible.
Which means to build a deck users now have to navigate child resources assignments using:
    - 'rails' (parent=deck only)
    - 'slots'
    - 'location'
    - index-based
    - 'assign_tip_rack' (parent=MFXModule for tips only)
    - 'assign_plate' (parent=MFXModule for tips only)
...which sounds quite confusing to me. Which one of these different ways can be used to assign a child resource is dependent on the how the parent is configured, with only `location` being universally accepted?
I found index-based assignment to be the most useful as different child resources can be assigned using the same loop without having to create specialised conditional statements. 
Is there a way we can still allow the mfx_modules to use an index-based assignment?

BioCam:
To point 1, yes, two more specific MFXModule classes rather than one might be more specific.
But then we'd need to cover all different MFX module cases separately: throughs, tube-holder-racks that are screwed into the MFX carrier, micro-titer plate (MTP) sides, deep well plate (DWP) sites, tiprack sites for nested tip racks and normal tipracks, ....

That is why I tried to just create one class. Maybe your initial way of going about it and just having some lengthy class definitions that take the `MFXModule` base class as their start is the way to go here.

To point 2... How is `lh.move_plate(plate_instance, mfx_DWP_module)` supposed to take the mfx_DWP_module as input?

Normally we would tell `move_plate` to take the carrier_site but we just removed the equivalent site from the `MFXModule` and it is throwing up an error when handing it the MFX_module because its origin is at the top of the MFX_carrier.

BioCam:
Side note:
I took more detailed measurements MFXCarrier sites should be flush with the sides of the carrier itself, i.e. their x-origin is 0 in relation to their parent (i.e. the carrier).

The first y-origin is 5mm into the deck, each site is 94mm in y, and the next site starts at 96mm from the previous one.
This neatly results in the mfx_tip_module's and mfx_plate_module's "sites" to be exactly the same as those of the tip_carrier and plate_carrier respectively.

rickwierenga:
> To point 1, yes, two more specific MFXModule classes rather than one might be more specific.
But then we'd need to cover all different MFX module cases separately: throughs, tube-holder-racks that are screwed into the MFX carrier, micro-titer plate (MTP) sites, deep well plate (DWP) sites, tiprack sites for nested tip racks and normal tipracks, ....
> That is why I tried to just create one class. Maybe your initial way of going about it and just having some lengthy class definitions that take the MFXModule base class as their start is the way to go here.

Yes, perhaps a singular `MFXModule` base would be good. Right now, I see no need for it but it could be good in the future. What would this class manage?

It makes sense to me to have different classes for different modules that have different behavior. This way, the class can manage all the complexity for a module-kind (like plate carrier). This could lead to a few classes rather than 1, but if there is no good abstraction this has to be the case. If there is a common functionality that is shared across all MFXModules but not Resource itself, then we can make that into a common base class.

> To point 2... How is lh.move_plate(plate_instance, mfx_DWP_module) supposed to take the mfx_DWP_module as input? 
> Normally we would tell move_plate to take the carrier_site but we just removed the equivalent site from the MFXModule and it is throwing up an error when handing it the MFX_module because its origin is at the top of the MFX_carrier, i.e. 96mm further down (in the z-axis).

`LiquidHandler.move_plate` has checks for the destination type and adjusts the location of for a plate movement. I think there would be a good place to fix this.

You can just tell it to move to a `MFX_DWP_module` (`MFXPLateModule`?) directly: `lh.move_plate(plate, plate_module)`. You don't need a CarrierSite(-equivalent) for that.

> The first y-origin is 5mm into the deck, each site is 94mm in y, and the next site starts at 96mm from the previous one.
This neatly results in the mfx_tip_module's and mfx_plate_module's "sites" to be exactly the same as those of the tip_carrier and plate_carrier respectively.

Great! Did you also look at the Hamilton definition files? They usually contain most of the information you cared about in very precise numbers.

BioCam:
> Yes, perhaps a singular MFXModule base would be good. Right now, I see no need for it but it could be good in the future. What would this class manage?
It makes sense to me to have different classes for different modules that have different behavior. This way, the class can manage all the complexity for a module-kind (like plate carrier). This could lead to a few classes rather than 1, but if there is no good abstraction this has to be the case. If there is a common functionality that is shared across all MFXModules but not Resource itself, then we can make that into a common base class.

The common functionality across all `MFXModule`s, which does not exist in the `Resource` class, is that a `MFXModule` exists solely for the purpose of providing a site/slot that is to be used (to assign labware to, to move labware onto or off from).

Looking at the `Resource` class definition (`pylabrobot/resources/resource.py`) I cannot see a definition of any subpart of it for a particular purpose. I assume this is why the `CarrierSite` has been created in `carrier.py` - to provide a set of indepently accessible and useable sites on the `Carrier` instance, but importantly, these `CarrierSite`s are still part of the `Carrier`.

`MFXModule` is supposed to be a mixture between `Carrier` and `CarrierSite` -> A class that does not only provide its basic definition in x-y-z but also a definition of the site on it that is the actually useful part.
In this regard...

> LiquidHandler.move_plate has checks for the destination type and adjusts the location of for a plate movement. I think there would be a good place to fix this.
You can just tell it to move to a MFX_DWP_module (MFXPLateModule?) directly: lh.move_plate(plate, plate_module). You don't need a CarrierSite(-equivalent) for that.

...that is what I have done -> the error I was speaking of is not a software error but it is a hardware error caused by sending the plate to the mfx_plate_module using `lh.move_plate(plate, plate_module)`!

Here is an example to demonstrate what is going on and why `MFXModule`'s reason for existence is to provide a site on it that is independently accessible via code:

For the purpose of this PR, I quickly physically assembled the following MFXCarrier instance:

image

Using `lh.move_plate(plate, plate_module/MFX_DWP_module)` will try to move the plate to the center of the `Resource` it is targeting (whether this is the 3D- or the 2D- center, I don't know).
This means this command crashes the plate onto the top of the plate_module, trying to reach either the plate_module's z-center or z-bottom (depending on whether the 3D- or 2D-center is used).

This is why all `MFXModule`'s need a site/slot.
The way I implemented this was with a separate `MFXSite` class but I agree that this is too wordy and in the end just a copy of `Carrier`+`CarrierSite`. So instead `MFXModule` should declare its site directly in its definition - that is what would make it unique and useful.
By declaring `.assign_plate()` we never create this site. Instead we just assign a plate in a pre-specified location - but, importantly, this location is not independently accessible to the `lh.move_plate()` method (to my knowledge).

BioCam:
> > The first y-origin is 5mm into the deck, each site is 94mm in y, and the next site starts at 96mm from the previous one.
This neatly results in the mfx_tip_module's and mfx_plate_module's "sites" to be exactly the same as those of the tip_carrier and plate_carrier respectively.

> Great!` Did you also look at the Hamilton definition files? They usually contain most of the information you cared about in very precise numbers.

Unfortunately, that is not the case with MFX_carriers:

MFX_carriers are `.tml` files once they are created. They only define the actual sites that are used in each position. They do not define the MFX_module geometries that house these sites.

My guess is that VENUS doesn't need the precise definitions of the MFX_modules because it only ever uses the MFX sites themselves, ignoring the underlying geometry of the MFX_modules (and makes it appear as if it does in the 3D render by using a [.x file](https://learn.microsoft.com/en-us/windows/win32/direct3d9/x-files--legacy-) for the visualisation).

This makes sense in the offline system that VENUS operates in: 
MFX_carriers are created on a need-to-need basis for VENUS locally on a users computer. This means there is no need to create a common database of all possible MFX_carrier instances.

This is not the case for PLR.
In PLR we have a Git repo that should enable fast access to any MFX_carrier module combination instance that one wants to generate.
But with 5 carrier_sites `MFX_CAR_L5_base` and (let's assume a lower bound of) 10 possible MFX_modules, PLR would need 10**5 different MFX_carrier specifications if it were to following the same implementation that VENUS does (incidentally highlighting another issue with VENUS).

So instead I propose we have a small set of empty MFX_carrier classes, the most common one will be `MFX_CAR_L5_base` but there might be some with 4 carrier_sites, 7T widths or portrait-mode carrier_sites.
Then we provide all Hamilton MFX_modules as `MFXModule` definitions, with each `MFXModule` defining its site directly, i.e. without a second `CarrierSite`/`MFXSite` class which makes definitions and use easier.

Then anyone can use the `MFX_CAR_L5_base` and simply assign the `MFXModule` they like to each carrier_site.

Each `MFXModule` then represents the true, physical dimensions of that module with the real-world attachments to its carrier.
VENUS does not do that! And it doesn't have to because it abstracts the MFXModules away by providing new carrier instances every single time.

As a result, PLR depends on us to generate true, meaningful definitions of what `MFXModule`s look like in real-life.
And we can then cross-verify our more detailed definitions using the VENUS' carrier and carrier site definition.

BioCam:
Small update:
In line with the increased documentation of the PLR resource library I have added a README.md file to the `/pylabrobot/resources/ml_star` folder which explains in a fast, visual manner what labware has already been defined and used by other PLR users:

image

Due to the vast number of labware definitions (from various carrier types, to tip_racks and MFX_modules), this is only the start of this README and PLR users should be encouraged to add more information over time.

rickwierenga:
> The common functionality across all MFXModules, which does not exist in the Resource class, is that a MFXModule exists solely for the purpose of providing a site/slot that is to be used (to assign labware to AND to move labware onto or off from).

Sure. I am not convinced this behavior is a "functionality" that requires specific modeling in PLR. It is just the name and typical way of using these resources. A subclass would not really add anything.

> I assume this is why the CarrierSite has been created in carrier.py - to provide a set of indepently accessible and useable sites on the Carrier instance, but importantly, these CarrierSites are still part of the Carrier.

The `CarrierSite`s are not needed to model the relation between a Carrier and Plates/TipRacks. That could just be a simple child/parent relationship as elsewhere.

The reason `CarrierSite` exists is we need to refer to empty spots in `lh.move_resource(plate, <some spot>)`. This is impossible to do with the virtual resource of a `CarrierSite`. You could pass a carrier to `<some spot>` but then the spot itself is ambiguous, or you could do a `Coordinate` but then it isn't clear that the plate to be moved is supposed to become a child of the CarrierSite. This is the key issue. The CarrierSite serves as a way to resolve this issue. In LiquidHandler, specific code exists to deal with these sites.

For `MFXPlateModule` (any MFXModule) that takes a plate, this issue does not exist. When you specify the `MFXPlateModule` as a destination, there is only one spot. This means that it is unambiguous when you move a plate there where it is going to go. Compare this to the ambiguity that would exist when saying move_plate to a "carrier": it could be any of the 5 carrier spots.

Now I get your feeling that there is some duplication of code here, namely the code that assigns a new child resource should check whether a plate already exists as a child. It does ring the possible-abstraction-bell. I have not figured out what this abstraction should be to do this cleanly. A `PlateTakingResource`? For now, let's accept this slight duplication of code ("a little duplication is cheaper than the wrong abstraction"). This can be a follow-up PR.

> ...that is what I have done -> the error I was speaking of is not a software error but it is a hardware error caused by sending the plate to the mfx_plate_module using lh.move_plate(plate, plate_module)!
> This is why all MFXModule's need a site/slot.

Ah, here we disagree.

I would suggest fixing this by adjusting `LiquidHandler.move_plate`. If you look at the existing code, that should clarify things.

```python
    if isinstance(to, ResourceStack):
      assert to.direction == "z", "Only ResourceStacks with direction 'z' are currently supported"
      to_location = to.get_absolute_location()
      to_location = Coordinate(
        x=to_location.x,
        y=to_location.y,
        z=to_location.z  + to.get_size_z())
    elif isinstance(to, Coordinate):
      to_location = to
    else:
      to_location = to.get_absolute_location()
```

The conflicting axiom in this discussion appears to be that in my mind, the resource model should model the resource layout as simply and as close to reality as possible with the specialized PLR-modules like `liquid_handling` working with the provided structure. The resource module should not accommodate to higher-level modules unless absolutely necessary. I think with the CarrierSite, it appeared necessary and `move_plate` revealed that the simplest approach (a direct carrier -> plate relation) was not 'as simple as possible' because the 'possible' qualifier failed.

> MFX_carriers are .tml files once they are created. They only define the actual sites that are used in each position. They do not define the MFX_module geometries that house these sites.

In `C:\Program Files (x86)\HAMILTON\LabWare\ML_STAR\MultiFlexCarrier` I see many files that appear to be for the individual modules / base carriers. From first principles, it must include definitions for each of these modules because as you say it is able to generate these specialized carrier files in an offline environment.

As you say later "PLR depends on us to generate true, meaningful definitions of what MFXModules look like in real-life." My bet is this information is included in VENUS once you install the MFX tool, it is just not used directly when you create a deck layout, just for creating the custom carriers with the custom tool.

One other thing to consider when thinking about this: in theory it is possible to place the MFXModules directly on the deck or on a custom-designed base, it does not have to be used with a hamilton base carrier. For this reason, we definitely need the modules to have independent definitions.

> This is not the case for PLR.
> In PLR we have a Git repo that should enable fast access to any MFX_carrier module combination instance that one wants to generate.

Yes, with PLR they should be constructed in memory dynamically as the program runs. This is different from saving definitions to a file / defining them as creator-functions (like `Cos_96_EZWash` in PLR).

> But with 5 carrier_sites MFX_CAR_L5_base and (let's assume a lower bound of) 10 possible MFX_modules, PLR would need 10**5 different MFX_carrier specifications if it were to follow the same implementation that VENUS does.

Not if you generate them dynamically. It should be very similar to how you create a plate carrier (mfx base module) and then configure this carrier with different plates (mfx modules), with the extra step of assigning plates to the mfx modules.

> Each MFXModule then represents the true, physical dimensions of that module with the real-world attachments to its carrier.
VENUS does not do that! And it doesn't have to because it abstracts the MFXModules away by providing new carrier instances every single time. You can imagine, looking at the image of the MFX_carrier above, that VENUS actually only sees the cyan-coloured carrier and the purple-coloured MFX_sites - it doesn't need more.

Exactly

> 2) it would require the generation of a new class either way because afaik there is not way to assign a CarrierSite (top of the MFXModule) to another CarrierSite (site on the MFX_carrier?

Not important to the discussion, but asking because I hope your mental model should be correct: why do you think this?

rickwierenga:
Perhaps the location of where the resource goes (a "site") can be an attribute of `MFXModule` so that in lh.move_resource we can have a single `isinstance(to, MFXModule)`.

```diff
+ elif isinstance(to, MFXModule):
+   to_location = to.resource_site # Coordinate
```

BioCam:
There is a lot I don't fully understand and will message you about later.

For this PR, I think we should figure out how to solve the current issues with the implementation:
1. Plates cannot be moved onto a MFXModule.
2. I just realised that the latest change breaks the visualizer. The plates and tip_racks assigned to their MFX_modules suddenly appear in the bottom left corner on top of one another (even though their `.get_absolute_location()` shows their correct location and in real-life they do work:
image


Neither of these issues were present in the initial version of this PR using `MFXModule` and `MFXSite`. But I agree there should be a better way.

I am not sure why this happens but regarding point 1:

> I would suggest fixing this by adjusting LiquidHandler.move_plate. If you look at the existing code, that should clarify things.
```python
    if isinstance(to, ResourceStack):
      assert to.direction == "z", "Only ResourceStacks with direction 'z' are currently supported"
      to_location = to.get_absolute_location()
      to_location = Coordinate(
        x=to_location.x,
        y=to_location.y,
        z=to_location.z  + to.get_size_z())
    elif isinstance(to, Coordinate):
      to_location = to
    else:
      to_location = to.get_absolute_location()
```
To be honest, I was not aware of the existence of the `ResourceStack` class until now but it looks very useful to place Resources directly on top of one another, i.e. the same x-y origin and upwards in z by the height of the ResourceStack.

But if I understand this correctly, in our case the MFX_module to move onto is not of type `ResourceStack` but of type `Resource` which would trigger the `else` statement?
This would then explain why I see the `move_plate(plate, mfx_plate_module)` crash the plate into the top of the mfx_plate_module triggering a hardware error. 

As I just messaged you about in our DM: I have a number of use cases where I have to assign a child to a parent - both of class `Resource` - where each `Resource` can only accept a child on top its parent (i.e.  ` z=to_location.z  + to.get_size_z())` as in the `ResourceStack` example above) AND with an offset in the x- and y-dimension:

Do you think we could solve all of these issues by enabling the definition of one or multiple "sites" inside the `Resource` class?
As long as these sites are directly callable they would allow stacking of `Resource`s on top of each other in precise offsets without having to declare them every time anew and enable moving of `Resources`/plates directly to them.


Just about the VENUS-based mfx_module definitions:

> In C:\Program Files (x86)\HAMILTON\LabWare\ML_STAR\MultiFlexCarrier I see many files that appear to be for the individual modules / base carriers. From first principles, it must include definitions for each of these modules because as you say it is able to generate these specialized carrier files in an offline environment.

I looked in detail at this folder and here is its content breakdown:
image

The majority of the files are visualisation files for MFX_modules. The actual MFX_modules are not in the folder.
It appears that instead the MFX_module_sites are added to a MFX_CAR_ ... base in the "LabwareAssistant3D.exe" app and the app then stores the final mfx_carrier as a .tml file inside this folder.
The app appears to know the offsets for each MFX_module_site.
But VENUS does not have any need to precisely define the underlying MFX_module itself because it never uses them.

rickwierenga:
> Plates cannot be moved onto a MFXModule.

I will fix this soon with a temporary implementation of what I proposed above. Feel free to refactor it later.

> I just realised that the latest change breaks the visualizer. The plates and tip_racks assigned to their MFX_modules suddenly appear in the bottom left corner on top of one another (even though their .get_absolute_location() shows their correct location and in real-life they do work:

Thank you for bringing this up. Will fix.

> But if I understand this correctly, in our case the MFX_module to move onto is not of type ResourceStack but of type Resource which would trigger the else statement?

It would be of type MFXModule or MFXPlateModule, depending on the way we go. That's the diff I proposed above.

But currently, yes, this current implementation is what caused the hardware error you observed.

> To be honest, I was not aware of the existence of the ResourceStack class until now but it looks very useful to place Resources directly on top of one another, i.e. the same x-y origin and upwards in z by the height of the ResourceStack.

I added this class to support the stacking of plates. See https://github.com/PyLabRobot/pylabrobot/blob/main/pylabrobot/liquid_handling/backends/hamilton/STAR_tests.py#L644 for an example. I actually plan on removing this class in the near future in support of an `anchor` parameter to `move_plate`, which would take topcenter/left front bottom/etc.

> Do you think we could solve all of these issues by enabling the definition of one or multiple "sites" inside the Resource class?

This should be specific to the MFXModule/MFXPlateModule imo. Not all resources (eg `Well`s) need this. Will write something for you to look at.

>The majority of the files are visualisation files for MFX_modules. The actual MFX_modules are not in the folder.
It appears that instead the MFX_module_sites are added to a MFX_CAR_ ... base in the "LabwareAssistant3D.exe" app and the app then stores the final mfx_carrier as a .tml file inside this folder.
The app appears to know the offsets for each MFX_module_site.

Counterintuitive. I will look into that as well.

rickwierenga:
Visualizer issue fixed. It seems a little bit off though?

I think the resource movement should also work, but can't confirm.

image

BioCam:
Thank you @rickwierenga for your modifications.

In summary, you have modified the `MFXModule` class to take a new "site", which is now called `child_resource_location()` on it as a fixed destination to be used when `lh.move_plate(some_plate, MFX_module_instance)`.
That means `lh.move_plate` instantly takes the child_resource_location as its target destination if it is given an instance of `MFXModule`.
I think this is a very neat and useful fix!

My testing:
1. This does fix the issue of not being able to move a plate onto a mfx_module, i.e. I can move plates onto MFX modules now without any (hardware) errors.
2. But, this eliminates the ability to then move that plate again off that same mfx_module. It appears that the `absolute_location` of the plate is not correctly updated after moving it to the `MFX_module.child_resource_location`:

```python
# Generate MFX & PLATE carrier instances
mfx_carrier_1 = MFX_CAR_L5_base(name='mfx_carrier_1')

lh.deck.assign_child_resource(mfx_carrier_1, rails=1)
lh.deck.assign_child_resource(plt_carrier_1, rails=25)

mfx_carrier_1[0] = mfx_dwp_module_1 = MFX_DWP_module(name="MFX_DWP_module")
mfx_dwp_module_1.get_absolute_location()
# ===> Coordinate(x=100.0, y=68.0, z=118.195)


### TEST 1: Assigning plate straight onto MFX_DW_module ###
prep_plate_TEST1 = Azenta4titudeFrameStar_96_wellplate_skirted(name='prep_plate_TEST1')
mfx_dwp_module_1.assign_child_resource(prep_plate_TEST)

prep_plate_TEST1.get_absolute_location()
# ===> Coordinate(x=104.0, y=71.5, z=178.73) <=== CORRECT


### TEST 2: Assigning plate to plate_carrier and then move onto MFX_DW_module ###
plt_carrier_1[0] = prep_plate_TEST2 = Azenta4titudeFrameStar_96_wellplate_skirted(name='prep_plate_TEST2')
prep_plate_TEST2.get_absolute_location()
# ===> Coordinate(x=644.0, y=71.5, z=186.15)

await lh.move_plate(
    prep_plate_0,
    mfx_dwp_module_1, 
    use_arm="core", 
    channel_1=1, channel_2=2
)
prep_plate_TEST2.get_absolute_location()
# ===> Coordinate(x=204.0, y=139.5, z=296.925)  <=== INCORRECT
```

Troubleshooting effort:
```python
Coordinate(x=204.0, y=139.5, z=296.925) - Coordinate(x=104.0, y=71.5, z=178.73) # wrong plate location after new move command - plate location it should be at
= Coordinate(x=100.0, y=68.0, z=118.195)
```
===> that is the origin of the `mfx_carrier_1[0].get_absolute_location()`

3. Even after updating my system and syncing with the latest changes the visualiser still shows the same erroneous bottom-left location of the mfx_module assignments. 
I am still not familiar with the way the visualiser works and couldn't troubleshoot potential reasons.

rickwierenga:
2 should be fixed

BioCam:
This is perfect.
Thank you, @rickwierenga !

So, `MFXModule`s are Resources to which a child resource can be assigned at a specific location **without having to specify that location directly** because it has been specified when the Resource function has been written.

This is incredibly useful for re-using other people's designs.

At the moment the PLR labware library has two `MFXModules` integrated:
- `MFX_TIP_module`
- `MFX_DWP_module` (specifically the "rack-based" one; we will have to rename this in the future because there are multiple mfx_dwp_modules available, some rack- and some container-based (i.e. they have a skirt).

But otherwise this PR aimed to integrate the concept of `MFXModule`s and give specific examples for immediate use and v1 of these has now been added.

rickwierenga:
I will fix the visualizer. After that, what else do I need to do / would you like to do before merging?

BioCam:
I think the only thing left to do is to rename `MFX_DWP_module` to either `MFX_DWP_rackbased_module` or something similar that is a bit shorter.
But since we haven't integrated the distinction between rack-based (without a skirt) and container-based (with a skirt) resources yet we might want to discuss this more on the forum and merge the PR as is?

rickwierenga:
Simulator is working well now. Perhaps it was a cache issue? I did not have to make any changes.

image

```python
from pylabrobot.resources import MFX_CAR_L5_base
from pylabrobot.resources.ml_star import MFX_TIP_module
from pylabrobot.resources.ml_star import MFX_DWP_rackbased_module

# define mfx_car
mfx_car = MFX_CAR_L5_base(name='mfx_carrier')
mfx_car[0] = mfx_tip_car = MFX_TIP_module(name='mfx_tip')
mfx_car[1] = mfx_plate = MFX_DWP_rackbased_module(name='mfx_plate')
lh.deck.assign_child_resource(mfx_tip_car, rails=1)

# assign children
new_tip_rack = HTF_L(name='new_tips', with_tips=False)
mfx_tip_car.assign_child_resource(new_tip_rack)

new_plate = Cos_96_DW_1mL(name='new_plate')
mfx_plate.assign_child_resource(new_plate)
```

pylabrobot/resources/ml_star/mfx_modules.py
```diff
@@ -100,8 +100,8 @@ def MFX_TIP_module(name: str) -> MFXModule:
   )
 
 
-def MFX_DWP_module(name: str) -> MFXModule:
-  """ Hamilton cat. no.: 188042
+def MFX_DWP_rackbased_module(name: str) -> MFXModule:
+  """ Hamilton cat. no.: 188229
```
rickwierenga: I can't find the 188229 identifier in the document.

Also, I'm not the biggest fan of adopting the rack-vs-container terminology.

Isn't "188042 MULTIFLEX DWP MODULE" a well-defined resource? (page 126 of MICROLAB STAR Line Operator’s Manual)

BioCam: > I can't find the 188229 identifier in the document.

Yes, that is because the document is heavily incomplete.
Interestingly one cannot find the catalogue number either on [Hamilton's website](https://www.hamiltoncompany.com/automated-liquid-handling/carriers) either - it's almost as if Hamilton really doesn't want to sell their products.

But I've got this catalogue number directly from the Hamilton invoice used to purchase this MFX module and cross-verified its correctness using the Labware Assistant 3D:
image
This is therefore a well-defined Hamilton resource that I have also seen in Synthace videos and other Hamilton setups, including at Hamilton trainings.

> Isn't "188042 MULTIFLEX DWP MODULE" a well-defined resource? (page 126 of MICROLAB STAR Line Operator’s Manual)

It appears that this is actually a "skirted"/"container-based" DWP, i.e. the opposite of what I added to the PLR resource library here:
image

> Also, I'm not the biggest fan of adopting the rack-vs-container terminology.

Yes, I agree. A clear definition of the concept is deeply needed for PLR because currently the lack thereof causes crashes (and has to be solved in a hard-coded way using offsets in my experience).
But the word "rack-based" is illogical to mean "sitting on its frame".
If we don't adapt the Hamilton-tested terminology we have to invent our own though, especially to avoid the confusion about naming which causes people to buy the wrong modules and costs up to ~$500 per wrong purchase (in the new-purchase market).
I'm happy to brainstorm ideas of how to incorporate the concept into PLR with different naming and rewrite all labware the takes plates on a specific site.

BioCam:
> Simulator is working well now. Perhaps it was a cache issue? I did not have to make any changes.
> 
> image
> ```python
> from pylabrobot.resources import MFX_CAR_L5_base
> from pylabrobot.resources.ml_star import MFX_TIP_module
> from pylabrobot.resources.ml_star import MFX_DWP_rackbased_module
> 
> # define mfx_car
> mfx_car = MFX_CAR_L5_base(name='mfx_carrier')
> mfx_car[0] = mfx_tip_car = MFX_TIP_module(name='mfx_tip')
> mfx_car[1] = mfx_plate = MFX_DWP_rackbased_module(name='mfx_plate')
> lh.deck.assign_child_resource(mfx_tip_car, rails=1)
> 
> # assign children
> new_tip_rack = HTF_L(name='new_tips', with_tips=False)
> mfx_tip_car.assign_child_resource(new_tip_rack)
> 
> new_plate = Cos_96_DW_1mL(name='new_plate')
> mfx_plate.assign_child_resource(new_plate)
> ```

Yes, I think you are right - I restarted everything and it works very well now. Sorry for the confusion :)

pylabrobot/resources/ml_star/mfx_modules.py
BioCam: Yes... I do believe the PLR resource library could end up becoming the digital Hamilton labware catalogue that everyone I know wishes exists 😅

rickwierenga: OK, we should adopt the name from the mfx tool "dwp rack based". That makes sense. Thank you for the detailed reference.

rickwierenga:
> Yes, I think you are right - I restarted everything and it works very well now. Sorry for the confusion :)

Also an issue on my end, I previously saw it not working... https://github.com/PyLabRobot/pylabrobot/pull/77#issuecomment-2014100229. But it's all good now.

rickwierenga:
Sorry for the long timeframe in merging this. I'll be faster in the future. One of the problems here, I think, is we used too much text to describe what we want. It would have been faster to communicate my idea by just giving you the code. Another thing is I simply need to respond faster. Responding doesn't take a lot of time, and you're blocked here until I respond, so I should prioritize it more over longer tasks.

BioCam:
haha don't worry @rickwierenga 

This PR brings in a big new chunk of code & functionality - I think the PR **should** take a bit of time!

I do think though that this is a prime example of how peer review should work, a constant back & forth to iron out ambiguities and come up with better solutions - ending up with a cleaner, objectively better product :)

Plus, none of us are making any money from this. So sometimes there will be higher priority tasks that slow things down a bit, that's okay.

PR #81, opened 2024-03-24
[STAR] require iswap is parked before running any core actions

jkhales:
core actions can collide with the iswap if not parked. Added the decorator to all core commands that make a firmware command but left out the convenience functions since they call the underlying decorated functions.

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR.py
@@ -1250,7 +1250,6 @@ class STAR(HamiltonLiquidHandler):
 
       await self.park_iswap(minimum_traverse_height_at_beginning_of_a_command=
                             int(self._traversal_height * 10))
-      self._iswap_parked = True
 
     if self.core96_head_installed:
       core96_head_initialized = await self.request_core_96_head_initialization_status()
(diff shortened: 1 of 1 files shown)


pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -1250,7 +1250,6 @@ async def setup(self):
 
       await self.park_iswap(minimum_traverse_height_at_beginning_of_a_command=
                             int(self._traversal_height * 10))
-      self._iswap_parked = True
```
jkhales, line 1253: this is set in `park_iswap` and should also in theory only be set there.

rickwierenga:
In the past I only decorated the higher level functions so that the "firmware-methods" just execute exactly what is written in the documentation. In my mind these higher level functions are the "smart" wrappers around the dumb firmware commands. Do you think we should be decorating the firmware methods instead? Happy to change if you argue that's better and as long as we're consistent.

jkhales:
> In the past I only decorated the higher level functions so that the "firmware-methods" just execute exactly what is written in the documentation. In my mind these higher level functions are the "smart" wrappers around the dumb firmware commands. Do you think we should be decorating the firmware methods instead? Happy to change if you argue that's better and as long as we're consistent.

I can see arguments for both sides.

For me personally, I want to use  the "low level" commands with safety guarantees. I'd rather everything be safe by default and have the decorator intercept an explicit "unsafe=True" kwarg to skip the check.

rickwierenga:
Let's do that. Adhering as strictly as possible to a manufacturer document is optimizing an artificial goal, and not a good design principle for PLR.

PR #82, opened 2024-03-25
Update basic.ipynb docs to reference PyLabRobot instead of PyHamilton

lennijusten:
Seems like this should reference PyLabRobot instead of PyHamilton, but I could be wrong?


rickwierenga:
Oops, a rare relic from the time PLR was a pyhamilton fork. Thanks for fixing!

PR #87, opened 2024-03-27
[Chatterbox] add support for iswap commands

DonSqualo:


ben-ray:
based additions to plr

rickwierenga:
The ChatterBoxBackend is supposed to be a hardware-agnostic dummy backend for all liquid handlers.

Perhaps we can subclass ChatterBox into Chatstar/Chatster which adds STAR-specific functionalities?

DonSqualo:
> The ChatterBoxBackend is supposed to be a hardware-agnostic dummy backend for all liquid handlers.
> 
> Perhaps we can subclass ChatterBox into Chatstar/Chatster which adds STAR-specific functionalities?

I was wondering the same. But why not have chatterbox support all liquid handlers then? I kind of think it should be completely hardware agnostic by always working

rickwierenga:
I think supporting all possible methods any backend is going to define in this single ChatterBoxBackend will be messy, and potentially undoable if there is a naming conflict between methods. I prefer to have one chattering backend for each robot.

DonSqualo:
refactored.
Why is CI failing?

rickwierenga:
Thanks!

PR #88, opened 2024-03-27
[libusb] add libusb to simplify installation

jkhales:
I found libusb_package debugging the "no backend available" error here https://github.com/pyusb/pyusb/blob/master/docs/faq.rst#how-do-i-fix-no-backend-available-errors

Seems like it's a very thin wrapper that adds the dll's in a findable place:

https://github.com/pyocd/libusb-package/blob/main/src/libusb_package/__init__.py

Seems like we could just use it from the start and avoid people from encountering this error in the first place?


rickwierenga:
Thank you for this!

backlink for future reference: https://forums.pylabrobot.org/t/missing-backend-when-connecting-to-a-hamilton-star/3277

PR #91, opened 2024-04-03
[backend] separate USB from Hamilton backend

jkhales:
this allows USBBackend to be a more general class.

### DIFF
--- a/pylabrobot/liquid_handling/backends/USBBackend.py
+++ b/pylabrobot/liquid_handling/backends/USBBackend.py
@@ -6,7 +6,7 @@ import time
 from typing import List, Optional, TYPE_CHECKING
 import libusb_package
 
-from pylabrobot.liquid_handling.backends.backend import LiquidHandlerBackend
+from pylabrobot.machine import MachineBackend
 
 try:
   import usb.core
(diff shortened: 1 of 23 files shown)


rickwierenga:
foundational change that will make future backends easier to add!

can you move the usb backend file to the top level / out of plr.liquid_handling?

jkhales:
> foundational change that will make future backends easier to add!
> 
> can you move the usb backend file to the top level / out of plr.liquid_handling?

good call. I consolidated it with the `machines.py` file into a top level `machines` directory.

pylabrobot/machines/resources.py
```diff
@@ -23,19 +25,6 @@ async def wrapper(self: Machine, *args, **kwargs):
     return await func(self, *args, **kwargs)
   return wrapper
 
-
```
jkhales, line 26: moved this to `backends.machine.py`

rickwierenga:
I think machines/machine.py could be a bit more descriptive than machines/resource.py (so we only have one resource.py in the entire project / the 'main' class name then matches the file name)

rickwierenga:
Thanks!

PR #92, opened 2024-04-05
[heat_shaking] implement the shaker for the hamilton heat shaker

jkhales:
Use the "Heater Shaker Module Command Specification" version 1.5 as a reference.

Tested locally.

### DIFF
--- a/pylabrobot/heating_shaking/hamilton.py
+++ b/pylabrobot/heating_shaking/hamilton.py
@@ -1,0 +1,15 @@
+from pylabrobot.heating_shaking.backend import HeaterShakerBackend
+from pylabrobot.machines.backends.usb import USBBackend
+
+class HamiltonHeatShaker(HeaterShakerBackend, USBBackend):
+  """
+  Backend for Hamilton Heater Shaker devices connected through
+  an HSB
+  """
+
+  def __init__(self, id_vendor:int = 2223, id_product:int = 32770) -> None:
+    HeaterShakerBackend.__init__(self)
+    USBBackend.__init__(self, id_vendor, id_product)
+
+  async def setup(self):
+    await USBBackend.setup(self)
--- a/pylabrobot/heating_shaking/hamilton.py
+++ b/pylabrobot/heating_shaking/hamilton.py
@@ -23,6 +23,7 @@ class HamiltonHeatShaker(HeaterShakerBackend, USBBackend):
     assert shaker_index >= 0, "Shaker index must be non-negative"
     self.shaker_index = shaker_index
     self.command_id = 0
+    self.lock_initialized = False
 
     HeaterShakerBackend.__init__(self)
     USBBackend.__init__(self, id_vendor, id_product)
@@ -48,21 +49,22 @@ class HamiltonHeatShaker(HeaterShakerBackend, USBBackend):
     self,
     speed: float = 800,
     direction: Literal[0,1] = 0,
-    acceleration: float = 1_000,
+    acceleration: int = 1_000,
   ):
     """
     speed: steps per second
     direction: 0 for positive, 1 for negative
     acceleration: increments per second
     """
-    assert 20 <= speed <= 2_000, "Speed must be between 20 and 2_000"
+    int_speed = int(speed)
+    assert 20 <= int_speed <= 2_000, "Speed must be between 20 and 2_000"
     assert direction in [0, 1], "Direction must be 0 or 1"
     assert 500 <= acceleration <= 10_000, "Acceleration must be between 500 and 10_000"
 
-    # Initialize plate lock
-    await self._initialize_lock()
-    self._move_plate_lock(PlateLockPosition.UNLOCKED)
-    self._start_shaking(direction=direction, speed=speed, acceleration=acceleration)
+    if not self.lock_initialized:
+      await self._initialize_lock()
+    await self._move_plate_lock(PlateLockPosition.UNLOCKED)
+    await self._start_shaking(direction=direction, speed=int_speed, acceleration=acceleration)
 
   async def stop_shaking(self):
     """ Shaker `stop_shaking` implementation. """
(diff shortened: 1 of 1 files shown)


pylabrobot/heating_shaking/hamilton.py
```diff
@@ -0,0 +1,81 @@
...
+  CLOCKWISE = 0
+  COUNTER_CLOCKWISE = 1
+
+class HamiltonHeatShaker(HeaterShakerBackend, USBBackend):
+  """
+  Backend for Hamilton Heater Shaker devices connected through
+  an HSB
+  """
+
+  def __init__(self, shaker_index:int, id_vendor:int = 2223, id_product:int = 32770) -> None:
```
jkhales, line 26: I found these numbers by trial and error, not sure if there is a better way.

pylabrobot/heating_shaking/hamilton.py
```diff
@@ -0,0 +1,81 @@
...
+  WAIT_FOR_STOP = "SW"
+
+class RotationDirection(Enum):
+  CLOCKWISE = 0
+  COUNTER_CLOCKWISE = 1
+
+class HamiltonHeatShaker(HeaterShakerBackend, USBBackend):
+  """
+  Backend for Hamilton Heater Shaker devices connected through
+  an HSB
```
jkhales, line 23: Not sure what the HSB acronym stands for 😅

pylabrobot/heating_shaking/hamilton.py
rickwierenga: Check lsusb

jkhales: yea, I think that's what I mean by trial and error. I found the object then retrieved the values from it.

pylabrobot/heating_shaking/hamilton.py
rickwierenga: https://forums.pylabrobot.org/t/add-carriers-to-hamilton-worktable-using-pylabrobot/2985/6 Heater Shaker Box?

pylabrobot/heating_shaking/hamilton.py
rickwierenga: perhaps we can keep these in hex?

it's very likely to be fixed for a given device (it's been the same for all STARs for example), so fine to hard code like this.

pylabrobot/heating_shaking/hamilton.py
```diff
@@ -0,0 +1,90 @@
...
+
+  def _send_command(self, command: Command, **kwargs):
+    args = "".join([f"{key}{value}" for key, value in kwargs.items()])
+    USBBackend.write(
+      self,
+      f"T{self.shaker_index}{command.value}id{str(self.command_id).zfill(4)}{args}",
+    )
+
+    self.command_id = (self.command_id + 1) % 10_000
+    return USBBackend.read(self)
```
rickwierenga, line 52: can we write a function to abstract this? it's the same for STAR

rickwierenga:
sorry for the delay, my comments were all "pending", I forgot to click submit, smh

pylabrobot/heating_shaking/hamilton.py
```diff
@@ -0,0 +1,81 @@
...
+
+  async def stop(self):
+    await USBBackend.stop(self)
+
+  def _send_command(self, command: Command, **kwargs):
+    args = "".join([f"{key}{value}" for key, value in kwargs.items()])
+    USBBackend.write(self, f"T{self.shaker_index}{command.value}id{str(self.command_id).zfill(4)}{args}")
+
+    self.command_id = (self.command_id + 1) % 10_000
+    return USBBackend.read(self)
```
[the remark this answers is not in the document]
jkhales: Yea, I guess the idea being that all firmware commands to Hamilton follow a similar format?

1. This could or could not be true. The "rule of three" suggests we wait for a third hamilton component before doing this.
2. If we did do this, It seems like one path would be to subclass `USBBackend` to `HamiltonUSBBackend` and add the command there.
3. There are some differences already, like the `tip_pattern` param and the parse firmware string. We could overwrite these for sure, but maybe end up being more complex
4. I think the command to unify would be `assemble_command`, but I'm curious about the decision to use bools there instead of directly feeding the firmware string. Seems like the abstraction boundary isn't clearly defined. 

I tend to be on the "wait and see" side of this.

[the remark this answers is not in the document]
rickwierenga: agreed

pylabrobot/heating_shaking/hamilton.py
```diff
@@ -0,0 +1,81 @@
...
+class PlateLockPosition(Enum):
+  LOCKED = 1
+  UNLOCKED = 0
+
+class Command(Enum):
+  INITIALIZE_PLATE_LOCK = "LI"
+  MOVE_PLATE_LOCK = "LP"
+  SHAKE = "SB"
+  STOP_SHAKING = "SC"
+  WAIT_FOR_STOP = "SW"
```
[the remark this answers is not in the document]
rickwierenga: In this case I'm pretty sure having the two letter commands in the methods is better. I definitely agree with this practice if I assume the 'magic numbers' point of view, but I don't :)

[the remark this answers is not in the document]
jkhales: OK, will rework!

rickwierenga:
the diff is a bit odd, but what do you think about this?

pylabrobot/heating_shaking/hamilton.py
```diff
@@ -0,0 +1,91 @@
...
+    speed: steps per second
+    direction: 0 for positive, 1 for negative
+    acceleration: increments per second
+    """
+    assert 20 <= speed <= 2_000, "Speed must be between 20 and 2_000"
+    assert direction in [0, 1], "Direction must be 0 or 1"
+    assert 500 <= acceleration <= 10_000, "Acceleration must be between 500 and 10_000"
+
+    # Initialize plate lock
+    await self._initialize_lock()
```
rickwierenga, line 63: is this needed every time or just setup?

jkhales: added an if to run it only once, tested and it works!

pylabrobot/heating_shaking/hamilton.py
```diff
@@ -0,0 +1,105 @@
...
+    return self._send_command("SW")
+
+  async def set_temperature(self, temperature: float):
+    raise NotImplementedError()
+
+  async def get_current_temperature(self) -> float:
+    raise NotImplementedError()
+
+  async def deactivate(self):
+    raise NotImplementedError()
```
jkhales, line 105: brought these back since the abstract super class demands it. Left them for the next adventurer who needs heat to implement 🙂

rickwierenga:
Thanks!

PR #93, opened 2024-04-10
[hamilton] make core grippers work on both hamilton decks

jkhales:


pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -4046,10 +4050,17 @@ async def get_core(self, p1: int, p2: int):
     if not 1 <= p2 <= self.num_channels:
       raise ValueError(f"channel_2 must be between 1 and {self.num_channels}")
 
+    deck_size = self.deck.get_size_x()
+    if deck_size == STARLET_SIZE_X:
+      xs = "07975"
+    elif deck_size == STAR_SIZE_X:
+      xs = "13375"
```
rickwierenga, line 4057: this we should be able to get from `deck.get_size_x()` right?

rickwierenga:
Thanks!

any particular reason for making `deck` optional here?

I'm gonna add this to all backends, then merge.

pylabrobot/liquid_handling/backends/hamilton/STAR.py
jkhales: one would think! But these actually turn out to be different values, perhaps because of excess padding by hamilton.

jkhales:
> Thanks!
> 
> any particular reason for making `deck` optional here?
> 
> I'm gonna add this to all backends, then merge.

optional prevents it from being a breaking change and also gives it typing.

jkhales:
> Thanks!
> 
> any particular reason for making `deck` optional here?
> 
> I'm gonna add this to all backends, then merge.

just noticed typing is off, fixing

jkhales:
fixed all typing except an issue with liquidhandler, but if you are going to add this to the PR, I will let you handle it! 🙏

rickwierenga:
making backends take decks in setup gives conflicts with `Machine.setup` ("This violates the Liskov substitution principle"), doing this in `__init__` would requires users to have a deck when instantiating a new backend. I think this can be a compromise with minimal code change that still generalizes

jkhales:
> making backends take decks in setup gives conflicts with `Machine.setup` ("This violates the Liskov substitution principle"), doing this in `__init__` would requires users to have a deck when instantiating a new backend. I think this can be a compromise with minimal code change that still generalizes

looks great to me!

rickwierenga:
this revealed that backends have some state (knowing about the deck) instead of solely being 'dumb command executors' is a good pattern to have! adds a bit of complexity, but only because it's necessary (so we're as simple as possible)

PR #94, opened 2024-04-10
[resources] add parent to plate lid

jkhales:
this allows the plate lid to be unassigned. This is useful, for example, when moving the lid off the plate.


rickwierenga:
Does super.assign_child_resource not do this?

jkhales:
> Does super.assign_child_resource not do this?

You're right, I'm going crazy 😅. Closing.

PR #95, opened 2024-04-12
[liquid_handling] fix unawaited machine backend setup

jkhales:
`lh.setup()` is broken since we don't await the machine backend setup


rickwierenga:
Smh

PR #96, opened 2024-04-12
Tecan dev branch

rickwierenga:
Merging from https://github.com/PyLabRobot/pylabrobot/pull/37/ (merged into plr:tecan-dev-branch) into plr:main


b-mahjour:
fyi, my absolute coordinates for the arm positions as well as their movement commands are reliant on the sequence of robotic arms connected to our tecan

```
# FIXME: what are these two lines for?
await self.liha.position_initialization_x()
await self.pnp.position_absolute_x_axis(1500)

```

Maybe there's a mechanism for this in PLR that I didn't realize, but for certain operations you have to move the other arms out of the way to prevent a crash. 

Our robot's arms are in this sequence: LiHA, PnP, RoMA

So it was just necessary (well, easiest) to have the LiHA and PnP as far left as possible when using the RoMA to move a plate. And likewise, I had the PnP and RoMA shift to the right if the LiHA was in operation.

Please let me know if there's an abstraction for this

PR #97, opened 2024-04-16
fix StarChatterBox

ben-ray:
made into an instance of type STAR to automatically inherit all methods as we add them to STAR.py 
added lh.setup chatter to make it work


rickwierenga:
as it should have been!

PR #100, opened 2024-04-17
[heat_shaking] split out locking

jkhales:
It's not obvious we always want the plate locked or unlocked when we shake.

### DIFF
--- a/pylabrobot/heating_shaking/hamilton.py
+++ b/pylabrobot/heating_shaking/hamilton.py
@@ -61,19 +61,24 @@ class HamiltonHeatShaker(HeaterShakerBackend, USBBackend):
     assert direction in [0, 1], "Direction must be 0 or 1"
     assert 500 <= acceleration <= 10_000, "Acceleration must be between 500 and 10_000"
 
-    if not self.lock_initialized:
-      await self._initialize_lock()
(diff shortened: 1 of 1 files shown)


pylabrobot/heating_shaking/hamilton.py
```diff
@@ -61,20 +61,25 @@ async def shake(
...
 
   async def stop_shaking(self):
     """ Shaker `stop_shaking` implementation. """
     await self._stop_shaking()
     await self._wait_for_stop()
-    await self._move_plate_lock(PlateLockPosition.LOCKED)
 
   async def _move_plate_lock(self, position: PlateLockPosition):
+    if not self.lock_initialized:
+      await self._initialize_lock()
```
rickwierenga, line 73: i thought this was only required on `setup`?

jkhales: We may not want to run it on setup, so this is more flexible, but either way.

rickwierenga: I generally try to initialize everything on setup to get super predictable protocols

jkhales:
oh nice, was just sitting down to fix this. Thank you for doing it for me!

rickwierenga:
very welcome! thank *you* for contributing this!

PR #106, opened 2024-04-23
Updating the Opentrons temperature control module

lennijusten:
Checking in work so that @rickwierenga can see/review progress.

### DIFF
--- a/pylabrobot/liquid_handling/backends/opentrons_backend.py
+++ b/pylabrobot/liquid_handling/backends/opentrons_backend.py
@@ -1,5 +1,5 @@
 import sys
-from typing import Dict, Optional, List, cast
+from typing import Dict, Optional, List, cast, Union
 
 from pylabrobot.liquid_handling.backends.backend import LiquidHandlerBackend
 from pylabrobot.liquid_handling.errors import NoChannelError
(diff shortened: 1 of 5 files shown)


lennijusten:
Still todo:
- Allow users to specify which aluminum block plate sits atop the temp control module
- Test the implementation!

lennijusten:
@rickwierenga I updated the PR to instead get the aluminum tube racks from the [Opentrons shared data](https://github.com/Opentrons/opentrons/tree/26929a2d92715ab23d7a4832f2548a2bab565d11/shared-data/labware/definitions/2). 

I also added a `tube_rack` arg to the `OpentronsTemperatureModuleV2`. By default, this is `None` indicating no tube rack on the temp controller module.

lennijusten:
Do you see anything obvious that still needs to be done here? I'll hopefully be able to test in the lab later this week.

rickwierenga:
> I updated the PR to instead get the aluminum tube racks from the [Opentrons shared data](https://github.com/Opentrons/opentrons/tree/26929a2d92715ab23d7a4832f2548a2bab565d11/shared-data/labware/definitions/2).

nice!

> Do you see anything obvious that still needs to be done here? I'll hopefully be able to test in the lab later this week.

I think you should be slightly careful when testing, as the resource definition might differ from what I manually calibrated before. Probably best to not use real bacteria while testing.

PR #109, opened 2024-04-25
[STAR] add hotel commands

jkhales:
This also adds the idea of an "unsafe" attribute to put dangerous functions

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR.py
@@ -1055,158 +1055,6 @@ def _dispensing_mode_for_op(empty: bool, jet: bool, blow_out: bool) -> int:
   else:
     return 3 if blow_out else 2
 
-class UnSafe:
-  """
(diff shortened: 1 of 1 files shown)


pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -1092,6 +1244,12 @@ def __init__(
     self._core_parked: Optional[bool] = None
     self._extended_conf: Optional[dict] = None
     self._traversal_height: float = 245.0
+    self._unsafe = UnSafe(self)
+
+  @property
+  def unsafe(self):
```
jkhales, line 1250: Just noticed I forget to add the return type, will do tomorrow!

rickwierenga:
```diff
- """ actions that are not guaranteed to be safe. Use with care!"""
+ """ Actions that have a higher risk of damaging the robot. Use with care! """
```

clarifies that not everything else in PLR _is_ guaranteed to be safe. this is the user's power and responsibility. as discussed, we should not set users up for damaging their robots.

PR #110, opened 2024-04-26
[hamilton] add low level command to rotate iSWAP

jkhales:


rickwierenga:
just checking: should this be under unsafe?

jkhales:
> just checking: should this be under unsafe?

i put it under unsafe because if you rotate it can hit things, especially if you are near the edge of the deck

rickwierenga:
by this logic, essentially all iswap commands should be `unsafe`. where do you draw the line here?

jkhales:
> by this logic, essentially all iswap commands should be `unsafe`. where do you draw the line here?

hmm, my impression was that normal iswap commands where careful about not moving off deck? If not happy to move into star. Also makes me wonder if the hotel commands deserve to be in unsafe as well? I'm not sure where the line should be either 😄

rickwierenga:
we use the move_plate command primarily for moving plates into the deck-adjacent plate reader.

did you damage the robot developing the hotel commands?

jkhales:
> did you damage the robot developing the hotel commands?

No, but I could have if I wasn't paying close attention.

In general, I don't see a strong separating line on the danger of various commands. My only opinion is that if we are going to make a separation, then using an `unsafe` attribute is better (since it is more explicit) than a flag.

rickwierenga:
> No, but I could have if I wasn't paying close attention.

I'd say that rotating the iSWAP has behavior is exactly as you expect, so it's not necessarily unsafe. When calling it, there are no unexpected side effects. Can that be the dividing line? (still largely subjective - this can be a heuristic)

>  My only opinion is that if we are going to make a separation, then using an unsafe attribute is better (since it is more explicit) than a flag.

You convinced me of this. Current discussion is just around what should go in there.

jkhales:
> > No, but I could have if I wasn't paying close attention.
> 
> I'd say that rotating the iSWAP behavior is exactly as you expect, so it's not necessarily unsafe. When calling it, there are no unexpected side effects. Can that be the dividing line? (still largely subjective - this can be a heuristic)
> 
> > My only opinion is that if we are going to make a separation, then using an unsafe attribute is better (since it is more explicit) than a flag.
> 
> You convinced me of this. Current discussion is just around what should go in there.

cool, will move this to the main STAR!

rickwierenga:
great change, setting the structure for further documentation of Hamilton STARs that will encourage more people to hack on it

PR #113, opened 2024-04-26
Fix temp controller

lennijusten:
Somehow the changes with `OTModule` were not in the last PR. Something probably got messed up with git. Does this look right?


rickwierenga:
looking good!

PR #119, opened 2024-04-30
Change tc to t for temperature controller in temperature.ipynb

lennijusten:


rickwierenga:
(in favor of tc:thermocycler) (everything is a 'controller')

PR #120, opened 2024-04-30
allowing serialization and deserialization of Machines and their backends

maraxen:


pylabrobot/machines/machine.py
```diff
@@ -49,6 +49,48 @@ def __init__(
...
+        The class with the given name, or `None` if no such class exists.
+      """
+
+      if cls.__name__ == class_name:
+        return cls
+      for subclass in cls.__subclasses__():
+        subclass_ = find_subclass(class_name=class_name, cls=subclass)
+        if subclass_ is not None:
+          return subclass_
+      return None
```
rickwierenga, line 81: this should be able to use `get_resource_class_from_string(class_name, Machine)`

maraxen: I edited this so it imports the function from from resources

maraxen: realized that this is not going to work due MachineBackend only inheriting from ABC, reverted the changes

PR #126, opened 2024-05-09
update STAR.py docstrings

ben-ray:
just adding swap_speed

current unknown docstrings can be discovered & updated in STAR.py by navigating through the pages in the HTML Help topics when using Hamilton CO-RE Liquid Editor & Method Editor


rickwierenga:
thanks!

PR #129, opened 2024-05-14
Alpaqua MAGNUM FLX integration

BioCam:
Hi everyone,

This is a small PR to integrate the first magnetic rack into the PLR resource library.

Of course, this has to start with the super powerful Alpaqua MAGNUM FLX.

Since it behaves similar to MFX_modules, i.e. it only takes one child_resource at a predefined location I implemented it as a MFX_module.

There are some assumptions made with this implementation:
- Your chosen well-bottom geometry does not enter the magnetic rings for more than about 1.6 mm (each ring actually has a depth of around 2-2.1mm). This is actually dependent on the geometry of the plate's bottom well -> a V-shaped bottom with steep angles will enter deeper into the ring magnet than a U-shaped bottom. => perfect alignment will have to be made by end users at the moment, but in my hands, no bottom shape resulted in an error message so far.
- The ring magnets are always fixed in place. The plates placed on top of it might have different `dx` and `dy` values. This means that slight offsets are needed to place the plate into the exactly correct position for some plates whose well position massively differs from the standard `Cos_96_Rb`.

Happy to discuss and iterate this implementation but thought I'd get the ball rolling to get magnetic racks integrated into PLR :)


rickwierenga:
great addition!

> no bottom shape resulted in an error message so far.

what error?

> The plates placed on top of it might have different dx and dy values. This means that slight offsets are needed to place the plate into the exactly correct position for some plates whose well position massively differs from the standard Cos_96_Rb.

Why is this not handled automatically when assigning a plate as a child resource of the `Alpaqua_96_magnum_flx`?

BioCam:
> > no bottom shape resulted in an error message so far.

> what error?

If someone was to place a flat-bottom well on this magnet then no part of the well would enter the round magnet. In that case the tip would crash into the flat-bottom causing a Z drive error.

> Why is this not handled automatically when assigning a plate as a child resource of the Alpaqua_96_magnum_flx?
Because the `assing_child_resource` command will always assign a new plate to the same pre-defined coordinate, which I currently set to semi-universal 
```python
child_resource_location=Coordinate(-1.0, 0, 26.5)
```

However, this assumes is oversimplified:
The ring magnets have their own `dx` and `dy` values.
Any plate has also their own, e.g. `dx=6.9` & `dy=6.9` for `Cos_96_Rd`.
This is now decently aligned.

But the Azenta4titudeFrameStar_96_wellplate_skirted has `dx=11.0` & `dy=8.7`.

But importantly the spacing between wells is still the same, `item_dx=9` & `item_dy=9`, because they are both SLAS/SBS microtiter plates.

rickwierenga:
do you think it's appropriate to raise a `NotImplementedError` when the child resource is not conform?

rickwierenga:
also, `skirt_height` is defined on MFXModule, but not saved/used, causing a linting error. Do you think this should be stored as an attribute?

BioCam:
> do you think it's appropriate to raise a `NotImplementedError` when the child resource is not conform?

Regarding the flat bottom issue, I think this is a great idea! (though I'm not sure how to implement it)
Regarding the variation in `dx` and `dy` values for different plates, I don't see how this could be done.

To be honest, I think there is a point to be made to generate a new PLR class, probably called `plate_adapter`.
This class would cover how a plate is moved onto an adapter that has holes for the wells of the plate to sit on in the SLAS-standardised 9x9mm grid.
With a bit of logic it can move the origin of the plate to the correct position (basically just knowing where the bottom left adapter hole is and subtracting the plates dx and dy from there).
This would be representing what is truly happening, and VENUS clearly deals with this in the background somehow because moving plates onto adapters is no problem.

To exemplify plate_adapters:
- on-deck thermocycler
- magnetic racks
- adapters for semi- and non-skirted plates
- adapters for shakers
- adapters for temperature control units

th-2005215003

csm_2016-0110_excangeable_adapter_0883852988
Screenshot 2024-05-14 at 23 01 28


But this might be too much for this PR?




> also, `skirt_height` is defined on MFXModule, but not saved/used, causing a linting error. Do you think this should be stored as an attribute?

Yes, I think having it as a standard attribute for anything that can take a plate will be very useful for the next steps in correcting for dz vs skirt_height issues.

rickwierenga:
Pushed a hacky `_Alpaqua_96_magnum_flx` class so we can have specialized behavior. Unique behavior in resources warrant specific classes, and in this case there is no other resource with this behavior so I made it unique to this one specific labware item.

I like the idea of a `PlateAdapter`, which could cover the "single-resource-taking" resource we discussed [earlier](https://github.com/PyLabRobot/pylabrobot/pull/77). But probably a second PR is better, as you mentioned.

BioCam:
This looks perfect for this PR; thank you very much, @rickwierenga!

Happy to merge, and work on a `PlateAdapter` that might supersede the current implementation in the near future :)

PR #132, opened 2024-05-16
Added functions to TubeRack and Tube and added them to the visualizer

joelaforet:
Hi Rick! We met at SLAS and now I'm working in the Chory Lab. We're using PLR to build some teaching exercises for a BME class at Duke, and I wanted to add some functionality to the visualizer. We plan to use the OT2, and tube racks are very common. I saw that there was no support for this in the visualizer, so I added some functions. Previously, tube racks and tubes were defaulted to "Resource" objects, and there was no volume tracking. I copied over the functions from Plate and Well, and tweaked them to make sense for tube racks. This is my first time contributing to OpenSource, so please let me know if you have any comments/suggestions. Thanks!


rickwierenga:
Hi Joe! That's great to hear and thanks for the PR! Super excited about this class.

This reminds me we should probably refactor a little bit, since so much code is copied, but that's a worry for another time.

rickwierenga:
great first OSS contribution btw!

PR #133, opened 2024-05-16
[STAR] fix core gripper location

jkhales:
previously there we intermittent pick up failures. tested this a bunch and it seems robust. Also visual inspection.

Hamilton had it's pipettes calibrated last week so I think this is legit.

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR.py
@@ -4069,7 +4069,7 @@ class STAR(HamiltonLiquidHandler):
     if deck_size == STARLET_SIZE_X:
       xs = "07975" # 1360-797.5 = 562.5
     elif deck_size == STAR_SIZE_X:
-      xs = "13375" # 1900-1337.5 = 562.5
+      xs = "13385" # 1900-1337.5 = 562.5, plus a manual adjustment of + 10
     else:
       raise ValueError(f"Deck size {deck_size} not supported")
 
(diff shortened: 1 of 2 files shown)


pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -4069,7 +4069,7 @@ async def get_core(self, p1: int, p2: int):
     if deck_size == STARLET_SIZE_X:
       xs = "07975" # 1360-797.5 = 562.5
     elif deck_size == STAR_SIZE_X:
-      xs = "13375" # 1900-1337.5 = 562.5
+      xs = "13385" # 1900-1337.5 = 562.5, plus a manual adjustment of + 10
```
jkhales, line 4072: should I tweak the starlet by the same amount?

more generally should there be a constants.py file containing a "core_hotel" dict? This would ensure that the get and put commands use the same value.

rickwierenga: could this be a constant at the top of the file? / what is the rationale for putting this in a separate file if it's only used in STAR.py?

jkhales: it's just a massive file 🙃  either way works for me.

rickwierenga: it's massive, but i don't really see a clean way of splitting it up. moving tiny things to different files seem more confusing to me

rickwierenga:
can't verify because we don't use the core gripper, but trust this works for you, and others can revisit to create a dynamic solution if this doesn't work for them.

PR #134, opened 2024-05-17
Visualizer showing volume updates for add_liquid() and remove_liquid() calls.

joelaforet:
Found a bug where the visualizer doesn't change colors when the add_liquid() and remove_liquid() functions are called. set_liquids() works fine, so I moved the callback check into these functions too.


rickwierenga:
good catch!

PR #135, opened 2024-05-21
Thermo scientific 96 1200ul rd integration

BioCam:
Hi everyone,

A small PR to integrate `ThermoScientific_96_1200ul_Rd` (Fisher Scientific/Thermo Fisher cat. no.: 10243223/AB1127).

Empirically tested volume function & dimensions verified on a Hamilton STARlet system.

PR #136, opened 2024-05-21
Calculate height from volume in container (LLF part 1)

BioCam:
Hi everyone,

## PR content
Currently PLR does not allow liquid level following (LLF) during aspiration and dispensation.

This is the first PR in a series to expose this feature in PLR:

This PR adds the `resources/height_functions.py` in which functions are stored that represent the inverse from  `resources/volume_functions.py`.

Using these functions the **height of a liquid** inside **container of known geometry** can be calculated solely requiring the **liquid_volume to be known**.

## "Big picture" idea

The step-by-step procedure for liquid level following could be:
- search for liquid (`height_start`)
- calculate volume based on LLD (`volume_start`)
- subtract volume to be aspirated OR add volume to be dispensed (`volume_end`)
- calculate height_end on volume_end and these new **height_functions** (`height_end`)
- move tip into liquid x mm
- start aspiration/dispensation command given `height_start-x` and `height_end-x`

-> **resulting in liquid level following**

I'm very happy to discuss alternatives, and additions to this proposed implementation.

## Current status

Due to time constraints I "inverted" 3 out of the 5 functions currently stored in `resources/volume_functions.py`. 
The new functions are:

- `calculate_liquid_height_in_container_2segments_square_vbottom` <- quite straight forward geometry task
- `calculate_liquid_height_in_container_2segments_square_ubottom` <- turns out calculating the subvolume of a hemisphere / spherical cap is a bit more complicated than I initially thought due to the non-linear nature of the volume. I tried out a couple of numerical solutions and turns out a binary solver is the fastest solver (as always)... at least in my hands :)
- `calculate_liquid_height_container_1segment_round_fbottom` <- straight forward

I'll aim to "invert" the remaining 2 in the next couple of days.

## Proposed next steps

After this I propose:
1. adding a `compute_height_from_volume` method and a `_compute_height_from_volume` attribute to the `class Well` in `resources/well.py`, using the existing `compute_volume_from_height` as its counterpart.
2. going through the PLR resource library and updating labware.

Please let me know if I missed anything or you have ideas for optimisation.

If someone gets to any of these tasks before I do, please get cracking :)

### DIFF
--- a/pylabrobot/resources/height_functions.py
+++ b/pylabrobot/resources/height_functions.py
@@ -1,0 +1,145 @@
+# General use functions for calculating the height of 
+# a liquid inside different container geometries
+
+import math
+
+
+def calculate_liquid_height_in_container_2segments_square_vbottom(
+  x: float,
+  y: float,
+  h_pyramid: float,
+  h_cube: float,
+  liquid_volume: float
+) -> float:
+  """
+  Calculate the height of liquid in a container consisting of an upside-down
+  square pyramid at the bottom and a cuboid on top. The container has the
+  same x and y dimensions for both the pyramid and the cuboid.
+
+  The function calculates the height based on the given liquid volume.
+
+  Parameters:
+    x (float): The base length of the square pyramid and cube in mm.
+    y (float): The base width of the square pyramid and cube in mm.
+    h_pyramid (float): The height of the square pyramid in mm.
+    h_cube (float): The height of the cube in mm.
+    liquid_volume (float): The volume of the liquid in the container in cubic millimeters.
+
+  Returns:
+    float: The height of the liquid in the container in mm.
+  """
+  base_area = x * y
+
+  # Calculating the full volume of the pyramid
+  full_pyramid_volume = (1/3) * base_area * h_pyramid
+
+  if liquid_volume <= full_pyramid_volume:
+    # Liquid volume is within the pyramid
+    scale_factor = (liquid_volume / full_pyramid_volume) ** (1/3)
+    liquid_height = scale_factor * h_pyramid
+  else:
+    # Liquid volume extends into the cube
+    cube_liquid_volume = liquid_volume - full_pyramid_volume
+    cube_liquid_height = cube_liquid_volume / base_area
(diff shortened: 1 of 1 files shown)


rickwierenga:
shared `_height_of_volume_in_spherical_cap` between calculate_liquid_height_in_container_2segments_square_ubottom and calculate_liquid_height_in_container_2segments_round_ubottom.

I assume for square (cuboid) + u bottom (hemisphere) this is not entirely accurate, but is the current implementation _good enough_?

BioCam:
> shared `_height_of_volume_in_spherical_cap` between calculate_liquid_height_in_container_2segments_square_ubottom and calculate_liquid_height_in_container_2segments_round_ubottom.

Thank you, that is a lot cleaner!
Just added a docstring and a volume check (in case one wants to use the function by itself).


> I assume for square (cuboid) + u bottom (hemisphere) this is not entirely accurate, but is the current implementation _good enough_?

This is a very good question and yes, you are of course correct: this is a slight simplification. The hemisphere appears to be most commonly connected through a slant in the bottom cuboid edges, and it seems that the hemisphere's diameter is actually slightly smaller than the x dimension of the cuboid. 

Here is a little mock up I made of a [ThermoScientific_96_1200ul_Rd](https://github.com/PyLabRobot/pylabrobot/pull/135) well which I added to PLR yesterday (**left** is the real, physical container geometry, **right** is the simplified geometry used for volume calculations):

image

So far though, this definition has been very accurate in my hands. If someone else or I notices a problem with this implementation then I'd recommend we generate height_ and volume_functions for a container with 3 segments (i.e. hemisphere, frustum of a rectangular pyramid, cuboid).
But I'd like to try to avoid such complex definitions if possible.

rickwierenga:
sweet! that makes a lot of sense - just wanted to double check.

do you think this is ready to merge?

BioCam:
> `surface_following_distance` is probably the parameters in STAR you're looking for

Thank you, that is exactly what I was looking for! This massively simplifies my proposal of the "'Big Picture' Idea" I mentioned above.

PR #138, opened 2024-05-23
Cross Contamination Tracker

joelaforet:
This is the first iteration of a simple cross contamination tracker. I implemented it as an optional flag that users can turn on if they are interested. It adds a new variable to the volume tracker class called liquid_history, where we store the identity of all liquids that this tracker has encountered. I also added methods to set, erase, and update the liquid_history.

The "CrossContamination" exception is thrown when attempting to aspirate from a location that contains liquid not identical to the current tracker's liquid_history. 

Liquid history is updated on `dispense` calls. If you dispense into a well, the liquid history becomes everything now inside of that well.

Let me know what you think!


rickwierenga:
nice! initial pass looks good.

it appears this code broke some tests. could you please look into that? This will enable quicker iteration than me looking at your code manually right now :) (It is possible to run this locally with `make lint`, `make typecheck` and `make test`).

I also see some diffs that seem unrelated to this change. Perhaps you need to sync the fork?

I'll look more in depth when ci is happy :)

joelaforet:
Thanks for bearing with me as I noob my way through GitHub!

rickwierenga:
what do you think about moving the cross contamination check out of LiquidHandler into the VolumeTracker? `add_liquid` seems reasonable. This way, if we get more liquid-changing devices in the future, your code will also work there. It also keeps LiquidHandler a bit more organized. Happy to discuss as always

joelaforet:
> what do you think about moving the cross contamination check out of LiquidHandler into the VolumeTracker? `add_liquid` seems reasonable. This way, if we get more liquid-changing devices in the future, your code will also work there. It also keeps LiquidHandler a bit more organized. Happy to discuss as always

I agree! I kept the check in the dispense part of the liquid_handler.py because we need both the tip's volume tracker and the container resource's volume tracker to pull the liquid history and do the update. However, the aspirate totally makes sense to be moved to add_liquid, and now the code is much easier to follow.

pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -913,6 +915,10 @@ async def dispense(
     for op in dispenses:
       if does_volume_tracking():
         if not op.resource.tracker.is_disabled:
+          # Update the liquid history of the tip to reflect new liquid
+          if (not op.tip.tracker.is_cross_contamination_tracking_disabled) and \
+              not op.resource.tracker.is_cross_contamination_tracking_disabled:
+            op.tip.tracker.liquid_history = op.resource.tracker.liquid_history
```
rickwierenga, line 921: should `op.tip.tracker.liquid_history` not be a `set` that is extended?

pylabrobot/resources/volume_tracker.py
```diff
@@ -97,6 +134,18 @@ def add_liquid(self, liquid: Optional["Liquid"], volume: float) -> None:
       raise TooLittleVolumeError(
         f"Container has too little volume: {volume}uL > {self.get_free_volume()}uL.")
 
+    # Update the liquid history tracker if needed
+    if not self._is_cross_contamination_tracking_disabled:
+      # If this is a new liquid, and if the tip isn't clean
+      if liquid is not None and (liquid not in self.liquid_history) and self.liquid_history:
+        raise CrossContaminationError(
+            f"Attempting to aspirate {liquid} with "
+            f"a tip contaminated with {self.liquid_history}.")
```
rickwierenga, line 143: this code will actually also be called on the well when dispensing liquid into it. In that case, this error should not be raised (= it should be possible to dispense a new liquid into a well without raising). Perhaps I was wrong in suggesting this code go here, and your initial idea to put it in LiquidHandler was better.

pylabrobot/liquid_handling/liquid_handler.py
joelaforet: It can be. In my head, I want the liquid history to be literally changed to the destination's history because it seems more permanent. I don't think the behavior of the code will change in either case though.

pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -738,8 +752,24 @@ async def aspirate(
...
           op.resource.tracker.remove_liquid(op.volume)
         for liquid, volume in reversed(op.liquids):
+
+          # Cross contamination check
+          if does_cross_contamination_tracking():
+            if check_clean(liquid, op.tip.tracker.liquid_history):
+              raise CrossContaminationError(
+                f"Attempting to aspirate {next(reversed(op.liquids))[0]} with \
+                    a tip contaminated with {op.tip.tracker.liquid_history}.")
+
```
rickwierenga, line 771: Could you explain what's going on here?

You loop over the liquids in the aspiration, and call check_clean for each. It seems that in the loop it is checked that none of the liquids in the aspiration have been in the tip before. (If the tip has seen any liquid at all.) If you aspirated some liquid with this tip before, and then aspirate from the same container again, does this (incorrectly?) raise an error?

There seems to be duplication with line 758, where this is checked only for the last liquid in the aspiration.

I thought cross contamination checking would check that the tip has not touched liquids that are not in the well/container. This would mean looping over the liquids in the tip and ensuring there are none which are not also in the well. If the liquid history of the well and tip are represented as sets, this can be done by checking that the `tip_history` is a strict subset of `well_history`.

joelaforet: > You loop over the liquids in the aspiration, and call check_clean for each. It seems that in the loop it is checked that none of the liquids in the aspiration have been in the tip before. (If the tip has seen any liquid at all.) If you aspirated some liquid with this tip before, and then aspirate from the same container again, does this (incorrectly?) raise an error?

Your interpretation is correct. Check clean takes in a tip and returns true if the candidate aspiration liquid is not in the liquid history AND if liquid history is not empty. We want check_clean to return false for a successful aspiration. I think a better name for this would be check_contaminated. We only want the check to fire if the candidate liquid is not already in the liquid history, but only if the liquid_history is not empty. Imagine a tip touches red dye. Red dye is now stored in the liquid_history. If we want to aspirate from a well containing blue dye, we check that blue dye is not in liquid_history. If it is, and the liquid history isn't empty, then the tip is allowed to aspirate, so the check_clean returns false. Otherwise, we return true and except. This does not return an error if we aspirate from the same container twice.

> There seems to be duplication with line 758, where this is checked only for the last liquid in the aspiration.

I think you only need one check in either spot. I removed the first one. We do the check for every liquid entering the tip.

> I thought cross contamination checking would check that the tip has not touched liquids that are not in the well/container. This would mean looping over the liquids in the tip and ensuring there are none which are not also in the well. If the liquid history of the well and tip are represented as sets, this can be done by checking that the `tip_history` is a strict subset of `well_history`.

Since we do the check for every liquid entering the tip, I believe that is the same as checking the strict subset condition.

pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -55,6 +57,18 @@
...
+     would result in cross contamination"""
+
+  return (liquid not in liquid_history and bool(liquid_history))
+
+def check_updateable(src_tracker, dest_tracker):
+  """Helper function used to check if it is possible to update the
+     liquid_history of src based on contents of dst"""
+
+  return (not src_tracker.is_cross_contamination_tracking_disabled) and \
+      bool(dest_tracker.is_cross_contamination_tracking_disabled)
```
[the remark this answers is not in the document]
joelaforet: This is a logic error from my use of double negatives. Doh!

rickwierenga:
it seemed the implementation of `check_contaminated` (`check_clean`) got lost, added it back in https://github.com/PyLabRobot/pylabrobot/pull/138/commits/877769d65aa7828e00219c9faab3882f317c0f03.

Let's try to merge this soon.

Some things I think would be good to do:

- Figure out if the for loop in the aspiration should be over the well or tip liquids / if a set difference makes the implementation easier.
- Write some simple tests for confidence + stability in the future.
- Could you explain the following code?

```python
          # Update the liquid history of the tip to reflect new liquid
           if check_updatable(op.tip.tracker, op.resource.tracker):
             op.tip.tracker.liquid_history = op.resource.tracker.liquid_history
```

joelaforet:
> Some things I think would be good to do:
> 
> * Figure out if the for loop in the aspiration should be over the well or tip liquids / if a set difference makes the implementation easier.

I agree that the set difference is easier and more straight-forward to think about. Made the change.

> * Write some simple tests for confidence + stability in the future.

I added a few tests, but my check for if attempting to aspirate from a well with partial overlap fails. The desired behavior is throw error if tip_history = {'dye_1', 'dye_2'} and we want to aspirate from a spot where well_history = {'dye_1'}. This would contaminate the well with dye_2. It correctly throws the error in my Jupyter notebook though, so I'll take a deeper look at it. If you have any suggestions on other tests to add I would love to hear them!

> * Could you explain the following code?
> 
> ```python
>           # Update the liquid history of the tip to reflect new liquid
>            if check_updatable(op.tip.tracker, op.resource.tracker):
>              op.tip.tracker.liquid_history = op.resource.tracker.liquid_history
> ```
When we perform a dispense step, I update the state of the tip's liquid history to be identical to the liquid_history of the place where the dispense happened. My intent was to wipe the tip_history and re-populate it with everything now in resource_history, but this code may not actually be doing that.

rickwierenga:
> I added a few tests, but my check for if attempting to aspirate from a well with partial overlap fails. The desired behavior is throw error if tip_history = {'dye_1', 'dye_2'} and we want to aspirate from a spot where well_history = {'dye_1'}. This would contaminate the well with dye_2. It correctly throws the error in my Jupyter notebook though, so I'll take a deeper look at it. If you have any suggestions on other tests to add I would love to hear them!

`test_aspirate_from_well_with_partial_overlap` runs correctly for me!

rickwierenga:
anything left to do before merging?

(sorry this took so long, will be faster in the future)

rickwierenga:
it seems in https://github.com/joelaforet/pylabrobot/commit/1ab756b7ebf50001efa8dabdb9ca922f54e84c1e, a lot of new changes from plr:main were rolled back, for some reason. fixing that now.

rickwierenga (approved):

joelaforet:
> it seems in [joelaforet@1ab756b](https://github.com/joelaforet/pylabrobot/commit/1ab756b7ebf50001efa8dabdb9ca922f54e84c1e), a lot of new changes from plr:main were rolled back, for some reason. fixing that now.

Looks good to me! Sorry, must have pulled from main incorrectly.

rickwierenga:
Thanks for contributing this! It will save many experiments.

As you've probably seen, I recently started plr propaganda on Twitter. Is it ok with you if I share a link to this PR? Do you have an account i should @ / how should I give credit?

joelaforet:
Hey Rick! Referencing my LinkedIn page would be great! https://www.linkedin.com/in/joseph-laforet-jr-64845a203/

PR #143, opened 2024-05-26
Implement pedestal for CarrierSite & MFXModule (fix PLR Plate location 1)

BioCam:
Hi everyone,

This PR's goal is to add pedestal information into the `CarrierSite` & `MFXModule` classes.

This is step 1 of a series to fix the true location of plates in PLR, please see this forum discussion, [PLR Carrier z-Height Issue](https://labautomation.io/t/plr-carrier-z-height-issue/3056/32), for more information.

## Summary of the issue

In short, liquid handling systems place plates onto their deck at specified sites which PLR calls `CarrierSite`.
The majority of these exhibit an elevation in the center of the `CarrierSite` which PLR is now on calling `pedestal` (shown in cyan in this infographic).
carrier_elevation_naming

Some important information:
- The pedestal top z coordinate is regarded as the top  of the `CarrierSite`.
- The distance between the top of the pedestal (cyan) and the `CarrierSite` section that the skirt/frame of a plate would sit on (yellow) is known as the `plate_size_z`.
- A plate placed on top of this `CarrierSite` can either 
  1. sit with its skirt/frame on the `CarrierSite` directly, `if` the clearance height below the wells (`= dz`) > `plate_size_z`
  2. sit with its wells flush on the pedestal, `if` the `dz < plate_size_z`

At the moment PLR is completely unaware of these relationships.

As a result correctly defined plates placed on a `CarrierSite` without a pedestal will create crashes between tips and the bottom of plate wells.

Additionally, plates with a large `dz` (e.g. shallow-well plates) will crash.

Furthermore, the true origin of plates placed on `CarrierSite` with a pedestal is below what PLR believes it to be, generating inaccuracies in plate movement that have to be compensated with adjusted grip heights (which will have to be different for `CarrierSite`s with and without pedestals for the same plate).

## PR content

Focusing solely on adding pedestal information, I added:
- the new class attribute `pedestal_size_z` to both `CarrierSite` and `MFXModule`, with default set to `None`,
- updated functions `create_carrier_sites` & `create_homogeneous_carrier_sites` to enable integration of this new attribute in the definition of entire `Carrier`s,
- updated the `MFXModule` docstring, and
- showcase an updated `Carrier` by adding `pedestal_size_z` information to `PLT_CAR_L5AC_A00`

```python
def PLT_CAR_L5AC_A00(name: str) -> PlateCarrier:
  """ Carrier for 5 deep well 96 Well PCR Plates 
  Hamilton cat. no.: 182090
  """
  return PlateCarrier(
    name=name,
    size_x=135.0,
    size_y=497.0,
    size_z=130.0,
    sites=create_homogeneous_carrier_sites([
        Coordinate(4.0, 8.5, 86.15),
        Coordinate(4.0, 104.5, 86.15),
        Coordinate(4.0, 200.5, 86.15),
        Coordinate(4.0, 296.5, 86.15),
        Coordinate(4.0, 392.5, 86.15)
      ],
      site_size_x=127.0,
      site_size_y=86.0,
      site_pedestal_size_z=-5.0
    ),
    model="PLT_CAR_L5AC_A00"
  )
```

This information can now be called quickly:
```python
# CarrierSite example
lh.deck.assign_child_resource(plt_carrier_1, rails=1)

plt_carrier_1[0].pedestal_size_z
# returns: -5

# MFXModule example
mfx_carrier_1 = MFX_CAR_L5_base(name='mfx_carrier_1')
mfx_carrier_1[1] = mfx_dwp_module_1 = MFX_DWP_rackbased_module(name="mfx_dwp_module_1")

mfx_dwp_module_1.pedestal_size_z
# returns: None (default behaviour)
```

NB.: The `pedestal_size_z` is a **negative** float, reflecting the fact that the `CarrierSite` z coordinate is `86.15` mm and the pedestal is moving downward in z from that height.

## Next Steps

Nothing about the current behaviour of PLR has been changed in this PR; only information has been added to enable changed behaviour in the future.

To get to that point we have to execute a series of further steps:

- "unlock"/integrate necessary information:
    1. pedestal information in `CarrierSite` -> this PR
    2. `well_z_thickness` in `ItemizedResource.create_equally_spaced()`
    3. correct wrong `dz` in `ItemizedResource.create_equally_spaced()` (currently `dz == well_z_thickness`, has to be changed to `dz == plate.size_z - well.size_z - well_z_thickness`, i.e. the clearance height below a well on a non-pedestal `CarrierSite`/flat surface).
- enable correct placement of correctly-defined plate in plate assignment process
    1. I added some #TODOs in the `asssign_child_resource` of the `CarrierSite` and `MFXModule` classes for this in the current PR.


rickwierenga:
I'm so happy to see code for this topic. Thank you for your continued effort in fixing this.

> the new class attribute pedestal_size_z to both CarrierSite and MFXModule, with default set to None,

Like we discussed on the forum, I think we should raise an error when a CarrierSite of MFXModule is initialized with `pedestal_size_z=None`, referring to this PR and asking people to submit the pedestal_size_z.

What do you think about implementing something like a PlateAdapter/PlateHolder here?

pylabrobot/resources/ml_star/ims/Hamilton_carrier_naming_guide.png
rickwierenga, line 1: do you think we can put this in the docs so it is displayed on the website?

BioCam: Yeah, I had the same question and wasn't 100% sure.

But in the end I added it because this information is freely available on the internet, and/or some deductions based on what Eric shared.

I first made a Markup table for this table actually, especially because this is old information, missing even the MFX carriers.
A table is also easier to maintain.

But in the end it the Markdown table looked ugly and readability was poor 😅

PLR is an open-source project without profit generation for developers.
If anyone at Hamilton wants it to be taken down - even though this is free advertising for them that makes it easier for customers to purchase carriers from them - then they can just request it to be removed and we will of course comply.

BioCam:
> Like we discussed on the forum, I think we should raise an error when a CarrierSite of MFXModule is initialized with `pedestal_size_z=None`, referring to this PR and asking people to submit the pedestal_size_z. The parameter in the initializer would still accept an optional, but the attribute would not be optional.

I 100% agree and did remember our conversation.
The reason why I haven't added the raise Error yet was intentional: I think we need to make the necessary changes incrementally, ensuring first that all information is set up correctly and then add the error.

This will also give people the opportunity to add their carrier's information in the meantime.
If we switch this on instantly it will break everyones scripts, and while this is good for PLR progression purposes (it forces people to add information right now), I think it better to give users an "adaption phase" before turning on the error raise.

(Tangent: Ideally we would add pedestal information to other PLR machines as well. If I'm not mistaken the OT-2 sites/slots don't have pedestals, i.e. Hamilton plate definitions will all break on that machine?)

> What do you think about implementing something like a PlateAdapter/PlateHolder here?

I'd love to but that is quite a lot of implementation work, and I think it goes beyond the scope of this PR because a pedestal is different from a PlateAdapter, though I agree that a PlateAdapter has to have pedestal_size_z information.

BioCam:
> Can we do this in this PR? Otherwise I'm happy to merge this into a dev branch in this repo if you prefer. Once it's ready, we can merge that into plr:main.
> 
> I think having `is not None` checks for an attribute that seems required is suboptimal for plr:main.

I see what you are saying but I do not think all PLR users together have all the possible carrier types that exist available to them to define all their pedestal heights.
This means the `None` condition will be part of the implementation any way... or, instead of raising an error we raise a warning?

I don't think a separate dev branch is necessary tbh, and I would rather give `pedestal_size_z` definition ability to everyone through main, since that is the master branch that everyone syncs to, i.e. slowly building up main is more likely to be elicit contributions, I believe.



> I'll ask Ben & jon to send me the pedestal heights for their carriers and also add ours.

That is amazing, thank you! Do you want to keep this PR open until then or generate some `update pedestals for carrier xyz` when you have them?



> I am hopeful CarrierSite will disappear.

I have the feeling I asked this before but cannot remember: Why?

pylabrobot/resources/ml_star/mfx_modules.py
```diff
@@ -19,39 +19,42 @@ class MFXModule(Resource):
...
     category: Optional[str] = "mfx_module",
+    pedestal_size_z: Optional[float] = None,
     model: Optional[str] = None):
     super().__init__(name=name, size_x=size_x, size_y=size_y, size_z=size_z, category=category,
       model=model)
     # site where resources will be placed on this module
     self._child_resource_location = child_resource_location
     self._child_resource: Optional[Resource] = None
-    self.skirt_height = skirt_height
+    self.pedestal_size_z: Optional[float] = pedestal_size_z
```
rickwierenga, line 56: I think we should make this non optional, and raise an error if the parameter given is `None`

pylabrobot/resources/carrier.py
```diff
@@ -14,17 +14,24 @@ class CarrierSite(Resource):
...
-    super().__init__(name=name, size_x=size_x, size_y=size_y, size_z=size_z, category=category,
-      model=model)
+    category: str = "carrier_site",
+    pedestal_size_z: Optional[float] = None,
+    model: Optional[str] = None):
+    super().__init__(name=name, size_x=size_x, size_y=size_y, size_z=size_z,
+      category=category, model=model)
     self.resource: Optional[Resource] = None
+    self.pedestal_size_z: Optional[float] = pedestal_size_z
+    # TODO: add self.pedestal_2D_offset if necessary in the future
```
rickwierenga, line 24: this too

rickwierenga:
> I see what you are saying but I do not think all PLR users together have all the possible carrier types that exist available to them to define all their pedestal heights.

No. I was thinking we just have people PR the data as they go. To facilitate this, I think we should raise errors when people try to instantiate a resource from a definition that is incomplete. The error should link to this thread for an easy explanation and perhaps continued discussion. (My belief is that if people don't like to contribute this, they can go back to an earlier version, which is also free.)

Having `None` for the parameter has to be part of this implementation, but the attribute should not be if we want to merge to plr:main. Having `if pedestal_size_z is not None` checks is slightly too experimental for plr:main imo. I highlighted the two lines that I think should be updated here.

Since this PR does not actually make use of `site_pedestal_size_z` yet, I was thinking perhaps it's worth merging into a dev branch. That way, when we start asking people for this measurement they get immediate benefit. If we merge with the change I described above, they would have to do this without getting immediate benefit. I'm fine with either.

> Do you want to keep this PR open until then or generate some update pedestals for carrier xyz when you have them?

Let's try to merge this PR asap, into main or a dev.

> I have the feeling I asked this before but cannot remember: Why?

As you pointed out previously, a lot of code is shared between CarrierSite and MFXModule. Many resources take a singular Plate and it would be good to consolidate this into a single abstraction. You find the perfect name for it.

BioCam:
> Let's try to merge this PR asap, into main or a dev.
> 
> >

Alright, I'm convinced - let's make contribution non-optional and merge with main, adding a link to this PR in the error message :)



> > I have the feeling I asked this before but cannot remember: Why?
> 
> As you pointed out previously, a lot of code is shared between CarrierSite and MFXModule. Many resources take a singular Plate and it would be good to consolidate this into a single abstraction. You find the perfect name for it.

I see... mhmmm, I need to think a bit about this: 
`CarrierSite`s are parts of a `Carrier` with a specific function (to host plates) while `MFXModule`s are separate objects but with the same function... I am not sure how to best compromise the physical difference between these two because it would mean giving up either of the two terms, `site` or `module`, which both make sense in their respective use cases.

One problem I am seeing is that both `CarrierSite` and `MFXModule` are both used for **plates** and for **tips** even though physically they are very different from one another.
If I'm not mistaken, this means we have to state all tip_carriers as having a pedestal_size_z=0 to not break them even though they don't have a pedestal?

rickwierenga:
> One problem I am seeing is that both CarrierSite and MFXModule are both used for plates and for tips even though physically they are very different from one another.

I totally lost track of this smh.

It does not make sense for these to have a `pedestal_size_z` attribute. I have been playing with a `class PlateCarrierSite(CarrierSite):`, but this requires some refactoring of `create_carrier_sites`. I have been thinking to remove this method altogether actually, in favor of a `create_equally_spaced` (currently in itemized resource) (and perhaps `create_equally_spaced_2d`). Having a single way of creating grids of resources seems simpler than having a general one and then a specialized one for `CarrierSite`s. I aim to have code tomorrow.

On second thought, I think the parameter `pedestal_size_z` should not even be optional. The issue really exists in methods like `PLT_CAR_L5AC` which lack the information. Implementing it this way is more code, but very doable because we have these `/tools/make_resources` scripts that will do it automatically. Will do.

rickwierenga:
sorry for the delay, i was moving back.

made the following edits:
- https://github.com/PyLabRobot/pylabrobot/commit/d16d995b01ab6b6eda51b323271bdea5a11b953a: 
create_equally_spaced_{x,y,2d}. It's actually slightly annoying to use this because the carriers might not all be evenly spaced. I think it's better to keep the create_carrier_sites methods. Still a sensible refactor imo.
- https://github.com/PyLabRobot/pylabrobot/commit/4756af76b1533da9b944cf079b80a14cd78f0762: create carrier sites with klass, opens up the possibility for PlateCarrierSite with pedestal info.

I will continue with this and create a PlateCarrierSite here.

BioCam:
Awesome, thank you @rickwierenga!

Good to merge from my end :)

PR #149, opened 2024-05-30
OT2 'reservoirs'

joelaforet:
OpenTrons has a set of labware distinct from plates called "reservoirs". To my understanding, these are called "troughs" in Hamilton speak. The trough class, however, is just an extension of container, rather than resource, so it can't have multiple wells as children. OpenTrons reservoirs come in single and 12-well varieties (others too but not in standard labware library). I added this as a new object and implemented it in the visualizer. It functions the same as a 'plate', but the way that load_opentrons_resource() is set up, we need something that handles 'reservoir's. Open to any tweaks/suggestions you may have!


rickwierenga:
like this?
image

joelaforet:
Yep! They group those with the single well ones too.

joelaforet:
That's effectively what I did. The problem is that the .json file that contains the labware information calls these pieces of labware 'reservoirs', not plates. Without changing the code for how other pieces of OT labware are loaded in, the easiest way is to just give it the 'reservoir' class its looking for. If it helps, I'm referring mainly to the logic of the ot_definition_to_resource() function in resources/opentrons/load.py

pylabrobot/resources/opentrons/load.py
```diff
@@ -143,11 +170,20 @@ def make_tip() -> Tip:
...
       )
+    if display_category == "reservoir":
+      return Reservoir(
+        name=name,
+        size_x=size_x,
+        size_y=size_y,
+        size_z=size_z,
+        items=cast(List[List[Well]], wells),
+        model=data["metadata"]["displayName"]
+      )
```
rickwierenga, line 181: would it be possible to return a Plate here?

rickwierenga:
I think returning a Plate in the location I suggested above (which is totally reasonable imo because these are just oddly shaped Plates) makes things a lot simpler.

joelaforet:
That worked! Wish I would've seen it sooner.

rickwierenga:
always good to expand the labware library!

PR #152, opened 2024-05-31
Creating PlateAdapter

BioCam:
Hi everyone,

In this PR I created a new `Resource` subclass called `PlateAdapter` which enables the use of...
- semi-skirted plates,
- non-skirted plates,
- magnetic racks,
- plate adapters for shakers and temperature control machine, and
- ODTCs
...in PyLabRobot (PLR).

## Summary of the issue

**Issue 1: Missing Plate Support due to lack of PlateAdapter**

Plates require robust placement on the deck. So far only "skirted" plates have been supported by PLR (to my knowledge). Semi- and non-skirted plates require a `PlateAdapter` that allows the wells to directly be placed into, rather than the plate being correctly positioned by its plate_skirt.

**Issue 2: Complexity of automated placement of Plate onto PlateAdapter**

`PlateAdapters` are a block of metal or plastic in which holes are arranged in a grid that follows the [ANSI SLAS 4-2004 (R2012)](https://www.slas.org/SLAS/assets/File/public/standards/ANSI_SLAS_4-2004_WellPositions.pdf) Well Position Standards (if you encounter a Plate or PlateAdapter that doesn't, please throw it in the bin).
This means the wells are spaced as follows: 96-well plate, 9x9mm^2; 384-well plate, 4.5x4.5mm^2; 1536-well plate, 2.25x2.25mm^2.

However, there are 2 complications to this standard:
1. The well-grid is well standardized, the placement of the well-grid on the plate is not.
2. PlateAdapters have a well standardized hole-grid to fit the well-grid, but PlateAdapters come in all shapes and sizes.

The positioning of the well-grid on a plate is given in PLR using `Well.dx` and `Well.dy`, denoting how far away the origin of the bottom-left well ("H1" for a 96-well plate) is from the origin of the plate. I have previously generated this infographic to visualise this and facilitate plate definition generation (though this shows a 24-well plate the same position arguments apply):
plate_definition_explainer

PlateAdapters have holes to fit the wells from the Plate but, to my knowledge, there is no geometric definition for PlateAdapters and their hole-grid. I will therefore use an adapted terminology of the Plate geometry for PlateAdapters.
E.g. `adapter_hole_size_x` is the PlateAdapter equivalent to `Well.size_x` and `adapter_hole_dx` is the equivalent to `Well.item_dx`.

As a result of this, each, `Plate` and `PlateAdapter` have their own `dx` and `dy` and `Hole/Well.size_x`.

**The mission of a** `Plate`**->**`PlateAdapter` **movement command is to align the 2D center of the bottom-left Plate.well with the 2D center of the bottom-left PlateAdapter.hole, despite their differences in relatve grid positioning.**

Here is a visualisation of what happens if this is ignored and a Plate is moved onto a magnetic rack before this PR:

PLATE_overlays

## Content of this PR
I added...
- a new module called `resources/plate_adapter.py` which defines the new `PlateAdapter` class and its complex self-adjustment of various `Plate`s,
- necessary changes in `liquid_handler.py` to integrate the new `PlateAdapter` in the liquid handling and resource movement handling ecosystem,
- a small update in the `Plate` definition to clarify whether a plate_type is "skirted", "semi-skirted", or "non-skirted", and
- 2x examples of `PlateAdapter` instances:
    - a revamped `Alpaqua_96_magnum_flx` definition
    - a definition for the `Hamilton_96_adapter_188182` which can be placed on top of an HHC or TCC_site (stored inside a new `resources/ml_star/plate_adapters.py` module)

`PlateAdapter` is fundamentally different to other `Resource` subclasses because the location of a child that is assigned to it **does not have to be hardcoded**.
Instead the `PlateAdapter.assign_child_resource()` method calculates the necessary location adjustment automatically.
However, this method is used to assign the `Plate` placed onto the `PlateAdapter` to a specific location in the assignment process, which is fine when one starts an automation script with a  `Plate`on a `PlateAdapter`.
If one uses `lh.move_plate(Plate, PlateAdapter)` then this command requires information of where to place the Plate before this assignment has taken place. This is why I had to create a second location adjustment calculation directly in `liquid_handler` if the `to` location is an instance of `PlateAdapter`.

## Next steps

At the moment PLR only has the two `PlateAdapter`s defined in this PR. In the future we should add other adapters, and ODTCs (e.g. he Opentrons Aluminium PlateAdapter actually has an SLAS footprint and could therefore be easily integrated to enable semi-/non-skirted plate use on standard plate_carriers).

Furthermore, I had to re-calculate some Plate information because the required information like `Well.item_dx` is, to my understanding, currently not callable. I think we should discuss how to retain all information from the definition of a `Plate`.

In the change of the `Alpaqua_magnum_flx` from its own specialised `Resource` subclass to a `PlateAdapter` I had to remove the error generation if a flat-bottom well is used. We might want to find a way to re-implement this without generating a sub-sub class?

The current z position to place the `Plate` onto the `PlateAdapter` has to be empirically evaluated.
This is because z-height information is currently broken in PLR's `Plate` definition.
This is not an issue for the current PR, in my opinion, but should be changed to an auto-adjustment, similar to the x- and y-adjustment, in the near future.


pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -1878,6 +1880,19 @@ async def move_plate(
...
+        for well_n in plate.children]))
+      y_locations = sorted(OrderedDict.fromkeys([well_n.location.y
+        for well_n in plate.children]))
+      plate_dx, plate_dy = x_locations[0], y_locations[0]
+      well_size_x = plate.children[0].get_size_x()
+      well_size_y = plate.children[0].get_size_y()
+      plate_x_adjustment = to.dx - plate_dx + to.adapter_hole_size_x/2 - well_size_x/2
+      plate_y_adjustment = to.dy - plate_dy + to.adapter_hole_size_x/2 - well_size_y/2
+      adjusted_plate_anchor = Coordinate(plate_x_adjustment, plate_y_adjustment, to.dz)
+      to_location = to.get_absolute_location() + adjusted_plate_anchor
```
rickwierenga, line 1895: is there a way to make this a private function on `PlateAdapter` (eg `_compute_child_location`) to minimize code duplication between PlateAdapter and LiquidHandler?

pylabrobot/resources/plate.py
```diff
@@ -51,7 +51,8 @@ def __init__(
     category: str = "plate",
     lid_height: float = 0,
     with_lid: bool = False,
-    model: Optional[str] = None
+    model: Optional[str] = None,
+    plate_type: Literal["skirted", "semi-skirted", "non-skirted"] = "skirted"
```
rickwierenga, line 55: For semi-skirted plates, is it not important to have the height of the skirt? / do these 3 discrete options capture all required information?

I imagine that for:
- skirted platess: `skirt_height > well_size_z`
- semi-skirted platess: `0 < skirt_height < well_size_z`
- non-skirted platess: `skirt_height = 0`
with `skirt_height` being a float.

pylabrobot/resources/plate_adapter.py
```diff
@@ -0,0 +1,175 @@
...
+    self.adapter_hole_dy = adapter_hole_dy
+    self.site_pedestal_z = site_pedestal_z
+
+  @property
+  def child_resource_location(self) -> Optional[Coordinate]:
+    return self._child_resource_location
+
+  def assign_child_resource(
+    self,
+    resource: Plate,
```
rickwierenga, line 110: This leads to the type error. According to the "Liskov substitution principle", APIs cannot be more specialized/constrained for subclasses:

`PlateAdapter` is a `Resource` and so the following code should be valid:

```python
my_resource: Resource = PlateAdapter(...)
```

but then

```python
my_resource.assign_child_resource(non_plate)
```

would fail, even though `Resource.assign_child_resource` allows it.

In https://github.com/PyLabRobot/pylabrobot/pull/143/commits/d076b471a0df9c784c4cb10cf8323c215266b7e1#diff-c76c0cd2964c9c9ee4af709b9dd64d713a464a50ccede9f2c0a798c0f994bf70R200 I solved this with a dynamic type check, which isn't the cleanest, but works for now I guess 🤷‍♂️

pylabrobot/resources/plate_adapter.py
```diff
@@ -0,0 +1,175 @@
...
+    reassign: bool = True
+  ):
+    """Assign a Plate to a PlateAdapter. If `location` is not provided,
+    the resource will autoadjust the placement location based on the
+    PlateAdapter-Plate relationship."""
+
+    if self._child_resource is not None and not reassign:
+      raise ValueError(f"{self.name} already has a child resource assigned")
+    if not isinstance(resource, Plate):
+      raise ValueError("Only plates can be assigned to Alpaqua 96 magnum flx.")
```
rickwierenga, line 121: I think this should be a TypeError

pylabrobot/resources/plate_adapter.py
```diff
@@ -0,0 +1,175 @@
...
+
+    # TODO: have discussion oon whether to transfer flat bottom error checking
```
rickwierenga, line 123: Can we simply check `well.bottom_type != WellBottomType.flat`?

pylabrobot/resources/plate_adapter.py
```diff
@@ -0,0 +1,175 @@
...
+    y_locations = sorted(OrderedDict.fromkeys([well_n.location.y
+      for well_n in resource.children]))
+
+    def calculate_well_spacing(float_list: List[float]):
+      """ Calculate the difference between every x and x+1 element in the list of floats. """
+      if len(float_list) < 2:
+        return None
+      differences = [round(float_list[i+1] - float_list[i],2) for i in range(len(float_list) - 1)]
+      if differences[0] is None:
+        raise ValueError("well spacing has to be uniform, and cannot be None")
```
rickwierenga, line 139: how could this be None?

rickwierenga: I think it would crash on line 137.

pylabrobot/resources/plate_adapter.py
```diff
@@ -0,0 +1,175 @@
...
+    plate_item_dx = abs(calculate_well_spacing(x_locations))
+    plate_item_dy = abs(calculate_well_spacing(y_locations))
+    well_size_x = resource.children[0].get_size_x()
+    well_size_y = resource.children[0].get_size_y()
+    # true_dz = resource.get_size_z() - resource.children[0].get_size_z()
+
+    # Well positioning check
+    assert (plate_item_dx, plate_item_dy) in [(9.0, 9.0), (4.5, 4.5), (2.25, 2.25)], \
+      "PlateAdapter only accepts plates with a well spacing of " + \
+      "9x9, 4.5x4.5 or 2.25x2.25mm^2 (ANSI SLAS 4-2004 (R2012): Well Positions)"
```
rickwierenga, line 155: Is there a way to check this with the PlateAdapter's own attributes? This would not only make it more extendible, but also be a more precise check because currently eg 384 well plates are allowed on 96 hole-spaced adapters.

pylabrobot/resources/plate_adapter.py
```diff
@@ -0,0 +1,175 @@
...
+
+    # Calculate adjustment to place center of H1_plate on top of center of H1_adapter
+    plate_x_adjustment = self.dx - plate_dx + self.adapter_hole_size_x/2 - well_size_x/2
+    plate_y_adjustment = self.dy - plate_dy + self.adapter_hole_size_x/2 - well_size_y/2
```
rickwierenga, line 159: isn't it always true that `self.adapter_hole_size_x == well_size_x`?

pylabrobot/liquid_handling/liquid_handler.py
BioCam: Yes, I think that is a great idea!

I will implement this tonight

pylabrobot/resources/plate_adapter.py
BioCam: I am not sure: There might be flat-bottom wells that people want to place on a flat-bottom plate_adapter (e.g. for temperature control and/or shaking)?

pylabrobot/resources/plate_adapter.py
BioCam: > how could this be None?

This is a workaround that came  about in the type checking process:
```python
x_locations = sorted(OrderedDict.fromkeys([well_n.location.x
      for well_n in resource.children]))
    y_locations = sorted(OrderedDict.fromkeys([well_n.location.y
      for well_n in resource.children]))
```

If I remember correctly, this generated `None` if the adapter doesn't have a Plate on it, in the adapter assignment command.

pylabrobot/resources/plate_adapter.py
BioCam: Ohh that is a good point:
Considering this, the assertion should actually be a check of the _cross-compatibility_ of **adapter + plate_to_be_assigned_as_a_child**.

Since the information of both:
- adapter: `adapter_hole_dx`, `adapter_hole_dy`
- plate: `plate_item_dx`, `plate_item_dy`
is known in the `PlateAdapter.assign_child_resource()` method, yes, I can write a cross-compatibility check and replace it with this simple assertion.

pylabrobot/resources/plate_adapter.py
BioCam: No, it is not.

1) `well_size_x` does not include the side thickness of the well - the same issue that is breaking PLR's Plate definition in the z dimension, missing information regarding the well_material_thickness (just that in x-y it doesn't matter for aspiration/dispensation)

2) the `self.adapter_hole_size_x` can be significantly different from `well_size_x`:
2.i. it can be much smaller, e.g. in the magnetic rack -> then only a part of the well actually enters the PlateAdapter.hole with the remaining well sticking out of it
2.ii. it can be bigger (e.g. by the thickness of the well material, as mentioned in point 1, exemplified by the Hamilton TCC plate adapter)

That was one of the reasons I added two very different plate adapters in this PR, to showcase the variety of PlateAdapters that are covered by this new class :)

pylabrobot/resources/plate_adapter.py
BioCam: I agree, I just changed it

pylabrobot/resources/plate.py
BioCam: Yes, you are completely right.

I did not want to change too much about `Plate()` in this PR because it is secondary here to `PlateAdapter`.
That's why I only added this simple 3 word additional information.

I think it would be better to add user information that requires rewriting information in the `Plate()` class in a separate information.
Since here I just add standalone information, and establish that so far all `Plate` instance in PLR have been of "skirted" type, I thought it would be a nice little push to work on it in the future :)

pylabrobot/liquid_handling/liquid_handler.py
BioCam: Implemented and tested in the lab

pylabrobot/resources/plate_adapter.py
BioCam: cross-compatibility check now implemented

pylabrobot/resources/plate_adapter.py
rickwierenga: Ah. I thought PlateAdapter here always has holes. But perhaps this is intended to be a more general class? What are the attributes like .adapter_hole_size_x in such cases?

pylabrobot/resources/plate_adapter.py
rickwierenga: Thanks for the explanation!

pylabrobot/resources/plate.py
rickwierenga: Good to know this is a WIP. I shall add a comment

pylabrobot/resources/plate_adapter.py
rickwierenga: i am slightly confused by why OrderedDict is used here.

what I meant:
```python
differences = [round(float_list[i+1] - float_list[i],2) for i in range(len(float_list) - 1)]
if differences[0] is None:
```

how can `round(float_list[i+1] - float_list[i],2)` return None?

pylabrobot/resources/plate_adapter.py
rickwierenga: While I understand this follows slas formats, it seems artificially constrained. Why is it more complicated to support all possible sizes? (I might say it's simpler)

pylabrobot/resources/plate_adapter.py
BioCam: Yes, I think to house a `Plate` there will always be holes. But the holes can follow the geometry of any well.bottom that exists.
Bioshake has a [really nice website here](https://www.bioshake.com/collections/adapter), showcasing a variety of adapters that are designed for efficient temperature transfer and/or efficient shaking transmission into the plate.

I am not sure because I have not seen a flat-bottom well plate inside a flat-bottom adapter but I am not sure I want to exclude the option here.
Tbh placing a U-bottom plate inside a V-bottom adapter is in that sense also not a great idea... I'm not sure this kind of safety check can easily be implemented though.

pylabrobot/resources/plate_adapter.py
BioCam: > i am slightly confused by why OrderedDict is used here.

ahhh that one I can answer easily: I needed a `np.unique()` without having to make PLR dependent on numpy, OrderedDict is part of the standard library `collections`, I am happy to change it to a viable alternative though

> how can `round(float_list[i+1] - float_list[i],2)` return None?
Tbh I cannot remember. I agree that it should never come to this. We can remove it and see whether anything breaks?

pylabrobot/resources/plate_adapter.py
BioCam: I see your point, I have removed the well-grid and hole-grid spacing checks and only left the cross-compatibility check between the two grids.

pylabrobot/resources/plate_adapter.py
rickwierenga: hb a [`set`](https://docs.python.org/3/tutorial/datastructures.html#sets)? :)

> We can remove it and see whether anything breaks?
sounds good

pylabrobot/resources/plate_adapter.py
rickwierenga: how about [this](https://www.bioshake.com/collections/adapter/products/1808-1021-adapter-for-microplate-flat-bottom-standard) from the linked website? I'm not sure if this is the level of complexity we need to take in from the start, but perhaps worth discussing how we might think about this in the future.

pylabrobot/resources/plate_adapter.py
```diff
@@ -0,0 +1,183 @@
...
+    site_pedestal_z: float,
+    adapter_hole_dx: float = 9.0,
+    adapter_hole_dy: float = 9.0,
+    category: Optional[str] = None,
+    model: Optional[str] = None
+    ):
+    if adapter_hole_dx not in {9.0, 4.5, 2.25}:
+      raise ValueError("adapter_hole_dx must be one of 9.0, 4.5, or 2.25")
+    if adapter_hole_dy not in {9.0, 4.5, 2.25}:
+      raise ValueError("adapter_hole_dy must be one of 9.0, 4.5, or 2.25")
```
rickwierenga, line 87: does this make things easier?

pylabrobot/resources/plate_adapter.py
rickwierenga: went ahead with https://github.com/PyLabRobot/pylabrobot/pull/152/commits/5d3b8208728e39fb3aedfd6fb186e3dc6abbee1f

pylabrobot/resources/plate_adapter.py
BioCam: No, I thought it makes PlateAdapter generation safer but I'm happy to remove it to make the class simpler?

rickwierenga: It seems to me the constraint is artificial, adding a tiny bit of complexity without adding much value. Let's remove it if you agree

BioCam:
Just verified that everything works well on the machine. Happy to merge from my side 🚀

BioCam:
Thank you @rickwierenga, another great peer review session! :D

rickwierenga:
thank you for developing on this!

PR #154, opened 2024-06-04
fix: close event loop that starts in thread

jt05610:
## Problem
The `_run_async_in_thread` method in the `LiquidHandler` class was causing `ResourceWarning` issues related to unclosed sockets. This warning was being raised at the end of pytests in a library of mine that is starting to use PyLabRobot: [graphmix](https://github.com/jt05610/python-graphmix). 

Screenshot 2024-06-04 at 5 07 18 PM

## To reproduce
Set filterwarnings = error in pytest.ini:

```ini

[pytest]
python_files = *_tests.py

filterwarnings =
    error
```

## Solution
Added `loop.close()` to the end of the callback in _run_async_in_thread to close the event loop that gets created for the callback.

## Testing
- Ran the existing test suite, and all tests pass with the applied changes.
- Verified that the `ResourceWarning` is no longer raised at the end of Python tests.

Please review and consider merging this pull request to resolve the socket warning issue.

### DIFF
--- a/pylabrobot/liquid_handling/liquid_handler.py
+++ b/pylabrobot/liquid_handling/liquid_handler.py
@@ -182,6 +182,7 @@ class LiquidHandler(Machine):
       loop = asyncio.new_event_loop()
       asyncio.set_event_loop(loop)
       loop.run_until_complete(func(*args, **kwargs))
+      loop.close()
 
     t = threading.Thread(target=callback, args=args, kwargs=kwargs)
     t.start()


rickwierenga:
thanks!

wanna share more about graphmix?? it looks very interesting.

jt05610:
For sure! Its from one of my PhD projects and is for designing protocols that account for things like serial dilutions, container dead volumes, pipette min/max transfer volumes, etc. As the name implies, it uses graphs to describe solutions, and it can be used to solve the composition of anything in a network of fluid transfers. I use Pint everywhere to avoid naked numbers.

I'm also using graphs to map transfers with liquid handlers so that I can say, for example, "I want concentrations X, Y, and Z of stock solution A diluted with solution B, and to have the final volume of everything be 100 µL," then it will solve the transfer volumes and initial volumes of my stock and diluent. 

I'm still moving and improving it from a private repo, so the graph transfer mapping isn't in graphmix yet, but should be by the end of the week!

This is what it looks like when it solves the volumes for the standard curve from the BCA assay protocol, which was my original motivation:

standard_curve

The most concentrated stock is in 0:A1, the diluent is 0:A2, concentrations decrease from 0:A4 (red) through 0:A12, (blue) and everything ends up being 100 µL

The graph was generated with just:

```python
BCA_STANDARD_CURVE = (
    StandardCurveStep(ref="A", factor=1),
    StandardCurveStep(ref="B", factor=0.75),
    StandardCurveStep(ref="C", factor=0.5),
    StandardCurveStep(ref="D", source="B", factor=0.5),
    StandardCurveStep(ref="E", source="C", factor=0.5),
    StandardCurveStep(ref="F", source="E", factor=0.5),
    StandardCurveStep(ref="G", source="F", factor=0.5),
    StandardCurveStep(ref="H", source="G", factor=0.5),
    StandardCurveStep(ref="I", factor=0),
)

G = graph.standard_curve(
    G=G,
    stock=source,
    diluent=diluent,
    final_volume=Q_("100 µL"),
    grid="0",
    positions=source.location.pos_increment,
    steps=BCA_STANDARD_CURVE,
)

graph.solve(G)
```
I have a high level AbstractProtocolBuilder base class that uses the builder pattern, then use that to create more crazy things like this: 

Formulation%20Prep%20with%20Sampling

which 

1. makes a product
2. samples it for different assays
3. makes standard curves for the assays
4. prepare the product that was made for the assays


Thanks for your work on PyLabRobot, it's a perfect fit for what I've been working on!

rickwierenga:
Wow that's really cool & this would be super useful to integrate with PLR/another automated liquid handling system! Perhaps a premature question: is there any ordering to the edges or is that found using a separate algorithm? I look forward to you open-sourcing and developing it more :)

jt05610:
Great question! Since I'm using directed acyclic graphs, I'm just using [topological sorting](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.dag.topological_sort.html) to order the edges. Then, I group by source node, and will assign a priority so that aqueous buffers and larger volumes are transferred first when the sources are at the same topological level. I'm also doing binning so that if 50 µL of something needs to go to 4 places, it will pickup 200 µL and distribute instead of 4 separate 50 µL transfers.

We can definitely find a way to integrate everything! I'll share on your forum once I get everything moved over if that's alright. I have a few more related projects that I am going to open source and publish papers on as well.

Thank you!

PR #155, opened 2024-06-05
Add config file

jt05610:
# Problem

There is no module-level configuration for `pylabrobot`. This results in .log files being created in the current working directory every time `pylabrobot` is imported.

# Solution

Add `pylabrobot.config` module for project configuration. A `Config` object was created that just configures logging for now. A `pylabrobot.ini` file is found in the current directory. If not found, it is searched for in all parent directories. If it still is not found, it gets created at either the project level that contains the .git directory, or the current directory. This gets called whenever `pylabrobot` is imported.

I also made the testing logs collect in a test_logs directory for each directory that has tests by adding a module-level pytest fixture in `conftest.py`

Currently, `.ini` files are supported, however, I wrote it so that .json, .yaml, .toml, etc can quickly be implemented and supported.

The current configuration that is used for pylabrobot is now in `pylabrobot.ini` for this project.

Also, I created `.editorconfig` with the current project formatting style.

# Example config

```ini
[logging]
level = DEBUG
log_dir = .

```

(Equivalent to the current configuration)

# Testing 

* Wrote `config_test.py` and tested reading and writing .ini-formatted 
* Ran existing test suite to make sure this doesn't break it.


pylabrobot/__init__.py
```diff
@@ -1,15 +1,61 @@
...
+    log_dir = Path(log_dir)
+  if not log_dir.exists():
+    log_dir.mkdir(parents=True)
+  logger = logging.getLogger("pylabrobot")
+  logger.setLevel(level)
+
+  now = datetime.datetime.now().strftime("%Y%m%d")
+  # remove file handler if it exists
+  if len(logger.handlers) > 0:
+    logger.handlers.clear()
```
rickwierenga, line 36: what is the rationale for this?

pylabrobot/config/service/writer.py
```diff
@@ -0,0 +1,18 @@
...
+class ConfigWriter(Generic[T], ABC):
+  """ConfigWriter is an abstract class for writing a Config object to a
+  stream."""
+
+  open_mode: str = "w"
+
+  @abstractmethod
+  def write(self, w: T, cfg: Config):
+    """Write a Config object."""
+    raise NotImplementedError
```
rickwierenga, line 18: is raise needed when method is marked as abstarctmethod?

pylabrobot/config/service/file.py
```diff
@@ -0,0 +1,48 @@
...
+  def read(self, r: Union[str, Path]) -> Config:
+    """Read a Config object from a file."""
+    if isinstance(r, str):
+      r = Path(r)
+    file_ext = r.suffix.lstrip(".")
+    reader = self.reader_map.get(file_ext)
+    if reader is None:
+      raise ValueError(f"Unknown file extension: {file_ext}")
+    with open(r, reader.open_mode) as f:
+      return reader.read(f)
```
rickwierenga: should this exist in service.file.py?

pylabrobot/config/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+
+from pylabrobot.config.config import Config
+from pylabrobot.config.service.file import MultiReader, FileWriter
+from pylabrobot.config.service.ini_file import IniReader, IniWriter
+
+DEFAULT_READER = MultiReader(
+  reader_map={
+    "ini": IniReader(),
+  }
+)
```
rickwierenga, line 20: is `FileReader` missing here?

pylabrobot/config/config.py
```diff
@@ -0,0 +1,53 @@
...
+
+@dataclass
+class Config:
+  """The configuration object for the application."""
+
+  @dataclass
+  class Logging:
+    """The logging configuration."""
+    level: int = logging.INFO
+    log_dir: Union[str, Path] = Path(".")
```
rickwierenga, line 31: would it make things easier to just convert to a Path before storing in Config?

pylabrobot/config/config.py
```diff
@@ -0,0 +1,53 @@
...
+  "CRITICAL": logging.CRITICAL,
+}
+
+LOG_TO_STRING = {
+  logging.DEBUG: "DEBUG",
+  logging.INFO: "INFO",
+  logging.WARNING: "WARNING",
+  logging.ERROR: "ERROR",
+  logging.CRITICAL: "CRITICAL",
+}
```
rickwierenga, line 20: perhaps `LOG_TO_STRING = {v: k for k, v in LOG_FROM_STRING.items() }`?

pylabrobot/config/config_test.py
```diff
@@ -0,0 +1,62 @@
...
+    (IniReader, IniWriter, "fake_config.ini"),
+  )
+  for rdr, wr, fp in cases:
+    run_file_reader_writer_test(
+      rdr, wr, tmp_dir / fp, fake_config
+    )
+
+
+def test_load_config_creates_default():
+  cfg = load_config("test_config", create_default=True,
```
rickwierenga, line 56: should we assert this test_config does not exist & unlink/warn if it does? It seems this test can easily fail.

pylabrobot/config/config_test.py
```diff
@@ -0,0 +1,62 @@
...
+  return tmp_path
+
+
+@pytest.fixture
+def fake_config(tmp_dir):
+  return Config(
+    logging=Config.Logging(
+      log_dir=tmp_dir / "logs",
+    )
+  )
```
rickwierenga, line 25: why are these fixtures?

pylabrobot/__init__.py
jt05610: This lets `pylabrobot.configure` be called after importing to update the project's logger. It comes into play with the testing fixture I put in for sequestering all test logs into test_logs, if I make it:

```python
@pytest.fixture(autouse=True)
def setup_test_config():
  logger = logging.getLogger("pylabrobot")
  print("initial logging handlers", logger.handlers[0])
  configure(TEST_CONFIG)
  print("testing logging handlers", logger.handlers[0])
  yield

```
I get

```
 initial logging handlers <FileHandler ~/PycharmProjects/pylabrobot/pylabrobot/config/pylabrobot-20240605.log (INFO)>
testing logging handlers <FileHandler ~/PycharmProjects/pylabrobot/pylabrobot/config/test_logs/pylabrobot-20240605.log (INFO)>
```

without this check, it would end up with both loggers being used.

pylabrobot/config/service/writer.py
jt05610: Probably not, I got into the habit from a book I read which uses both to be explicit https://github.com/cosmicpython/code/blob/master/src/allocation/adapters/repository.py

Happy to take the raise out if you prefer!

pylabrobot/__init__.py
rickwierenga: makes sense, thanks for explaining :)

pylabrobot/config/service/file.py
jt05610: I named this poorly -- it is intended for reading from files which could be multiple formats. Right now it is just used with the .ini reader, but can quickly be updated to support config with .json or .yaml.

I separated the file opening, then reading so that i can support from any text stream, i.e. for configuration via http instead of from a file. Could definitely simplify if preferred. I write a lot of Go, so am used to using io.Reader/io.Writer.

pylabrobot/config/__init__.py
jt05610: Changed the name of the MultiReader to be MultiFormatFileReader for clarity. Also changing it to DEFAULT_CONFIG_READER

pylabrobot/config/config.py
jt05610: Yes, good call. Taking the union out!

pylabrobot/config/service/writer.py
rickwierenga: is explicitness the only motivation? In that case I think just "saying it once" is clearer. Happy to discuss

pylabrobot/config/config.py
jt05610: Yes that's a better idea, changed

pylabrobot/config/config_test.py
jt05610: yes, fixed

pylabrobot/config/config_test.py
jt05610: No good reason I guess :) simplified

pylabrobot/config/service/file.py
rickwierenga: Then my question would be why `FileReader` is needed? It seems that the `IniReader` is able to read files by itself. The only exception is `test_file_reader_writer`, where the opened file is passed.

It would make sense to either:
1. allow ConfigReaders to read files, which would complicate reading from http/other sources
2. always have ConfigReaders read from an io stream, but then FileReader will not be a ConfigReader I think

2 is more powerful, but also more code.

pylabrobot/config/config.py
rickwierenga: need to push? :)

pylabrobot/conftest.py
rickwierenga, line 1: It seems that this file is not imported anywhere, does it run?

rickwierenga: is this still needed?

rickwierenga: I was a noob: https://docs.pytest.org/en/7.1.x/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files

rickwierenga:
Compared to your initial PR, I renamed `Loader` to `Reader` and vice versa. Same for `Writer` and `Saver`. I also split up the files into `config.io` and `config.formats` to make the distinction clearer.

It makes a lot more sense to me to have `read` and `write` be methods of the `io`, eg `FileReader` and `FileWriter` instead of `FileLoader` and `FileSaver`. `JsonReader` is now `JsonLoader`.

What do you think?

jt05610:
> Compared to your initial PR, I renamed `Loader` to `Reader` and vice versa. Same for `Writer` and `Saver`. I also split up the files into `config.io` and `config.formats` to make the distinction clearer.
> 
> It makes a lot more sense to me to have `read` and `write` be methods of the `io`, eg `FileReader` and `FileWriter` instead of `FileLoader` and `FileSaver`. `JsonReader` is now `JsonLoader`.
> 
> What do you think?

I like it! makes sense to me

rickwierenga:
anything before merging?

two quick questions:
- mind if I take some of your text from this PR for a quick tutorial on the docs site?
- i recently started plr propaganda on twitter. is it ok with you if I share a link to this PR? do you have an account i should @ / how should I give credit?

jt05610:
> anything before merging?
> 
> two quick questions:
> 
>     * mind if I take some of your text from this PR for a quick tutorial on the docs site?
> 
>     * i recently started plr propaganda on twitter. is it ok with you if I share a link to this PR? do you have an account i should @ / how should I give credit?

* Feel free to use any text!
* I made a twitter with @jt05610 as the handle!

rickwierenga:
Thank you for contributing this!

PR #160, opened 2024-06-14
Make PlateAdapter accessible on PlateCarrierSite

BioCam:
Hi everyone,

A tiny PR in which I have made changes in `pylabrobot/resources/carrier.py`...

1. enabled the assignment of a resource of type `PlateAdapter` as a child to `PlateCarrierSite` (specifically thinking about SLAS layout plate adapters like [Opentrons 96-well adapter](https://labware.opentrons.com/opentrons_96_well_aluminum_block?category=adapter)),
2. removed the `ShakerCarrier` subclass because there is only one shaker carrier and its only instance is implemented as a `MFXCarrier`,
3. added `ReagentCarrier` as a new subclass of `Carrier` as we discussed on [this automation forum thread](https://labautomation.io/t/standardized-labware-naming-convention-discussion/3787/19) that we keep all Hamilton carriers as separate, and
4. added the carrier name prefix to the docstrings to help people adhere to the naming convention.

### DIFF
--- a/pylabrobot/resources/carrier.py
+++ b/pylabrobot/resources/carrier.py
@@ -6,7 +6,7 @@ from typing import Generic, List, Optional, Type, TypeVar, Union
 from .coordinate import Coordinate
 from .plate import Plate
 from .resource import Resource
-
+from .plate_adapter import PlateAdapter
 
 logger = logging.getLogger("pylabrobot")
 
@@ -172,7 +172,9 @@ class Carrier(Resource, Generic[S]):
 
 
 class TipCarrier(Carrier):
-  """ Base class for tip carriers. """
+  """ Base class for tip carriers.
+  Name prefix: 'TIP_'
+  """
   def __init__(
     self,
     name: str,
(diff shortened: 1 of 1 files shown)


pylabrobot/resources/carrier.py
```diff
@@ -243,32 +250,35 @@ def __init__(
...
     model: Optional[str] = None):
     super().__init__(name, size_x, size_y, size_z,
       sites,category=category, model=model)
 
 
-class TubeCarrier(Carrier):
-  """ Base class for tube/sample carriers. """
+class ReagentCarrier(Carrier):
+  """ Base class for reagent/trough carriers.
+  Name prefix: 'RGT_'
```
rickwierenga, line 272: I thought this would be the Trough carrier. (all Carriers can have reagents)

BioCam: Yes, I completely agree. Should we still keep the "RGT_" name prefix for instances though?

rickwierenga: I am fully switched to abandoning this manure argument and switching to first-principle reasoning wherever possible. Can we use "TGH" or even just "Trough" or similar?

BioCam: Yes, I would like that too. Simply `Trough_` sounds the simplest to me.

I only copied the prefixes directly over because that was the last consensus on the [labautomation forum thread discussion](https://labautomation.io/t/standardized-labware-naming-convention-discussion/3787/14). Could you please message your proposal for an alternative there.

I am happy to change this to docstring to "Trough_" to finish this PR and update if things change.

rickwierenga: merging this now, we can update the doc string later. It will is inconsequential in this timeframe.

PR #161, opened 2024-06-14
Fix Lid Placement

BioCam:
Hi everyone,

In this PR I...
1. added a new attribute called `nesting_z_height` to the `class Lid(Resource)` in `pylabrobot/resources/plate.py`
2. updated `Lid` instantiation during the `Plate` instantiation to auto-adjust the `Lid`'s origin z-coordinate to its physical world position.

## Summary of the Issue
`Lid` objects are created during the instantiation of a `Plate(with_lid=True)`, and can be generated by themselves as well.
However, in the instantiation process the `Lid` is currently assigned to the anchor `Coordinate(0, 0, plate.get_size_z() - plate.lid_height)` in relationship to the plate parent, i.e. it currently assumes that the top of the plate and the top of the lid are the same horizontal surface.

This is physically incorrect, both in terms of how much overlap, aka "nesting", there is between the lid and the plate, as well as how much lid is actually protruding above the plate_top_surface.

This results in inaccuracies in various lid-related actions, incl.
- plate stacking, and
- plate grabbing (by the iSWAP or CO-RE grippers). 

## PR Content
As we have been discussing in this labautomation forum post, [PLR Lid Definition](https://labautomation.io/t/plr-lid-definition/3837), I initially created this PR to share my idea of a minimal change necessary implementation to fix these lid placement issues.

Specifically, I add one new attribute to the `class Lid(Resource)` called `nesting_z_height` which represents the true z-distance overlap between the lid and its parent plate.

Using this information alone the `Lid` is now assigned to the anchor `Coordinate(0, 0, plate.get_size_z() - plate.lid_nesting_z_height)` in relationship to the plate parent.
This means any resource placed on top of the combined plate-lid will now account for the lid.top_thickness + any lid.elevation if present without explicitly having to declare it.

I also integrated this information into LiquidHandler's `move_lid()` method but please check whether I missed any further necessary integrations.

I computationally tested the new implementation using my recent contribution of the `Cos_6_wellplate_16800ul_Fb` (and used this example to update its naming to the [new PLR Naming Convention for Plates](https://labautomation.io/t/standardized-labware-naming-convention-discussion/3787/14)).

## Next Steps
At the moment only the `Lid` of `Cos_6_wellplate_16800ul_Fb` has a defined `nesting_z_height`, all other `Lid`'s are currently defaulted to 0.

Warning: this might change current protocols!

Maybe we should say that `if nesting_z_height ===0 -> Coordinate(0, 0, plate.get_size_z() - plate.lid_height)`, i.e. use the old way until one updates `nesting_z_height`?

This is a first version to get the ball rolling and collaborate on potential alternatives :)


pylabrobot/resources/plate.py
```diff
@@ -81,12 +86,18 @@ def __init__(
       num_items_y=num_items_y, category=category, model=model)
     self.lid: Optional[Lid] = None
     self.lid_height = lid_height
+    self.lid_nesting_z_height = lid_nesting_z_height
     self.plate_type = plate_type
 
     if with_lid:
       assert lid_height > 0, "Lid height must be greater than 0 if with_lid == True."
+      if lid_nesting_z_height == 0:
+        print(f"{self.name}: Are you certain that the lid nests 0 mm with its parent plate?")
```
rickwierenga, line 95: do you think it's better to move this check to the initializer of Lid?

rickwierenga:
> This means any resource placed on top of the combined plate-lid will now account for the lid.top_thickness + any lid.elevation if present without explicitly having to declare it.

I generally think it's better to move these physical attributes directly than modeling derived quantities. That way, it's easier for the end-user to measure this for new resources (PLR does the computation) and also for PLR it's easier to add new features / fix existing ones when we now a specific physical model.

I understand this information is contained in derived quantities, normally, but then it's harder for users to contribute + it makes PLR a lot more complicated.

I think we should measure and store these in attributes:

- `Plate.base_height`

image

- `Lid.thickness`: 

image

Then:

```python
lid_nesting_z_height = lid.get_size_z() - lid.thickness
plate_nesting_z_height = plate.get_size_z() - plate.base_height
if lid_nesting_z_height > plate_nesting_z_height:
  lid_z = plate.base_height  # lid stands on base
else:
  lid_z = plate.get_size_z() - lid_nesting_z_height
```

---

A refactor:

I think it's better to not store `lid_height` and any other `lid_` attributes on `Plate`. This is the current implementation, but you adding another attribute highlights to me this is not scalable. We should scrap all of this from `Plate`, and just have it take a `lid: Optional[Lid]` parameter in the initializer. I realize now there is no reason for the `self.lid_height` in `Plate.assign_child_resource`, it should simply be `lid.get_size_z()` (where `lid` it the resources to be assigned).

To support this, we should add `Lid`-maker functions to `plates.py`, so that users can import a `Lid`-maker similar to how they import `Plate`-maker functions. For example, we should add `Cos_6_wellplate_16800ul_Fb_lid` to `corning_costar/plates.py`.

rickwierenga:
just read your post on the forum, and actually with your initial proposal it's easier for users to measure. Let's use that. I'll do this refactor.

rickwierenga:
It turns out: the venus labware database has lid definition files like `Cos_96_Filter_L.lid`:

```

COVER,default
CategoryCnt0CoverThickness1.5CoveredStackHt15.5Description8Cover for Corning plate with 96 filter wells (Landscape)Dim.Dx127.5Dim.Dy85.2Dim.Dz7.9RackBaseToCoverBase7.5ReadOnly1StackHt7.9ViewName'Cover for Corning filter 96 (Landscape)XExtent127YExtent86
* $$author=cjoerg$$valid=1$$time=2006-09-21 08:18$$checksum=a021bbd4$$length=084$$
```

Interesting:
- slightly different dim.dx/dy
- RackBaseToCoverBase=7.5
- not quite sure what StackHt=7.9 does, but perhaps it turns out to be useful in the future.

rickwierenga:
Since `Lid.RackBaseToCoverBase + Lid.DimDz != Plate.DimDz + Lid.CoverThickness`, I am not sure if this information is usable...

BioCam:
> just read your post on the forum, and actually with your initial proposal it's easier for users to measure. Let's use that. I'll do this refactor.

Yes, as discussed on the forum, I believe all required information should be easily measurable.

But there are two more components why I don't think this implementation can work:

> I think we should measure and store these in attributes:
> 
> * `Plate.base_height`
> 
> image
> * `Lid.thickness`:
> 
> image
> Then:
> 
> ```python
> lid_nesting_z_height = lid.get_size_z() - lid.thickness
> plate_nesting_z_height = plate.get_size_z() - plate.base_height
> if lid_nesting_z_height > plate_nesting_z_height:
>   lid_z = plate.base_height  # lid stands on base
> else:
>   lid_z = plate.get_size_z() - lid_nesting_z_height
> ```

1. **unrealistic scenario**: `if lid_nesting_z_height > plate_nesting_z_height` I have never seen a plate-lid relationship like this before.d
Because what this means is that the top of the lid does not lie on the top of the plate, but instead stands above it. From a design perspective this would mean one allows cross-contamination between the wells underneath the lid through a.) liquid vapour, and b.) if shaken, through spilling into one another. This would be against the reason of using a lid for the plate, and maybe I am not imaginative enough to see a use case, but I cannot see and have never seen such a lid-plate relationship.
3. **restrictive scenario**: that being said I have come across one situation in which this "unrealistic scenario" does appear but with a twist: if plate.base_height is a dependant consideration, I don't know how we could integrate plates that are _longer and wider (x-y dimensions)_ than the plate that the lid is placed upon: e.g. imagine a plate with a plate.base_height=5 mm and a plate.size_z=14. An "anti-evaporation lid" (i.e. heavy big lid) with dimensions bigger than SLAS and a nesting_z_height = 12, can physically be used. But this condition would not place that lid into the correct origin position.
4. **unclear definition for different plate geometries**: imagine you want to use anti-evaporation lid to reduce evaporation on e.g. a PCR preparation on [this standard Azenta 96-well plate](https://github.com/PyLabRobot/pylabrobot/tree/main/pylabrobot/resources/azenta).
It doesn't have a step-change on the edges that clearly marks what we defined above as plate.base_height. But a lid can still be placed upon it. All we need to know is what the lid's `nesting_z_height` is and then test different lids to see whether they lay flush on the top of the plate as desired. If yes, then that lids `nesting_z_height` auto-calculates that lid's correct origin.

azenta_4titude_96PCR_4ti-0960

BioCam:
> It turns out: the venus labware database has lid definition files like `Cos_96_Filter_L.lid`:

This is a fantastic find, @rickwierenga!

Trying to decipher what everything means, my guesses are:

```
CoverThickness=1.5 <=== what we called "lid.top_plane_thickness"
CoveredStackHt=15.5 <=== height of plate-lid to place another plate-lid onto
Description: Cover for Corning plate with 96 filter wells (Landscape)
Dim.Dx=127.5
Dim.Dy=85.2
Dim.Dz=7.9
RackBaseToCoverBase=7.5 <=== what we called "lid.nesting_z_height"
StackHt=7.9 <=== best guess: the height required to place one lid on top of another lid, could be different from Dim.Dz due to elevations on top of the lid?
```

`StackHt` could be the reason for what @jkhales saw in a stacking problem he described in the [corresponding labautomation forum thread](https://labautomation.io/t/plr-lid-definition/3837/6) (if I understood his problem correctly): currently plates are located too far down (in z-dimension), add the fact we don't _know_ the correct location to place one lid on top of another, a `lh.move()` command that places one lid on top of another will currently forcefully push them into one another.
I therefore agree, and this could be a useful piece of information for PLR too.

@rickwierenga do you have a distribution of dates for this small set of lid definitions?
And do we have catalogue numbers for them?
If we don't have catalogue numbers for the lids and/or their parent plate I would highly discourage adding them to the PLR Resource Library, because nobody can buy and use them.
And as a consequence, nobody can physically validate their definitions.

In general, I don't see the VENUS as "ground truths" at all because even in the Hamilton trainings one is being taught to use the given definitions as "starting points" and to make copies for making precise definitions for ones setup.
Therefore, I would argue that these VENUS-extracted definitions are not to be 100% trusted, and if we can't trust them and don't know what the exact labware ID number is, we should remove them from the PLR Resource Library rather than adding them.

That way, I hope, we can generate a Resource Library that is more maintainable, user-verified and peer-reviewed, and gives all the information needed to buy and use the resource even in another 18 years from now - or allow us to find on the internet whether this resource has been discontinued.

jkhales:
> `StackHt` could be the reason for what @jkhales saw in a stacking problem he described in the [corresponding labautomation forum thread](https://labautomation.io/t/plr-lid-definition/3837/6) 

It seems mostly like a manufacturing flaw:

https://github.com/PyLabRobot/pylabrobot/assets/1220444/ab677006-b365-4f15-b052-75c741921ddd

rickwierenga:
Points taken, `lid_nesting_z_height > plate_nesting_z_height` is unrealistic.

I also agree that we should not be using these VENUS resources because they are unusable due to unclear origin (no s/n) or likely incorrect information. You make good points trying to decipher this, but it seems a waste of resources to spend much time on this. I will respond to your questions if you want, but I'll skip it for now. Let's just do it from scratch ourselves.

Nesting height seems to be a good approach because it is easy to measure and allows us to do what we want.

Lid stacking (JJ's issue) seems to be a slightly different topic that should get its own PR.

I will fix the tests & clean up, then merge.

jkhales:
Kind of a separate issue, but related to lids, wondering if `get_size_z` should return different values based on whether the plate is lidded? or if there should be a different function that does this?

BioCam:
> Kind of a separate issue, but related to lids, wondering if `get_size_z` should return different values based on whether the plate is lidded? or if there should be a different function that does this?

That is a great question, and I have been thinking how the management of stacked plates should be handled in general.

I don't think that the `get_size_z` of the plate itself should change because it is not the plate that changes its size but the plate-lid complex that appears.

But I assume you are asking this question because it would be very cumbersome to always target a `lh.move()` command to the `Lid` of a `Plate` if it has a lid?

I was thinking about integrating a way around this in the assign command: If old_plate has lid assign new_plate to location old_plate_lid_top?

Thinking about the deck tree, this would mean that each plate from the bottom up has two direct child resources, its Lid and the plate on top of it.
Visually this would look something like this:
image

This would make the `move` commands simpler but the alternative, I highlighted on the bottom assignment tree, is arguably a lot easier to interpret?

rickwierenga:
ready to merge?

Long thread, but all that really changed is that the Plate initializer now takes a `Lid` resource (arbitrary) instead of in the past taking `with_lid` and automatically creating a lid. Errors are raised when definitions are incomplete.

rickwierenga:
playing the angry dictator role: let's continue discussing plate/lid tracking in a separate thread for clarity and searchability.

rickwierenga:
merged! Thanks for the work.

@BioCam, could you please share a picture of measuring the `nesting_z_height`? I want to put this on the website and twitter. Similar to https://docs.pylabrobot.org/resources/plate_carriers.html#pedestal-z-height

BioCam:
> @BioCam, could you please share a picture of measuring the `nesting_z_height`?

@rickwierenga, yes of course. Here is a private picture of a spare lid I had:

measuring_lid_guid

BioCam:
Note: 
For this specific plate lid I have to see whether a second plate-lid complex placed on top of the lid will be sitting directly on top of the lid, or on top of the lid.elevation.

In the above image we can see that:
- lid.size_z = 8.8 mm
- lid.nesting_z_height = 7 mm
-> lid.thickness_of_top + lid.elevation = 8.8 - 7 = 1.8 mm

Turning the plate back upside, I measured the lid.elevation = 1 mm -> lid.thickness_of_top = 0.8 mm.

I will measure on Monday whether another plate-complex placed on top of an existing one is now 0.8 mm or 1.8 mm higher up (in the z-dimension).

There is therefore potentially a point to be made that we should not use the measurement method shown on the left for evaluating lid.size_z because...
1. the lid.elevation might add to the overal lid.size_z but it might not contribute to another plate placed on top being higher up, and
2. for the purpose of moving the lid with an iSWAP/CORE gripper the lid.elevation is a distraction, when specifying `pickup_distance_from_top=1` it will grab the top edge of the lid because `pickup_distance_from_top=0` will try to grab the top of the elevation, which is inset and therefore not realistically reachable by the arm.

jkhales:
> > @BioCam, could you please share a picture of measuring the `nesting_z_height`?
> 
> @rickwierenga, yes of course. Here is a private picture of a spare lid I had:
> 
> measuring_lid_guid

nice ghible Susuwatari 😄

PR #162, opened 2024-06-17
Protocol Visualizer Overhaul: Making GIFs

joelaforet:
Hello! As part of a suite of educational tools we are building for a class at Duke, I extended the functionality of the visualizer to now allow rendering of protocols as .gifs. There are now a few buttons added to the visualizer.

### Recording a protocol:
To use the recording feature, first **use the slider to select how many frames you would like to save**. Each time a command is given to the visualizer, the update() function is called. This slider represents the number of unique update() calls you care about. I.E. **selecting 8 would mean that a snapshot of the deck is taken every 8 update calls**. NOTE: every movement of tips/liquid counts as an update. Filling a full tip rack of 96 tips means there are 96 update calls. 

After selecting your desired frame rate, **press the blue "Start Recording" button**. This turns on the recording code. Every N updates will be saved as images to be later exported as frames of a gif, or as a zip of png files. Run your PLR code like usual.

When you are done recording, **select the red "Stop Recording" button**. This will automatically render all current frames, and the gif will be shown in the right panel of your visualizer window. You can now decide to either **export this as a gif** by selecting the **"Download GIF" button on the left**, or export the frames individually by selecting the **"Download as ZIP" button on the right**. There is a box that allows you to enter in a name for your gif. If nothing is provided, it will default to "generated_gif.gif".

If you would like to record multiple sessions, simply select the blue "Start Recording" button again. This will clear all saved frames, and you will start from scratch based on the current state of the deck.

If anyone has any suggestions for new features/bugs they may come across please reach out! I would love to see your protocols animated! Hopefully this is useful! :)


rickwierenga:
very cool!

rickwierenga:
Wow, this is awesome! Worked perfectly. Very useful feature.

image

I'm curious if we can make this a bit less "big" visually. Around 70% of screen real estate is now taken up by the GIF feature. 

Some ideas:
- Perhaps we can put the "start" button in the navbar? The stop button only has to be shown (and "start" hidden) when there is a gif being recorded. The download button + filename field can be shown after the recording has stopped, along with the start button again.
- I am thinking perhaps the preview is hard to work in here, because it more than halves the space available for the visualization. I think for most people it's very fast to just view the downloaded file.
- The rendering progress can be temporarily shown after the recording has stopped.
- The zip feature did not work for me, but I think exporting as GIF should be enough tbh.

Again, I love this feature. I just think we have to make it balanced with the primary purpose of the visualizer, which is visualizing the deck.

joelaforet:
Also not sure how to fix the Unit Tests sometimes failing on random versions of the ubuntu. Any advice? @rickwierenga

pylabrobot/visualizer/lib.js
```diff
@@ -701,6 +719,47 @@ class Trough extends Container {
...
+
+    // Rename the children
+    for (let i = 0; i < this.num_items_x; i++) {
+      for (let j = 0; j < this.num_items_y; j++) {
+        const child = this.children[i * this.num_items_y + j];
+        child.name = `${this.name}_well_${i}_${j}`;
+      }
+    }
+  }
+}
```
rickwierenga, line 761: is the reservoir js class still needed? it seems to be a relic from this: https://github.com/PyLabRobot/pylabrobot/pull/149

joelaforet: Nope! That should be deleted.

rickwierenga: but the OT reservoirs are just `Plate`s in PLR

joelaforet: Yeah, I just had them super the methods from plate in the visualizer. I think keeping them as their own class is better because it allows for more utility features if, for some reason in the future, someone wants specific utility with reservoirs.

joelaforet: @rickwierenga Is it okay to have the visualizer super the `Plate` methods?

rickwierenga: the way visualizer classes are found is through `classForResourceType`, which looks at `resourceData.type`. In PLR, the type is never `Reservoir` because no class `Reservoir` exists in Python. I think we should add this back if we ever add a `Reservoir` class in python plr.

rickwierenga:
refactored the UI handling a tiny bit, and made the layout flatter as you saw. i think this is good to merge!

sorry for the delays

PR #163, opened 2024-06-18
[STAR] allow core hotel to be configured

jkhales:


rickwierenga:
what is the rationale for making this an offset as opposed to an absolute coordinate?

jkhales:
> what is the rationale for making this an offset as opposed to an absolute coordinate?

my machine didn't change, but the pickup started failing after we moved it to a different building. 

Thus is seems like it's a calibration, not a model, issue.

rickwierenga:
makes sense.

how is the user supposed to use this? accessing a private attribute seems to be the only way rn

PR #167, opened 2024-06-19
[STAR] fix get_core typo

jkhales:
🙃


rickwierenga:
Smh I should've caught this

PR #168, opened 2024-06-20
TroughCarrier Integration

BioCam:
Hi everyone,

In this little PR I am adding...
- the first instance of `TroughCarrier`, specifically `Trough_CAR_4R200_A00` (Hamilton cat. no.: 185436)
- the first Hamilton-manufactured trough, specifically `Hamilton_1_trough_200ml_Vb` (Hamilton cat. no.: 56695-02), which fits into the above `TroughCarrier`
- created volume & height calculation functions for this trough (due to the non-uniform geometry I empirically measured a set of liquid volumes heights using PLR's STAR CMM function, and chose a polynomial fit to create these functions)

## Next Steps
The functions achieve the following prediction accuracies:
```python
# Calculation accuracy data:
# input_volumes = [6_000,  10_000, 20_000, 50_000, 100_000, 150_000, 200_000,
# 240_000, 300_000] in ul
# measured_heights = [5.8, 7.4, 10.1, 18.5, 32.9, 47.8, 61.7,
# 72.6, 88.4] in mm
# calculated_heights = [6.032, 7.163, 10.002, 18.597, 33.073, 47.529, 61.722,
# 72.726, 88.356] in mm
```
People might want to generate something more precise.

`Trough`s were previously just a `Container`. However this does not account for the `true_dz` and `material_z_thickness` of the through, i.e. the machine thinks the bottom of the trough is further down in z than it actually is, and when sent to the "bottom" will crash into the trough material itself.

Interestingly, this `Trough_CAR_4R200_A00` actually has small oval-shaped pedestals that mean that the "trough_well"(?) directly sits on the pedestal, i.e. the inside of the trough is only off by the `material_z_thickness=1.5` mm.

I temporarily added these 1.5 mm to the `CarrierSites` of the `TroughCarrier` to compensate for this problem.
This is not the cleanest solution.
However, there only appear to be 2 troughs for this specific carrier and it seemed better to push this definition out into the open-source community than try to solve this niche problem instantly.

Happy automation 🦾


pylabrobot/resources/carrier.py
```diff
@@ -251,21 +251,6 @@ def __init__(
...
-    name: str,
-    size_x: float,
-    size_y: float,
-    size_z: float,
-    sites: Optional[List[CarrierSite]] = None,
-    category="shaker_carrier",
-    model: Optional[str] = None):
-    super().__init__(name, size_x, size_y, size_z,
-      sites,category=category, model=model)
-
```
rickwierenga, line 268: why did you remove this?

pylabrobot/resources/ml_star/troughs.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def _compute_volume_from_height_Hamilton_1_trough_200ml_Vb(h: float):
+  """ Function to compute volume of liquid in trough,
+  based on poylonmial fit of z-probed, empirical data.
+  """
+  if h > 89: # Maximal measured height possible
+    raise ValueError(f"Height {h} is too large for Hamilton_1_trough_200ml_Vb")
+  a, b, c, d = 5.50216554e-02, -5.30245236, 3.61064575e+03, -1.56155485e+04
+  polynomial_fit_of_empirical_data = a * h**3 + b * h**2 + c * h + d
```
rickwierenga, line 22: why do you use a 3d polynomial?

linear or maybe a parabola seems better visually, based on your data in the PR text:

image

pylabrobot/resources/trough.py
```diff
@@ -13,9 +21,18 @@ def __init__(
...
     size_z: float,
     max_volume: float,
+    material_z_thickness: float = 0,
+    true_dz: float = 0,
     category: Optional[str] = "trough",
-    model: Optional[str] = None
+    model: Optional[str] = None,
+    bottom_type: Union[TroughBottomType, str] = TroughBottomType.UNKNOWN,
+    compute_volume_from_height: Optional[Callable[[float], float]] = None,
+    compute_height_from_volume: Optional[Callable[[float], float]] = None,
```
rickwierenga, line 30: what do you think about moving these to `Container` after the PR?

pylabrobot/resources/trough.py
```diff
@@ -25,3 +42,29 @@ def __init__(
       category=category,
       model=model
     )
+    self.material_z_thickness = material_z_thickness
+    self.true_dz = true_dz
```
rickwierenga, line 46: what is `true_dz`?

pylabrobot/resources/carrier.py
BioCam: Because there are no `ShakerCarrier`s:
There is only one 7-track carrier that can flexibly take different modules onto it, including shakers, coolers, plate stack sites, NTR sites, etc.
I have already added this specific 7-track carrier called `PLT_CAR_L4_SHAKER` as a `MFXModule`.

As a result this `ShakerCarrier` class has never been used and, to my knowledge, has no opportunity to be used in the future.

I created this class before I understood that `PLT_CAR_L4_SHAKER` is better represented as a `MFXModule`.

pylabrobot/resources/ml_star/troughs.py
BioCam: I compared linear and polynomial fitting of the measurements to make this decision:

image

These functions need be really precise, so I opted for what yielded the best RMSE values. I can see a point to be made about the precision maybe being too dependent on my specific setup, no matter how thoroughly I went through the testing to ensure accuracy. But of course, a polynomial fit will always be more precise, and since the extra computational load is insignificant I didn't see a reason why to simplify it.

It is quite a deviation from the geometry-first approach I have pursued previously but I found these fits to simply work better.
(I actually believe now that this is the reason for the unexplained functions we see in VENUS as well - though I hope we can document this type of information enough for users to understand what is happening).

pylabrobot/resources/trough.py
BioCam: Love it, let's do that!

pylabrobot/resources/trough.py
BioCam: The distance between the bottom of a plate/trough and the lower side of the well_material/trough_well_material.

I didn't want to name it `dz` yet because we haven't fixed `dz` in `Plate` yet.

But this information is needed if someone wanted to use this trough anywhere that is not a `TroughCarrierSite` with a pedestal. It is the same situation as with `PlateCarrierSite`.

BioCam:
I just realised I forgot to include `volume=0 / height=0` in the polyfit.

Here is the updated accuracy data:

image

pylabrobot/resources/carrier.py
jkhales: That makes sense. With plr maturing and starting to be used by multiple different parties, I would encourage breaking changes be introduced more gently. For example with this one we could do something like:

```
class ShakerCarrier(PLT_CAR_L4_SHAKER ):
  """ Deprecated -  use PLT_CAR_L4_SHAKER  """
  def __init__(
    self,
    name: str,
    size_x: float,
    size_y: float,
    size_z: float,
    sites: Optional[List[CarrierSite]] = None,
    category="shaker_carrier",
    model: Optional[str] = None):
    # TODO #<number>: On August 31, 2024, make this raise an exception with the warn message.
    logger.warn("ShakerCarrier has been deprecated in favor of PLT_CAR_L4_SHAKER ")
    super().__init__(name, size_x, size_y, size_z,
      sites,category=category, model=model)
```
this would be backwards compatible and also alert people that they should update their projects.

rickwierenga: Agreed, we should communicate breaking changes in a gentler way. I will watch this in PRs. We should discuss versioning and a deprecation schedule soon-ish. I will set up that discussion.

@jkhales, to my knowledge, Retro is the only one using the ShakerCarrier class. Do you think you could update this now? then we can avoid having this warning and unused carrier.

jkhales: yea, np. But we also don't know who else is using this stuff, there could be some quiet person who just set it up in a small midwestern lab for whom it breaks and they don't know what to do and just give up on plr because it's so frustrating.

The cost of adding the message is very cheap. At the very least perhaps make it raise an Exception with directions on what to do? Could save someone hours.

pylabrobot/resources/ml_star/troughs.py
rickwierenga: Tbh, this looks like overfitting to me. Why would it not be linear? Does it reproduce like this? Happy to merge and I think going empirically can make sense for complex geometry, I'm just skeptical :)

pylabrobot/resources/trough.py
rickwierenga: `dz` in plate is just a parameter that will determine the .location.z of wells.

I think since you're building this from scratch now, we might as well do it right from the start, I don't see a reason to sync up with plate. I hope `true` is something we should not have to use, because that implies there is a `false` somewhere. Probably that's true right now, but not something we should bake into the library like that.

Perhaps `through_base_to_container_base` or something could work?

pylabrobot/resources/carrier.py
BioCam: I don't understand: How is this a breaking change, there is no instance of a `ShakerCarrier` in the PLR Resource Library?

I made this class in anticipation that I will add one but when I contributed `PLT_CAR_L4_SHAKER` I did so straight away as a `MFXCarrier`.
PLR can only see what is in PLR. If people make their own definition and don't contribute them, how would PLR ever know about it?

pylabrobot/resources/ml_star/troughs.py
BioCam: It is non-linear because the cross-section is a rectangle with one half-circle attached to one end that partitions into two sections along the z-axis: the bottom is cone-shaped but not fully, instead it has a flat surface of about 2-3 mm with (kinda a frustum of a pyramid style), while the top continues the cross-section in a semi-linear way.
It is semi-linear because in the middle of the trough there is a partition wall, which is open above and below. This wall is designed to avoid the liquid inside the trough to take up momentum when the trough is moved, i.e. reduce splashing, but as a consequence it takes up volume of the trough overal, adding to the non-linear liquid filling behaviour.

I am not sure there is such a thing as over-fitting with getting height and volume calculations of liquids in containers right :)

pylabrobot/resources/carrier.py
jkhales: > I don't understand: How is this a breaking change, there is no instance of a ShakerCarrier in the PLR Resource Library?

if someone is using it (me) in the code and it disappears, it breaks all of the code. This means it is "breaking". 

I'm unfamiliar with any sort of special repository that is "experimental" or  conversely "stable", i believe @rickwierenga shared this with me as something i can use. Is this the case?

BioCam: > if someone is using it (me) in the code and it disappears, it breaks all of the code. This means it is "breaking".

I think I understand now: I contributed `ShakerCarrier` in [PR#59 ](https://github.com/PyLabRobot/pylabrobot/pull/59/files) (2024-01-14), and then  contributed the only Hamilton shaker carrier `PLT_CAR_L4_SHAKER` in [PR#130](https://github.com/PyLabRobot/pylabrobot/pull/130/files) (2024-05-15).
At some point in the 4 months in between you created `PLT_CAR_L4_SHAKER` but as a `ShakerCarrier` (not a `MFXCarrier`) instance yourself and didn't contribute it to PLR?

I'm referring to "breaking" in the domain of PyLabRobot. Nobody can see into what people privately do with PLR.
But this cannot be included in the development of PLR decision making because nobody knows about it except for the person that does not contribute.

Of course, I have a lot of privately defined resources as well.
A number of PRs/commits "broke" these on several occasions with new developments of PLR:
e.g. the recent switch from `pylabrobot.resources.create_equally_spaced` to `pylabrobot.resources.utils.create_equally_spaced_2d` (commit [d16d995](https://github.com/PyLabRobot/pylabrobot/commit/d16d995b01ab6b6eda51b323271bdea5a11b953a?diff=unified&w=0)) "broke" all my private, non-contributed plate definitions.
But I would not expect the contributor to know about this, and stop or delay development of PLR because they didn't know about my non-contributed plate definitions.
The contributor, in this case @rickwierenga, did go through the necessary due diligence and changed all instances of `create_equally_spaced` to `create_equally_spaced_2d` **inside PLR**, i.e. there was no "break" of any functionality for users who use PLR contributed plates.

Similarly, I went through the PLR Resource Library and found no instance of `ShakerCarrier`.
Together with the open discussion of the developing [PLR resource naming convention](https://labautomation.io/t/standardized-labware-naming-convention-discussion/3787/5), in which nobody mentioned the need for an extra `ShakerCarrier` that can only define one carrier that has already been defined, I therefore removed my unused contribution to clean up the Resource Library.

pylabrobot/resources/trough.py
BioCam: > `dz` in plate is just a parameter that will determine the .location.z of wells.

As we discussed in [[PLR Carrier z-Height Issue](https://labautomation.io/t/plr-carrier-z-height-issue/3056)](https://labautomation.io/t/plr-carrier-z-height-issue/3056/35) this is an error, resulting from the incorrect VENUS-to-PLR  transfer of `ThicknessOfContainerBase` to `dz`, which in VENUS is actually the distance between bottom/base of plate_skirt to outside_bottom_of_well_base).
I am confident that VENUS is not using `plate_skirt_base_to_outside_bottom_of_well_base` because it is too long, and instead simply called it `dz` in direct analogy to `dx` and `dy`.

> I hope `true` is something we should not have to use, because that implies there is a `false` somewhere.

That is correct. The current `well.dz` is `false` as we know it is actually `well.thickness_of_container_base` / `well.well_z_thickness`.

But I see your point about immediately starting with the correct attribute names for troughs because we are starting from scratch with their definition and don't have to carefully adapt them like we currently have to do with `Plate` because `Plate` is used by every user so far.

pylabrobot/resources/carrier.py
jkhales: I'm using it in a script directly:

image

> e.g. the recent switch from pylabrobot.resources.create_equally_spaced to pylabrobot.resources.utils.create_equally_spaced_2d (commit [d16d995](https://github.com/PyLabRobot/pylabrobot/commit/d16d995b01ab6b6eda51b323271bdea5a11b953a?diff=unified&w=0)) "broke" all my private, non-contributed plate definitions.

yea, so the way I would suggest doing this since plr is maturing is to follow a more standard approach. One aggressive way would be to do:

```
async def create_equally_spaced(...):
    raise Exception("deprecate - use create_equally_spaced_2d instead.")
```

this would at least allow people to know how to fix things quickly.

A more standard way would be to do:

async def create_equally_spaced(...):
    logger.warning("deprecate - use create_equally_spaced_2d instead.")
    return await create_equally_spaced_2d(...)
```
and then give people a few months to update.

>  stop or delay development of PLR
this doesn't stop or delay! it's just documenting the changes in a user friendly manner.

> But I would not expect the contributor to know about this

If plr is going to become a *mature* library that is trusted and used by people, then it will have to start thinking about backwards compatibility.  It's never going to be taken seriously if breaking changes are pushed constantly without any communication or thought. I know plr is in a rapid development phase, but I'm trying to phase in slowly and get people thinking about how a professional OSS project is maintained.

pylabrobot/resources/trough.py
rickwierenga: So `dz` is actually only a parameter of `create_equally_spaced_2d`. As with `dx` and `dy`, it is not an attribute of ItemizedResource (or Plate). In that method call, the name is self-explanatory.

I think for the attribute here, it should really be named something descriptive like `through_base_to_container_base`.

pylabrobot/resources/carrier.py
BioCam: > yea, so the way I would suggest doing this since plr is maturing is to follow a more standard approach. One aggressive way would be to do:
> 
> ```
> async def create_equally_spaced(...):
>     raise Exception("deprecate - use create_equally_spaced_2d instead.")
> ```
> 
> this would at least allow people to know how to fix things quickly.
> 
> A more standard way would be to do:
> 
> ```
> async def create_equally_spaced(...):
>     logger.warning("deprecate - use create_equally_spaced_2d instead.")
>     return await create_equally_spaced_2d(...)
> ```
> 
> and then give people a few months to update.

Okay, this convinced me. Let's try and implement this with all major changes we make and see how we can avoid creating a list of timers of warning triggers that we then have to manually maintain with the handful of people that actively develop PLR.
But I can see and share your point about backward compatibility, even though we are talking about the development branch, i.e. PLR:main, and everyone should still check what changes they are pulling into their private production branches when syncing to the main branch anyway.
In the end, people who don't contribute their resource definitions will still end up having to manually sync their non-contributed resources to PLR's changes, they just get a couple of weeks to do it before breaking - if they chose to sync to the developer's branch latest updates.

I think we should have a small developer meeting in the next 1-2 weeks to discuss and write down how to tackle this, and answer questions pertaining to its implementation, e.g. how long to keep warning triggers before removing them (e.g. 2 months), and potentially how to automate their removal?

In this light, a warning note: last week Rick and I updated the lid definition. Rick went through all PLR-hosted plate definitions and updated them. But all of my privately-hosted plate definitions broke after I synced to the latest developer version on PLR:main and I had to manually update them.

pylabrobot/resources/trough.py
BioCam: > As with `dx` and `dy`, it [`dz`] is not an attribute of ItemizedResource (or Plate).

Yes, I don't think we should keep it like this. You have seen the computational load this created for `PlateAdapter`, even though in the real world these three attributes _are_ attributes of the physical plate. `Plate` should enable access to all the attributes that define a physical plate.
But I will raise this issue - if I find the time - again in a separate discussion.

Okay, then you are suggesting we handle trough completely differently to well-plate even though it is exactly the same. As long as it is consistent I don't mind.
I have now changed `Trough.true_dz` to `Trough.through_base_to_container_base` and the PR is ready to merge from my side

rickwierenga: > Okay, then you are suggesting we handle trough completely differently to well-plate even though it is exactly the same.

I don't think Trough and Well are exactly the same tbh. `Well`s have a z-location to the `Plate` (so there are two layers of resources), whereas with a `Trough` there is only one layer. The `dz` parameter is simply a shorthand for creating many resources (wells) that are at a certain z-height.

Because a Trough has no child (or parent), we have to define the `through_base_to_container_base` as a separate attribute. The alternative would be a "trough-parent and trough-child"-like structure, where the trough-child (the actual `Container`) is raised wrt trough-parent.

As you mentioned though, for wells, we should be taking into account the well bottom material thickness. I see that as _roughly_ analogous to `through_base_to_container_base`.

BioCam: > As you mentioned though, for wells, we should be taking into account the well bottom material thickness. I see that as _roughly_ analogous to `through_base_to_container_base`.

That is incorrect here: I've quickly made some rough drawings to visualise what the trough actually looks like:

Hamilton_trough_200ml

I highlighted the `Trough.through_base_to_container_base` with the actual measurement I took for it, 1.2 mm.

Since the "TroughCarrierSite".pedestal_z_height = 1.5 mm, the trough will therefore sit with its container.base directly on top of the "TroughCarrierSite".pedestal, while the "Trough.skirt" is hanging in the air leaving a 0.3 mm gap below it.

This is exactly like a plate with just one well.

rickwierenga: This is actually not different from what I thought 😅 What exactly is wrong here? Do you want to update this PR? I can revert so you can update, then I can remerge.

I realize my original phasing was not clear. This is not analogous in the physical geometry of these resources. I was just saying these are properties of "single-class" Resources and cannot be modeled as relations between resources. The well-dz is modeled as the relation between resources (namely in Well.location).

BioCam: > Because a Trough has no child (or parent), we have to define the through_base_to_container_base as a separate attribute. The alternative would be a "trough-parent and trough-child"-like structure, where the trough-child (the actual Container) is raised wrt trough-parent.

Yes, this is becoming confusing. Because a container is currently essentially a "paperbox", i.e. an object with no material_thickness associated with itself, i.e. all walls are believed to be 0mm wide, if I understand this correctly.

This is fine for wells because they sit inside the bigger plate - with the notable exception of their z-height, which is part of the big problem with `Plate` that we are trying to fix atm.
But in the case of troughs this does not really works as you can see above, because not only do we need to account for the material_thickness (at least in z) but here we also have to account for the "trough.child" / "container inside the trough" to be located at an offset to the left-front-bottom / origin of the trough.

BioCam: > Do you want to update this PR? I can revert so you can update, then I can remerge.

> I realize my original phasing was not clear. This is not analogous in the physical geometry of these resources.

No, I think we can leave this definition for now because I updated `Trough` to include
- `material_z_thickness=1.5 &`
- `through_base_to_container_base = 1.2,`

-> With this information people can aspirate/dispense with `offsets = [Coordiante(x, y, through_base_to_container_base + material_z_thickness]` which will then correctly target the bottom of this trough.

If we find this is cumbersome we can figure out a neater way in the future.

rickwierenga: I think in lh.aspirate or even STAR.aspirate & co we can incorporate this. Specifying offsets for every operation is not neat, that is something we can easily do automatically.

pylabrobot/resources/trough.py
rickwierenga: Done with https://github.com/PyLabRobot/pylabrobot/commit/f9d496cedc83fbd065e9d3c2b045e0038522feaa

rickwierenga: done with https://github.com/PyLabRobot/pylabrobot/commit/60cc8eddc359f7cb7457e40e520561dc8cac2a86

PR #169, opened 2024-06-24
remove “currently defined”

rickwierenga:
I removed "Currently defined" from the README labware docs, because no resource in the list is "currently undefined".

@BioCam, what do you think?

### DIFF
--- a/pylabrobot/resources/alpaqua/README.md
+++ b/pylabrobot/resources/alpaqua/README.md
@@ -1,4 +1,3 @@
-
 ## Resource defintions: Alpaqua Engineering, LLC
 
 Company page: [Alpaqua Engineering, LLC](https://www.alpaqua.com/about-us/)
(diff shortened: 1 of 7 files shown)


BioCam:
I think it depends on how we define "Currently defined".

I define it as "PLR-definition of the physical object + metadata needed to use it".
In that sense only the resources in the README docs are "currently defined", they are the only ones from the PLR Resource Library that people can use because they provide the necessary information for their use by providing at the very least the manufacturer catalogue number.

rickwierenga:
Agree with the definition.

Do you think we will ever document "undefined"/incomplete resources, or should we assume that all documented resources are 'currently defined'?

BioCam:
> Do you think we will ever document "undefined"/incomplete resources, or should we assume that all documented resources are 'currently defined'?

Good question 😅 

Yes, let's change the phrase as per your proposal...

...and start adding "To be deleted in 2 months." warning triggers to all resources which do not have a manufacturer catalogue number in their docstring, nor anywhere else in the PLR Resource Library?

rickwierenga:
Let's do that. It is unfortunate that venus resources appear unusable, but little we can do about that 🤷‍♂️

A future we can 'guarantee' good "PLR-definition of the physical object + metadata needed to use it" for all Resources is worth a lot more than a bunch of vague methods that no one uses.

Issue #170, opened 2024-06-24
Remove Resource's that don't have a catalogue number or other identifier (sept 1)


rickwierenga:
It appears many resources imported from the VENUS labware library have no catalogue number, and are therefore not usable or reliable. These will need to be removed in the future. We will need to rely on a combination of reliable third party databases, information from manufacturers, and crowd-sourced definitions.

As per @BioCam's [definition](https://github.com/PyLabRobot/pylabrobot/pull/169), going forward all resources will need "PLR-definition of the physical object + metadata needed to use it". Camillo has documented all resources that qualify in the associated README files in the `pylabrobot.resources` submodule, complete with identifiers and pictures.

I will be adding exceptions to resource definitions which I believe no one has verified. This will give people the chance to communicate that the definitions _are_ correct, or what changes are needed.

Resources for which we have not heard anything will be removed on September 1st.

BioCam:
Sorry, @rickwierenga, did you mean "Remove resources that **do not** have a catalogue number or other identifier"?

rickwierenga:
smh

jonlaurent:
Hey guys, I am starting to boot up some PLR stuff again and catching up on some big changes apparently. It seems like this particuarly change is going to remove quite a lot of plates in a few days, no? Like just about every Costar plate, for example.

I can help get some new plates in. I've previously been using some random Costar or other plate defs that happen to work for non-Costar plates, so pretty much everything is broken on this lol.

In particular: 384 PCR plates (from Globe, but ABI or Roche copies), Echo plates (from Labcyte or Beckman), Costar deep wells (which I think I can just change to Cos_384_wellplate_2mL... for now) and some plate reader plates I'll have to look at again.

jonlaurent:
Although looks like Costar stuff may be coming via this: [https://github.com/PyLabRobot/pylabrobot/issues/171](https://github.com/PyLabRobot/pylabrobot/issues/171) ??

rickwierenga:
Sure yeah please create any plate definitions for the stuff that you're using!

#171 is a handy data sheet, but not a soon-to-do for me (is generally good to verify defs, and I don't have many of the plates listed there). use it as a starting point for your labware :)

PR #172, opened 2024-06-24
Raise error undefined resources

rickwierenga:
Start of implementation for https://github.com/PyLabRobot/pylabrobot/issues/170.

@BioCam, Cos_96_EZWash is not conform to the plate naming specification. What do you think the name should be? it is this product: https://ecatalog.corning.com/life-sciences/b2b/US/en/Microplates/Assay-Microplates/96-Well-Microplates/Falcon®-96-well-Polystyrene-Microplates/p/353376

As per our discussion, I will keep the `Cos_96_EZWash`, but have it raise an error with instructions, instead of simply renaming the method.


BioCam:
> Cos_96_EZWash is not conform to the plate naming specification

Haha true, sounds more like some lab's use case of this plate from 20 years ago 😅 

Based on our [labautomation forum discussion](https://labautomation.io/t/standardized-labware-naming-convention-discussion/3787/2), this plate is
- manufacturer: Costar
- throughput: 96
- type: wellplate
- volume per container: 370 ul
- bottom: flat (Fb)

=> `Cos_96_wellplate_370ul_Fb`

Though asked like this I notice 2 things:
1. Costar is actually owned by Corning. I think this is why Opentrons places "corning_" at the beginning of their [labware names](https://labware.opentrons.com/?manufacturer=Corning) and not Costar.
2. I am starting to wonder whether this naming convention can actually deal with the myriad of plates that exist?
E.g. this plate is sold TC-treated and sterile. I am not sure whether the alternatives exist. If they do, then these alternatives should also be covered by the same PLR-definition.
But what about alternative models than have a different dx and dy but the same manufacturer, throughput, volume per container and bottom geometry?
Is this the reason why the Opentrons Labware Library has barely grown in recent years?

Sorry, I'm just raising more questions than I answer 😅

rickwierenga:
Thanks! Good to discuss implementing your naming algorithm in this validation-phase :)

> Costar is actually owned by Corning. I think this is why Opentrons places "corning_" at the beginning of their [labware names](https://labware.opentrons.com/?manufacturer=Corning) and not Costar.

`Cor`?

(actually, I now notice I linked a Falcon-resource, not a corning or costar one. Will fix the link before merging. I swear our lab's plates are corning ones but they look exactly the same.)

> But what about alternative models than have a different dx and dy but the same manufacturer, throughput, volume per container and bottom geometry?

That would be ... odd ... . I have seen many odd things, so perhaps we shouldn't assume it's not true!

PR #173, opened 2024-06-24
Make all Plates into Plate Stacks

BioCam:
Hi everyone,

In this PR I am adding the ability to place one plate directly onto another plate.

This generates "plate stacks", i.e. towers of plates in which each plate is placed on top of the other (this is different to "plate nests" in which there is some overlap between plates - this is not yet implemented).

As a result, every position that previously could hold 1 plate can now hold at least 2 and potentially up to 9 plates (depending on what the plate sits on and the geometry of the plate in question).

The primary goal of this PR is to facilitate accurate plate stacks with each plate having a lid.

## PR Content

- updated `liquid_handler.py`: `move_plate` to
    1. check whether the `to`_plate (destination of the move) has a lid -> if yes, move plate on top the lid
    2. if no, then move plate on top of the plate
    3. update assignments accordingly
- updated `resources/plate.py`: `plate.assign_child_resource`
    1. check whether the current plate (parent) has a lid -> if yes, actually perform a parent.lid.assign_child_resource <- "jump over the plate straight to its lid; this is to generate an accurate deck tree that represents the physical stack
- updating `Cos_6_wellplate_16800ul_Fb`'s lid information and using this plate-lid complex as an example for this new plate_stacking capability.
- small update of `Cos_96_DWP_2mL_Vb_L` to `Cos_96_wellplate_2mL_Vb_L` to convert to new naming convention (while leaving instructions for the change)

Furthermore, I added a print statement into the `move` command that returns the topmost absolute z_height that the plate_stack is at, and returns a simple string warning the user if the plate_stack goes above 245 mm, i.e. the safe traverse height.
This is to inform users of the danger they are creating but not limit them in case they programme the machine with awareness of this information, e.g. to create a big plate_stack in a corner to increase throughput.

## Next Steps
1. We really need to fix the broken `Plate` definition of PLR: When moving one plate on top of another located on a PlateCarrierSite with a pedestal, the bottom plate is believed to raise above pedestal by its `size_z`.
But this is not true, in reality the bottom plate is "sunk" around the pedestal by the "true_dz" (which is currently not accounted for), and the real bottom plate is therefore lower than PLR thinks it is. Moving a plate onto the bottom plate will therefore **drop** the plate "true_dz" mm above the bottom plate.
2. The adaptive nature `lh.summary()` breaks the plate_stacks. I have not yet been able to identify why. Please let us know here if you find the bug.

I hope this enables the throughput increase that I believe it can.

Please give your thoughts on the implementation of plate_2->plate_1 movement, even if plate_1 already has a lid.
I had some conversations with @rickwierenga and @jkhales but this PR is now dedicated to discussing and potentially changes this implementation.

```go
(7)   ├── mfx_carrier_1                     MFXCarrier     (235.000, 063.000, 100.000)
      │   ├── PlateStackSite_1              MFXModule      (235.000, 068.000, 118.195)
      │   │   ├── test_6_well_1             Plate          (238.200, 073.300, 123.495)
      │   │   │   ├── test_6_well_1_lid     Lid            (238.200, 073.300, 136.345)
      │   │   │   │   ├── test_6_well_2     Plate          (238.200, 073.300, 144.145)
      │   │   │   │   │   ├── test_6_well_2_lidLid            (238.200, 073.300, 156.995)
      │   ├── MFX_DWP_rackbased_module_1    MFXModule      (235.000, 164.000, 118.195)
      │   │   ├── test_6_well_3             Plate          (239.000, 167.500, 178.730)
      │   │   │   ├── test_6_well_3_lid     Lid            (239.000, 167.500, 191.580)
      │   ├── <empty>
      │   ├── <empty>
      │   ├── <empty>
      │
```


jkhales (commented):
Have you considered using a `ResourceStack`?

https://github.com/PyLabRobot/pylabrobot/blob/b7b39361666baf56165f8b1632810c0fe2aa4b96/pylabrobot/resources/resource_stack.py#L7

BioCam:
> Have you considered using a `ResourceStack`?
> 
> https://github.com/PyLabRobot/pylabrobot/blob/b7b39361666baf56165f8b1632810c0fe2aa4b96/pylabrobot/resources/resource_stack.py#L7

Yes, a `ResourceStack` has a couple of problems (for my applications):
- you need to create it as a separate, "abstract" resource and then assign to it, rather than to the resources inside it
- if you try to remove the top two plates at the same time it will not update the deck tree correctly
- correct plate placement with lid is not part of the `ResourceStack` yet (though that would be easy to fix now with the new `Lid` definitions.

jkhales:
> * if you try to remove the top two plates at the same time it will not update the deck tree correctly

This seems like a reasonable thing to change. @rickwierenga are you opposed to allowing users to pop stacks?

I'm curious in general if this approach is taken, what the point of the `ResourceStack` is?

rickwierenga:
@jkhales: What would popping stacks look like? eg `resource_stack.pop(2) -> resource_stack([resource_n-1, resource_n])` could work, while supporting moving a plate+lid, potentially multiple at once?

If neatly implemented, I think having this explicit in a ResourceStack is better. Can ResourceStack be made more intelligent so that locations of its child resources (the stack) is made correctly, based on the current top resource on the stack?

Something like `lh.move_plate(my_plate, my_stack)` is very explicit. It is slightly more complicated for representing lid -> 'plate on platestack' movemens: is that `lh.move_lid(my_lid, my_plate)` or `lh.move_lid(my_lid, my_stack)`?

```
(7)   ├── mfx_carrier_1                      MFXCarrier     (235.000, 063.000, 100.000)
      │   ├── PlateStackSite_1               MFXModule      (235.000, 068.000, 118.195)
      │   ├── PlateStackSite_1               MFXModule      (235.000, 068.000, 118.195)
      │   │   ├── Stack                      ResourceStack
      │   │   │   ├── test_6_well_1          Plate 
      │   │   │   │   ├── test_6_well_1_lid  Lid
      │   │   │   ├── test_6_well_2_lid      Plate 
      │   │   │   │   ├── test_6_well_2_lid  Lid
      │   ├── MFX_DWP_rackbased_module_1     MFXModule      (235.000, 164.000, 118.195)
      │   │   ├── test_6_well_3              Plate          (239.000, 167.500, 178.730)
      │   │   │   ├── test_6_well_3_lid      Lid            (239.000, 167.500, 191.580)
```

Having a plate as a child of another plate is not the way I intended the tree to work. This is not a definitive argument that means we can never change it, but I do think in this case we can have a less "changing" implementation. It is quite "harsh" to have Plate have different child-behavior than other Resources imo.

rickwierenga:
> The adaptive nature lh.summary() breaks the plate_stacks. I have not yet been able to identify why. Please let us know here if you find the bug.

Solved with https://github.com/PyLabRobot/pylabrobot/commit/30377677903bc6a60313651cabe5d182091428c7. The issue is that we sorted the tuples of `(resource name length, depth)`, meaning resources with longer names are "preferred" even if there are incredibly deep lists. Fixed by simply weighing the depth in that method and sorting by the "total" resulting length of the column.

BioCam:
> Having a plate as a child of another plate is not the way I intended the tree to work.

Why not? :)

Having to generate a `ResourceStack` every time one wants to move plates on top of another is quite cumbersome, and `ResourceStack` then represents a resource that does not actually have a specific physical meaning but a process management meaning.
I do think it is very useful for large plate processing for exactly that reason actually: because it can act as a type of iterator, i.e. one simply places things in the stack and takes things off the stack without having to worry about taking a plate in the middle out of the stack... except for when that is exactly one wants to do, in which case it would be very useful to move multiple plates together and have their location updated automatically because they are linked to one another by the direct physical connection they form (either plate-plate or plate-lid-plate).

> It is quite "harsh" to have Plate have different child-behavior than other Resources imo.
I don't understand. Don't specific subclasses of `Resource` exist to represent that specific subclasse's unique physical properties?
`MFXModule` handles its child resource differently from `Resource`, so does `PlateCarrierSite`, and so does `PlateAdapter`. Each of them are clearly designed to maximise the utility of the specific subclass, and to do so they all handle child-behaviour differently.

rickwierenga:
> Why not? :)

I generally view that the parent resource should "enclose" the children. This is not enforced in code, but that's the philosophy. This would make writing a good collision checker a lot easier.

> Having to generate a ResourceStack every time one wants to move plates on top of another is quite cumbersome, and ResourceStack then represents a resource that does not actually have a specific physical meaning but a process management meaning.

Yes, it's like an abstract resource.

Is it really cumbersome? (asking honestly, not rhetorically.) I would expect that for most workflows, you know you're gonna be using a stack, and this would be a minimal code change. The ResourceStack class can be responsible for implementing complex stacking behavior, popping, height checking, etc. That's a lot better than replicating this behavior across unexpected locations (I think Lid and Plate are quite unexpected).

> except for when that is exactly one wants to do, in which case it would be very useful to move multiple plates together and have their location updated automatically because they are linked to one another by the direct physical connection they form (either plate-plate or plate-lid-plate).

For this, would implementing a `ResourceStack.pop(n: int) -> ResourceStack`, returning a smaller resource stack, work? That, to me, seems like a more reasonable to do "stacked"/"bulk" updates than having a single Plate that is the parent for all.

> Don't specific subclasses of Resource exist to represent that specific subclass' unique physical properties?

Yes, that is 100% right. I wrote this under the thought that Plate significantly alters the tree from how it is used elsewhere in PLR (where parent -> child has some semantic difference, something changes cross layers, which is not the case when X is a child of another X.). That would be conflicting with the base class. Subclasses are supposed to complement this shared, base-behavior. This is a semantic/abstract difference that has no real implications in terms of code-your solution would work-but it is important to me philosophically.

Also: stacking seems to be a behavior one might expect in a class labeled Stack, more so than in a Plate class.

In more concrete terms: how would we share this behavior across lid/plate/petri dish/... stacks?

rickwierenga:
I am also thinking about tip stacks. It would be nice to have a method like `count_tips()`, which is harder when one TipRack is the parent of other TipRacks. What happens if the stack is empty?

BioCam:
> I generally view that the parent resource should "enclose" the children. This is not enforced in code, but that's the philosophy. This would make writing a good collision checker a lot easier.

Maybe I don't understand what implementation of collision checker you have in mind and I can see your point about "enclosing" but in reality tubes massively stick out of their tube carriers, plates out of their PlateCarrierSite and troughs out of their CarrierSite. As a consequence, there is no full "enclosing" in most situations and a collision checker would have to account for this anyway.



> Is it really cumbersome? (asking honestly, not rhetorically.)

Yes, I found it quite cumbersome: I didn't see it as a big problem on big machines but on small machines found it to be a big constraint:
Let's say you only have 8 SLAS positions, 3 are taken up by tip racks (maybe NTRs), now do maximise deck utility one has to stack and move plates dynamically around, particularly if one of the SLAS positions is a machine (let's say temperature control machine).
It the current situation, I have to pre-assign an imaginary `ResourceStack` onto every free SLAS position, and then start assigning plates to the `ResourceStack`.
This generates a lot more code, and, frankly, I found very confusing to navigate because I cannot see the `ResourceStack` since it is a thought construct for management purposes.
I've had a very similar problem with VENUS because VENUS also requires declaration of stacks, and I always wondered why I cannot simply place one plate on top of another, the exact same way I would handle large numbers of plates manually.

Again, I want to emphasise that I see utility of both approaches, for different applications.

> For this, would implementing a ResourceStack.pop(n: int) -> ResourceStack, returning a smaller resource stack, work? That, to me, seems like a more reasonable way to do "stacked"/"bulk" updates than having a single Plate that is the parent for all.

I am not sure what `ResourceStack.pop(n: int)` would actually do physically, would it send the command to move/take `n` resources from the top off directly?
Because if not then it is another management command that then requires the actual `move` command afterwards to execute the physical move. I don't think that is good practice because it adds again another coding step, and it deviates even more from the current PLR ethos of leaving management to users. And physically speaking there is one plate that is the parent to all plates above it. If that plate is moved all of its children will move with it, just like with carriers and mfx_modules.
I would also suggest not using the term `.pop()` here because it already has a clear definition in regards to Python containers (lists, dicts, sets), removal of an item at the specified index and returning that removed item.
What we are talking about sounds more like a "slice from the back" of the container?

> Also: stacking seems to be a behavior one might expect in a class labeled Stack, more so than in a Plate class.

If a human can do it and it is the most natural thing to do for them, I don't think we should cut a machine's ability to do the same?

> In more concrete terms: how would we share this behavior across lid/plate/petri dish/... stacks?

In the exact same way that we do it for the `ResourceStack`, just removing the constraint of being dependent on the declaration of this abstract `Resource` (the existence of which changes the meaning of `Resource` to now include a management tool with no physical meaning).

> I am also thinking about tip stacks. It would be nice to have a method like count_tips(), which is harder when one TipRack is the parent of other TipRacks. What happens if the stack is empty?

I agree, something similar to a `ResourceStack`, maybe called a `NestedTipRack` (in direct terminology upkeep of what they sold at) would be nice to track its tip count.
But we are going deeper and deeper into complex command generation and PLR pre-handling management of the liquid handling system, which I am a big fan of. In my opinion, this makes PLR more user friendly. 
But I have been repeatedly told that PLR provides "building blocks" through atomic commands that are easy to understand, management and complex command generation is left to the user and been helping people who do ask for complex commands/automated management because PLR has so far chosen not to provide these (e.g. [tip list iteration question on the forum](https://labautomation.io/t/automatic-incrementation-and-repetition-of-basic-liquid-handling-tasks/3869)).

So the question I would like to raise with this discussion is whether we are deviating from the atomic-command-only-approach, because automated plate location assignment and access in a `ResourceStack` or this idea of automated tip handling inside a `NestedTipRack` does represent a deviation from the previous ethos?
(i.e. users do not clearly declare what is accessed / how resources are managed)

To answer your question "What happens if the stack is empty?"
Because of the current ethos I had to generate tip iterators that have a "mark" in between different levels of NTRs. When the iterator gets to the "mark" it simply triggers a removal of the top NTR_tip_rack, and the tip iterator continues.

rickwierenga:
Just wanted to respond to some points before we call (will post results of call in a new comment below).

> As a consequence, there is no full "enclosing" in most situations and a collision checker would have to account for this anyway.

Correct, this is not strictly enforced currently. For that reason it is practically useless and inconsequential, but if we keep the philosophy of parents enclosing children, this will be easier to implement in the future. Currently, it's more of a philosophical point aimed at future improvements than something that is useful right now.

> This generates a lot more code, and, frankly, I found very confusing to navigate because I cannot see the ResourceStack since it is a thought construct for management purposes.

Until you added pedestal heights, I roughly thought of `CarrierSite` in this way.

I agree that `Resource` is a confusing name for a non-physical item on the deck. I imagine we could create a base class `AbstractLocation` defining `parent`, `children`, etc. attributes - all that is required to be a part of the deck tree. Then `Resource` could inherent from that. But frankly, I think this class would look exactly like `Resource`. Something to be discussed in more detail.

> I am not sure what ResourceStack.pop(n: int) would actually do physically, would it send the command to move/take n resources from the top off directly?

This is just a method for manipulating the deck tree, similar to `assign_child_resource` and `unassign_child_resource`. They don't move anything automatically. But, they are used by machines that  alter the deck tree such as `LiquidHandler`. In the future, when we have robotic arms, they will use these same methods to manipulate the tree.

>  I don't think that is good practice because it adds again another coding step, and it deviates even more from the current PLR ethos of leaving management to users.

I disagree. It provides a nice utility for something that is currently not neatly supported (taking the plate as a child of plate, which is currently doable, as "not neat"). I do agree with "PLR ethos of leaving management to users", and I think this is a tool for that. Both for use by end-users, as well as LiquidHandler, RoboticArm, etc. The resource tree is a model of lab-reality independent of machines used, and anything that changes it, should update the tree.

> I would also suggest not using the term .pop() here because it already has a clear definition in regards to Python containers (lists, dicts, sets), removal of an item at the specified index and returning that removed item.

It is supposed to be an equivalent to that :)

> If a human can do it and it is the most natural thing to do for them, I don't think we should cut a machine's ability to do the same?

Machines should definitely have the ability to do this with PLR natively. I was talking about where the code should go: I think a class named Stack is more natural than having this in plate. Plus, there is the argument about code duplication if we want to do this with Lid, Plate, etc.

> But I have been repeatedly told that PLR provides "building blocks" through atomic commands that are easy to understand. Management and complex command generation is left to the user.

That is correct, we do aim to provide building blocks. At the same time, we should also provide more complex and powerful tools as appropriate (the `lh.transfer` function is one example). What is left to the user are things specific to their setup.

I do not think we have a "atomic-command-only-approach". Instead, we do provide atomic commands but we should also provide some slightly more complex building blocks because it improves user experiences. For backends, they only provide atomic blocks because then the implementation work for new robots is minimal. Frontends like LiquidHandler specifically exist to provide nice features around those atomic commands on the backend.

rickwierenga:
important things discussed during the call:

Hotel vs stack vs nest terminology:
- hotel: vertical Carrier. Will be discussed separately.
- nest: a stack of resources that partially go into each other (eg pcr plates)
- stack: a stack of resources

We will treat nest and stack the same.

clarify that `pop` is popping from a stack, not "popping" a plate out of the hotel

pros of `ResourceStack`:
- easier from a SWE perspective to have a list in ResourceStack than a linked list (of `Resource`)
- sometimes you lose track of which plate another plate should be moved on top of. It's easier to refer to a stack
- can be empty
- if you have many nested resources, the tree grows very deep and long, hard to visualize in `lh.summary`
- less code duplication

pros of plate->plate:
- don't need to have code for defining ResourceStack
- `ResourceStack` is potentially confusing to use (tutorial can fix this)
- Can’t see a `ResourceStack` when sitting in front of robot / it's not a real resource.
  - We can introduce the concept of an `AbstractResource` for clarity/onboarding/teaching. Currently, our priority should be to get it working, otherwise we won't have something to teach.
 
**conclusion: use `ResourceStack`** (~simply counting number of pros and cons~)

We will keep `Lid` as a child of `Plate`.

todos:
- Rewrite plate stacking with ResourceStack. Think about nesting height.
  - need to make assignment on ResourceStack dependent on the current top resource. Use nesting height
  - need to think about what happens when ResourceStack is on a CarrierSite with pedestal. The first plate needs to be treated differently.
  - For aspiration and dispense, add a check that if a plate is on a resource stack, it is the top resource
- ~It would be nice to implement `get_top_item` for ResourceStack~: this exists: https://github.com/PyLabRobot/pylabrobot/blob/60cc8eddc359f7cb7457e40e520561dc8cac2a86/pylabrobot/resources/resource_stack.py#L104
- It would be nice to have a `Plate.get_resource_stack` or `get_stack(Plate)` method for getting the stack a `Plate` is in.

BioCam:
Thank you @rickwierenga.

I've been convinced that `ResourceStack` is the way to go forward and helps avoid plate position confusion when handled directly by the user.

I will add the new features of this PR into `ResourceStack` in a separate PR and close this PR.

BioCam:
Thank you @rickwierenga.

I've been convinced that ResourceStack is the way to go forward and helps avoid plate position confusion when handled directly by the user.

I will add the new features of this PR into ResourceStack in a separate PR and close this PR.

BioCam:
@rickwierenga 

Does `ResourceStack.get_size_z()` only add the sum of child_resource z dimensions of the first layer?

For example,

```python
(7)   ├── mfx_carrier_input                    MFXCarrier     (235.000, 063.000, 100.000)
      │   ├── input_PlateStackSite_0           MFXModule      (235.000, 068.000, 118.195)
      │   │   ├── input_stack_00               ResourceStack  (239.500, 071.000, 124.995)
      │   │   │   ├── Cos_6_plate_0            Plate          (239.500, 071.000, 124.995)
      │   │   │   │   ├── Cos_6_plate_0_lid    Lid            (239.500, 071.000, 138.395)
      │   │   │   ├── Cos6test0                Plate          (239.500, 071.000, 146.195)
      │   │   │   │   ├── Cos6test0_lid        Lid            (239.500, 071.000, 159.595)
      │   │   │   ├── Cos6test1                Plate          (239.500, 071.000, 166.195)
      │   │   │   │   ├── Cos6test1_lid        Lid            (239.500, 071.000, 179.595
```

Assuming that a single `Plate.get_size_z() == 20`, then  `ResourceStack.get_size_z() == 60` at the moment?

In other words, does  `ResourceStack` currently forget about the lids in between the plates and returns a z height value for the Stack that is lower that what it is in real life?

rickwierenga:
Yes. Should we fix this by updating `Plate.get_size_z` to include part of the `Lid` it has one?

BioCam:
Do you mean modify the actual size of the plate?

I don't think we should do that because it confuses the definitions of plate and lid together.

Tbh this is one of the reasons why I implemented the above PR as plate as a child of a lid as a child of a plate.... It represents the physical stacking.

We have to figure out a way to add the lid offset to both ResourceStack assign and liquid handler move to ResourceStack.

BioCam:
Done, this solves the problem:

```python
  def get_size_z(self) -> float:
    if not self.children:
        return 0
    if self.direction == "z":
      return sum(
          child.get_size_z() + (
              (child.lid.get_size_z() - child.lid.nesting_z_height) if child.lid else 0
          ) for child in self.children
      )
    else:
      raise NotImplementedError("Only self.direction == 'z' is currently supported")
```

I will add this together with the fixed `Plate` placement location into a `ResourceStack` in a new PR after thoroughly testing all the changes on a machine.

rickwierenga:
Great!

PR #174, opened 2024-06-25
TubeCarrier Integration

BioCam:
Hi everyone,

This is a small PR to add the last missing Hamilton carrier, `TubeCarrier` into the `resources/ml_start/` directory.

I exemplify its use by contributing...
1. `Tube_CAR_24_A00` (Hamilton name "SMP_CAR_24_A00"; Hamilton cat. no.: 173400)
2. `Falcon_tube_14mL_Rb` (Corning cat. no.: 352059)

This completes the standard Hamilton carrier integration with the following being integrated:
- `TipCarrier`
- `PlateCarrier`
- `MFXCarrier` #71 
- `TubeCarrier` (this PR)
- `TroughCarrier` #168 

## PR Content
`Tube_CAR_24_A00` is generated like any other carrier:

```python
    Tube_CAR_24_A00, Falcon_tube_14mL_Rb
)
tube_carrier_1 = Tube_CAR_24_A00(name="tube_carrier_1")

# Assign plate carriers to deck
lh.deck.assign_child_resource(tube_carrier_1, rails=5)
```

Tubes can then be assigned to each `CarrierSite` using e.g. a simple for loop:

```python
for y in range(24):
    tube_carrier_1[y] = Falcon_tube_14mL_Rb(name=f"Falcon_tube_14mL_Rb_{str(y).zfill(2)}")
```

This enables...
- assignment of diverse tubes in the same `TubeCarrier` with ease,
- assignment of tube adapters (not currently implemented in PLR) onto `CarrierSite`s: these adapters can e.g. take 1.5/2ml Eppendorf tubes.

## Next Steps

`Container` does not currently carry information about the material_z_thickness of the `Container`.
As a subclass of `Container` this also applies to `Tube`.

As a result the liquid handler believes that the bottom of a `Tube` is flush with the origin z-coordinate of the tube.

This means that the tip will crash into the bottom material of a `Tube` by definition.

I corrected this temporarily by giving the `CarrierSite` z-coordinate an additional 1.2 mm, i.e. the material_z_thickness of the `Tube` I tested, `Falcon_tube_14mL_Rb`.
Other tubes might have a thicker or thinner material_z_thickness. If thinner the dead volume will increase, if thicker the tip will crash again.

Keeping in mind the broken `Plate` definition we are currently trying to fix, I suggest we add `material_z_thickness` in a separate PR to `Container` (previously I thought it would only needed to be added to `Well` but this issue is persistent across various `Container`s including `Well`, `Trough` and now `Tube`).


rickwierenga:
Important to note, as we discussed, Camillo, that sites are labelled from front to back (similar to PlateCarrier and other PLR carriers) whereas the hardware labels are back to front!

BioCam:
> Important to note, as we discussed, Camillo, that sites are labelled from front to back (similar to PlateCarrier and other PLR carriers) whereas the hardware labels are back to front!

Ohh just, pushed my edit while you wrote this 😅

PR #175, opened 2024-06-26
add async

jkhales:
closes [#341](https://github.com/retrobiosciences/wetlab_workcell/issues/341)

### DIFF
--- a/pylabrobot/liquid_handling/resources/tilt_module.py
+++ b/pylabrobot/liquid_handling/resources/tilt_module.py
@@ -13,7 +13,6 @@ class TiltModule(Resource):
     name: str,
     backend: TiltModuleBackend,
   ):
-    # TODO: Measure size of tilt module.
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -13,7 +13,7 @@
 
 extras_fw = [
     'pyusb',
-    'serial'
+    'pyserial'
```
jkhales, line 16: i think this is the more standard one, also works on windows. imported as `serial` so a drop in replacement.

pylabrobot/liquid_handling/resources/tilt_module.py
```diff
@@ -13,7 +13,6 @@ def __init__(
     name: str,
     backend: TiltModuleBackend,
   ):
-    # TODO: Measure size of tilt module.
```
jkhales, line 16: I think this is specific to the tilt module you are using. e.g. i'm using hamilton, but others maybe using a different version.

jkhales:
I think the linting failure is from the `tilt` branch not being rebased onto main.

PR #180, opened 2024-06-27
Correct plate dimensions

rickwierenga:
Venus apparently defines plate footprints as 127.0 and 86.0, whereas the [ANSI SLAS 1-2004 (R2012): Footprint Dimensions](https://www.slas.org/SLAS/assets/File/public/standards/ANSI_SLAS_1-2004_FootprintDimensions.pdf) has them standardized to 127.76 and 85.48. This mistakes has unfortunately spread throughout PLR, but this PR fixes it.

I noticed Hamilton gripping is slightly inaccurate as a result of these inaccurate plate definitions. That is the case in our lab, on which I based these PLR commands. It might be better to grip plates slightly differently in the future (that means updating the unit tests) rather than applying an ad-hoc de-correctifier in STAR (essentially treating 127.76 as 127.0 again). As an example of this, looking from venus perspective, command `PP` parameter `go: Open gripper position [0.1mm]` was always 1300, which is 127mm+3mm. The 3mm is much more reasonable than 2.62mm. So I guess 1300 needs to be 1307 for ideal gripping. In similar spirit, I also updated some other parameters (like the center of the resource gripped with `xs` and `yj`).

### DIFF
--- a/docs/resources/plates.md
+++ b/docs/resources/plates.md
@@ -1,6 +1,8 @@
 # Plates
 
-Microplates are modelled by the {class}`~pylabrobot.resources.plate.Plate` class consist of equally spaced wells. Wells are children of the `Plate` and are modelled by the {class}`~pylabrobot.resources.well.Well` class. `Plate` is a subclass of {class}`~pylabrobot.resources.itemized_resource.ItemizedResource`, allowing convenient integer and string indexing.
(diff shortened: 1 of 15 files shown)


rickwierenga:
I only updated the "defined" resources ("PLR-definition of the physical object + metadata needed to use it".), because for "undefined" resources, we techncially don't know what they are, and so for all we know 127.0 and 86.0 are actually correct (unlikely though). We will scrub or fix those resources according to https://github.com/PyLabRobot/pylabrobot/issues/170.

PR #183, opened 2024-07-01
Integrate material_z_thickness (Container) & skirt_base_to_well_base (Plate) (fix PLR Plate location 2)

BioCam:
Hi everyone,

This PR's goal is to add two additional pieces of information to fix the current issue with `Plate` placements on flat surfaces, i.e. surfaces without a pedestal, in PLR:

- `resources/container.py`: `material_z_thickness` ... the thickness of the container (well/tube/trough/petri dish) in the z-dimension
- `resources/plate.py`: `skirt_base_to_well_base` ... the distance between the base of the plate to the bottom site of the well material 

This is step 2 of a series to fix the true location of plates in PLR, please see this forum discussion, [PLR Carrier z-Height Issue](https://labautomation.io/t/plr-carrier-z-height-issue/3056/32), for more information.

## Summary of the issue
Please see #143 for a summary of the bigger issue regarding plate placements.

All physical resources are made of a material that defines the boundaries of the resource.
This material has a thickness by itself.
However, in PLR we model physical resources as cuboids without taking into account the thickness of the material that makes up the edges of the cuboid.

An example to showcase the problem:
- `Plate` is a subclass of `ItemizedResource` which is a sublcass of `Resource`, and for the `Azenta4titudeFrameStar_96_wellplate_Vb` plate is represented as a cuboid of dimensions `size_x=127.76`, `size_y=85.48`, `size_z=16` (turquoise box)
- This plate's `Well`s are subclasses of `Container` and are represented as a cuboid of dimensions `size_x=5.5`, `size_y=5.5`, `size_z=15` (yellow box)

96_wellplate_model_CUBOIDS_fancy

This is what PLR "sees" as a plate with one well cuboid inside it.

While this model is sufficient to give accurate pipetting directions (`Well.get_absolute_location(x="center", y="center", z="center")` will return the geometric center of the bottom plane of the yellow / well cuboid), it is currently too simplistic, and missing the information about the thickness of the material that makes the yellow / well cuboid.

This becomes clearer when visualising the full plate that this cuboid model represents and giving a cross-section of row "H" to see the well:
96_wellplate_model_CROSS_SECTION_9

This visualisation showcases that the material thickness in the x and y dimensions are not absolutely necessary to model the well.
But the space below the well has to be modelled from the ground up to give an accurate location of the front-left-bottom corner of the yellow / well cuboid, also known as that yellow cuboid's / well's _origin_.
This means the `material_z_thickness` and the plate's `skirt_base_to_well_base` distance have to be known:

96_wellplate_model_CROSS_SECTION_8

Due to an error in the import of VENUS-defined plates `material_z_thickness` is curently defined as `dz`.
However, in analogy to 
- `dx` = distance between the yellow cuboid / well "H1" and the turquoise cuboid / plate in the x-dimension, and
- `dy` = distance between the yellow cuboid / well "H1" and the turquoise cuboid / plate in the y-dimension,
`dz` should be the distance between the yellow cuboid / well "H1" and the turquoise cuboid / plate in the z-dimension, not the thickness of the well material.

This is step 2 to correct the current `Plate` definition. In the next step I recommend we make the "old dz" obsolete by replacing all its use cases with `material_z_thickness` - what it currently represents. Using this careful step by step approach we adhere to the "do no harm" pragmatic paradigm and free up the term `dz` to take up the definition given above (orange arrow).

However, more than this, adding `material_z_thickness` to `Container`s will enable us to fix another PLR issue: the broken `Tube` and `Trough` definition:
Since these are modelled as cuboids with no material thickness, sending a tip to their `bottom` / origin's z-coordinate will crash the tip into the material. Having `material_z_thickness` in `Container` will allow us to adjust the `bottom` coordinate accordingly.

## PR Content

To integrate these two pieces of information I added...
- `material_z_thickness` as a required attribute to the definition of the `Container` class in `resources/container.py`:  all subclasses from `Container` will therefore carry this attribute; to not break any definition it is set to `0` at the moment, since all physical objects have a material thickness, a definition that is set to `0` is therefore equivalent to being undefined, enabling tracking of the progress of the definition scheme,
- `skirt_base_to_well_base` as a required attribute to the definition of the `Plate` class in `resources/plate.py`: this is something to be discussed:
normally PLR considers "anchors", i.e. helper information that defines a child resource's relative position in relationship to its parent resource as part of the child resource's information.
However, to fix `Plate` placements onto flat surfaces this information needs to be readily available from the `Plate` instance +  `skirt_base_to_well_base` does not give the information to the yellow cuboid's / well's / child resource's origin but instead gives the information to its lower material side - happy to discuss this aspect more in this PR :)


BioCam:
> And very informative visuals as always! (Can I put them on the docs website when this is merged?)

Thank you; yes of course, I created these visualisations for the open-source domain.


> I fully agree that we should model the container base thickness as well as having this in Container.

Awesome, then part 1 of this PR `material_z_thickness` is done.




> I would like to discuss one alternative. Looking at the third image, I think the yellow box (the Well size_{x,y,z} should include the material of the well. In other words, the Well dimensions should cover the entire well, including its material - as opposed to just its empty space/hole. Related, the blue arrow (`skirt_base_to_well_base`) should simply be the Well's z location (that is `dz`). Imo, `skirt_base_to_well_base` is not an attribute that should exist other than `well.location.z`.

Yes, I know what you mean, and I was wracking my brain on this one for a while before pushing a proposal with this PR for discussion:
As you say, it all depends on how we define "what is a well". Whether it is meant to include the material that holds it or whether it is just the empty cavity.
I am not sure.

The problem I see with measurements: if we say that the material is part of the well, then users need to measure the `Well.material_x_thickness` (`Well.material_y_thickness` is likely the same), and include it in the well dimensions. 
Problem 1: In that case all PLR well definitions are wrong at the moment and need to be revisited.
Problem 2: there might not be an easy way of measuring the `Well.material_x_thickness`: 
- one can use caliper,
- turn the plate upside down, and
- perform a jaw measurement on the total well (upside down this includes the material thickness on each side), 
- then turn the plate back upside up,
- perform an internal well dimension measurement, and
- calculate: `Well.material_x_thickness` = (outside_well - internal_well) / 2
But this seems a bit much maybe?

There is also the potential issue that wells might be too close to one another to measure the "outside_well" at all.

> To get the location of a well, you would need to do sth like `well.get_absolute_location(z="bottom") + Coordinate(z=well.material_thickness)`, but we can write a short-hand on Well like `get_pipetting_location` (with a better name, but you get the idea). For all LH operations, this will still be easily available using the `offsets` parameter. To me, just saying "pipette go to `well.get_absolute_location(z="bottom")`" actually should crash the pipette into the plate (current behavior) because I consider that the bottom z-location for a well.

Personally, this change to an "actually usable" method to move the tip to the "inside" bottom the well would have to happen at the exact same time as changing the yellow box / well to include the material thickness in all dimensions.
Easily switching between `well.get_absolute_location(z="bottom")` and `well.get_absolute_location(z="top")` has become of the most essential commands in protocol writing for me for defining resources accurately and quickly.
E.g.:
In the guide "How to measure material_z_thickness and skirt_base_to_well_base" (hope to make it in the next 2 weeks) I would briefly have to explain:
- place plate on PlateCarrierSite with a pedestal,
- move tip to `plate["H4"][0].get_absolute_location(z="bottom").z+5`
- reduce position incrementally, testing whether the plate can still be lifted
- no height at which it cannot
- remove plate
- use CMM cLLD to measure pedestal height at position
- calculate `material_z_thickness` = height at which tip touched well inside bottom - CMM_cLLD_measured pedestal position

Now we need to measure the `skirt_base_to_well_base` separately!
- place the same plate on a PlateCarrierSite without pedestal,
- repeat the above instructions to get `skirt_base_to_well_base` in the end

Calculate `dz = skirt_base_to_well_base + material_z_thickness`

-> This means these two distances need to be measured either way.


I believe I know what you mean though: having the material of the well as a "gap" between the inside of the well seems a bit odd.
But from what I can see it is the most practical, in terms of not having to change all PLR definitions, and maintaining only measuring the inside diameter of wells.
What do you think about these limitations of changing the current "well is only the empty cavity" to "well includes the material that makes up the cavity"?

> All but one resource that are not currently deprecated are maintained by you.
😮 

> Why should we not make this change in this PR? Then the PR is more 'self-contained'/complete. I can make this change for Cor_96_wellplate_360ul_Fb. We can also deprecate other resources until this information is obtained for them?

I believe the answer is "because it will break individual user's plate definitions which they haven't shared with the PLR open-source community" 😅 
They don't have `material_z_thickness` defined for their plates / tubes / troughs / petri dishes.
Implementing the change from `z=dz` to `z= material_z_thickness` will mean that the default `material_z_thickness=0.0` will be triggered and it will crash all their aspirations/dispensations into their material, even on PlateCarrierSites with a pedestal.

rickwierenga:
OK, I am convinced that the Wells should just be the holes. Look at exhibit A below for example. What is Well.material_x_thickness there? Hard to tell.

image

Still, for troughs it's slightly complicated. Do we measure the inside or outside diameter? I am fine with Well being a slight exception because they are not standalone objects and essentially just the shape of the Plate.

> I believe the answer is "because it will break individual user's plate definitions which they haven't shared with the PLR open-source community" 😅
They don't have material_z_thickness defined for their plates / tubes / troughs / petri dishes.

In line with what we are doing with `pedestal_size_z` (forced crowed-sourcing of information when people want to use plr:main), we should similarly follow that here. We update the resource model so we can have more accurate operations, we need more information, we ask people for that information.

BioCam:
> OK, I am convinced that the Wells should just be the holes. Look at exhibit A below for example. What is Well.material_x_thickness there? Hard to tell.
> 
> image

That is a great example of the arising complexity.

Then we agree to define `Well` as the empty cavity without its "container" / material that defines the boundaries of the `Well`.


> Still, for troughs it's slightly complicated. Do we measure the inside or outside diameter? I am fine with Well being a slight exception because they are not standalone objects and essentially just the shape of the Plate.

I agree this is a bit more complicated but your reasoning regarding `Well` being an exception makes sense:
`Troughs` are standalone objects, they can be stacked, they can even be moved by the iSWAP and thrown into the trash. That means PLR must know the outside dimensions.
`Tube` and `PetriDish`, the other `Container` subclasses`, are very similar.

I don't like that `Troughs`, `Tube` and `PetriDish` will be cuboid models in which `material_z_thickness` is inside the cuboid, while `Well` is a cuboid in which `material_z_thickness` is outside the cuboid... but `bottom` should in all cases be the inside bottom of the empty cavity, that information is incredibly useful on a day to day basis. How about we add `cavity_bottom` to access that position for all `Container` subclasses?
`bottom` is then the outside bottom, i.e. equivalent to the `Container`'s origin z-coordinate, and `cavity_bottom` = `bottom` + `material_z_thickness`?

If we do this then we could think about whether to include the material_z_thickness of `Well` into its definition.
This feels a bit like cherry-picking information because we just agreed that material_x_thickness & material_y_thickness will not be included into the same definition.
But this would mean we can implement `cavity_bottom` at the `Container` level, and the only difference of `Well` to other `Container` sublcasses is that it does not include the x/y material thickness?

In this situation the yellow cuboid does grow downwards to include the `material_z_thickness` and then `skirt_base_to_well_base == dz` as you proposed, making `skirt_base_to_well_base` unnecessary. (The measurement guide I proposed above will then just measure `dz` instead of `skirt_base_to_well_base`.


Summary of the discussion so far:
- part 1 of this PR `material_z_thickness` is done.
- part 2 of this PR `skirt_base_to_well_base`:
  -  We agree that `Well` cannot include the side material thickness(es) because of ambiguity in their measurements.
  - This is not a problem for self-contained `Container` subclasses like `Trough`, `Tube` and `PetriDish` because their material thicknesses (in all dimensions) can easily measured without ambiguity.
  - We have to identify whether it is okay to include `material_z_thickness` in the definition of `Well` and generate the exception for `Well` to not include `material_x_thickness` & `material_y_thickness`. If yes, then `skirt_base_to_well_base` becomes `dz`.

BioCam:
Thinking about "why does PLR's resource management system face this issue but Opentrons' and Hamilton's resource/labware management system does not (to our knowledge?"

I see the reason to be with the fact that PLR's resource management system treats `Well` like any other `Resource`, i.e. anchoring/referencing with regards to (wrt) the front-left-bottom (flb).
Other resource management systems all reference wrt center-center-bottom for wells. As a result, they don't have to factor in the side material thickness, but instead just know the `item_dx/y` spacing.

But PLR's resource management system is a lot more versatile and can cover a broader range of resources with the same definition system (to my understanding).

rickwierenga:
> But this would mean we can implement cavity_bottom at the Container level, and the only difference of Well to other Container sublcasses is that it does not include the x/y material thickness?

Yes, that is a good compromise. This way, `Well` stays closer to other `Container`s.

> Other resource management systems all reference wrt center-center-bottom (ccb) for wells. As a result, they don't have to factor in the side material thickness, but instead just know the item_dx/y spacing.
> But PLR's resource management system is a lot more versatile and can cover a broader range of resources with the same definition system (to my understanding).

Yes

BioCam:
Awesome, then I suggest to wrap this PR up with the following tasks:

- remove `skirt_base_to_well_base` (we will rename this soon to `dz`)
- add `cavity_bottom` to `resource.get_absolute_location(z="cavity_bottom")` if resource == `Well`
- update the `Well.size_z` for all maintained plates and their wells (I'd assume most people think this value is simply a depth measurement of the well, but now it will be a depth measurement of the well + the material_z_thickness.
- merge

Here is an updated infographic that shows that the yellow cuboid / well includes now the material_z_thickness but not the material_x_thickness & material_y_thickness:

96_wellplate_model_CROSS_SECTION_91

But all other `Container` subclasses do also include material_x_thickness & material_y_thickness into their definition:
e.g. `Trough`
96_wellplate_model_CROSS_SECTION_92

(we can come back to the discussion of whether there should be a `dz` for `Trough` in the above at a later time 😅 )

---

With these definitions we can write out a warning that the current `dz` will be changed to `material_z_thickness` in the next week.
This gives users time to update their non-shared resource definitions as discussed with @jkhales.

Once that switch has been executed, we can add the attribute `dz` back to `ItemizedResource` / `Plate` => because this is what we need to fix plate placement on flat surfaces :)

rickwierenga:
> add cavity_bottom to resource.get_absolute_location(z="cavity_bottom") if resource == Well

It is confusing if `Container` and subclasses allow different parameter values than other `Resource`s given that `get_absolute_location` is a `Resource`-level method. I think we should have `Container.get_cavity_location()` as its own method.

> (we can come back to the discussion of whether there should be a dz for Trough in the above at a later time 😅 )

Addressable through `Container.get_cavity_location()`? If no, let's discuss later.

> Once that switch has been executed, we can add the attribute dz back to ItemizedResource / Plate but with the definition shown in the infographic above => because this dz is what we need to fix plate placement on flat surfaces :)

it sounds reasonable to me to fix the `dz` simultaneously with the pedestal issue because then we cause no user-facing issues with plate assignments on carriers. We simply update dz to be correct, but also update the plate assignment on the carrier site to be accurate. Then absolute well locations will be the same after the update, just the relative location of well->plate and plate->PlateCarrierSite that change.

rickwierenga:
is it possible to not introduce `skirt_base_to_well_base` here because it's not used and will be removed in a week?

BioCam:
> It is confusing if `Container` and subclasses allow different parameter values than other `Resource`s given that `get_absolute_location` is a `Resource`-level method. I think we should have `Container.get_cavity_location()` as its own method.

That is true, it is a bit confusing tbh
As long as we have a method that returns the inside bottom of the well it should be okay. This is essential for testing labware definitions.



> it sounds reasonable to me to fix the `dz` simultaneously with the pedestal issue because then we cause no user-facing issues with plate assignments on carriers.

We _can_ do this, i.e. fix everything in this PR.
But then people will have to update their private `material_z_thickness` (for which they already have the correct value, stored in the current `dz`) and at the same time update their current `dz` value to the updated/corrected `dz` value.
I initially proposed to move in multiple steps to avoid confusion and make sure nothing breaks along the way.
Do you think we should update everything immediately?
It would be nice for me, I had to work around the broken `Plate` issue since I found it a couple of months ago, and would be eager to have it fixed an done.

BioCam:
> is it possible to not introduce `skirt_base_to_well_base` here because it's not used and will be removed in a week?

Ohh just saw this - **yes**, that is what I was thinking as well: it _is_ a bit of nomenclature shuffle 

This PR:
- introduce `material_z_thickness` (identify it's value by placing plates on a PlateCarrierSite **with** a pedestal, no other way to identify [except if you want to laser-cut the plate in half 👀 ])
- introduce `skirt_base_to_well_base` (identify it's value by placing plates on a PlateCarrierSite **without** a pedestal + knowing `material_z_thickness`, **or** use a caliper depth measurement)

...give people a couple of days to update these changes...

Next PR:
- repurpose `dz` to mean `skirt_base_to_well_base`
- delete `skirt_base_to_well_base`
- add conditional logic for correct plate placement based on PlateCarrierSite **with** _or_ **without** a pedestal (i.e. `pedestal_z_height` is `> dz` or `< dz`.

That would mean we don't have to change anything about this PR except for adding a warning that people have to update their plate definitions, with instructions on how to do so?

BioCam:
Finally found some time to make some visualisations to explain this issue with the currently broken `Plate`:

Based on the new `Plate` definition the two main `PlateCarrierSite`-`Plate` relationships are shown in this figure:

PlateCarrierSite_Plate_combos
**Figure A** - `PlateCarrierSite`-`Plate` _relationships to cover_

In comparison, PLR currently places the Plate on the `PlateCarrierSite` top, no matter whether there is a pedestal or not.
This means that PLR sees Plates on a pedestal like this:
96_wellplate_CURRENT_PLR_VIEW
**Figure B** - _current view of PLR's plate on a pedestal_

This means that currently PLR warps what plates look like in the real world.
It also means that PLR currently does not move a `Plate` on a `PlateCarrierSite` with a pedestal correctly: It moves it into the position shown above, the metal springs keep it there and the only way to move the plate into the position it has to be, is to use `STAR.core_check_resource_exists_at_location_center()` (#78) to push the plate onto the pedestal.

I hope this visualisations help us find the fastest way to move from the bottom **Figure B** to **Figure A** in PLR 😃

rickwierenga:
> Do you think we should update everything immediately?

yes. I just removed `skirt_base_to_well_base`.

I also made `material_z_thickness` an optional parameter, for now, because I saw the entire opentrons library lacks this attribute, and otherwise we would brake PLR for Opentrons entirely. This is sad because the attribute is not needed there. I did add a `@property` for `material_z_thickness` on `Container` that is not optional, and will raise an error when called if the private attribute `_material_z_thickness is None`. I suggest we use this attribute, causing an error to be raised only when the user performs an action that would require information that we don't currently have. If they don't do such operations, nothing changes.

rickwierenga:
what is left to do here?

BioCam:
I think this is good to merge.

With material_z_thickness now existing as an attribute we can ask users to fill this value as soon as possible.

And in the meantime we can work on "fix PLR Plate location 3", the final step of...
- correcting `dz`, and
- adding conditional logic to correct plate placement based on the `Plate.Well.dz` and `PlateCarrierSite.nesting_z_height` relationship :)

BioCam:
> > correcting dz
> 
> is that not already done?

No, we only _defined_ what the corrected `dz` should be like.

We have not made any of the necessary code changes to implement the corrected definition.
And without adding the conditional logic statements I don't think we can correct it because it would break all plate definitions.

rickwierenga:
I actually changed this definition already:

https://github.com/PyLabRobot/pylabrobot/pull/183/files#diff-088be936c625688bf1a0832f5d91875f8c54f1260d1cc9be3502457db3c661f2L73

Since the plate isn't sinking yet, previously the tip would go to the `dz` of 1.0 and now it is 2.5. `Cor_96_wellplate_360ul_Fb` already had the correct numbers.

But I agree, that is not a good change here. I can revert the plate definitions, and then we fix them in the next PR alongside the update to `PlateCarrierSite.assign_child_resource`. That means that for now `dz` and `material_z_thickness` will have the same value, hopefully only for a day.

Issue #184, opened 2024-07-01
remove deprecation check for single values in LH (Sep 1)


rickwierenga:

rickwierenga:
postponing removal to sep 1 because it's quite violent

PR #185, opened 2024-07-01
deprecate single values in LH

rickwierenga:
two changes in this PR:

- The `offset` attribute of `Pickup`, `Drop`, `Aspirate`, and `Dispense` dataclasses are now always wrt the center of its `resource`.
- Passing single values to LiquidHandler `pick_up_tips`, `drop_tips`, `aspirate`, and `dispense` methods. These methods now require a list of values. (tracking deprecation in https://github.com/PyLabRobot/pylabrobot/issues/184)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## Unreleased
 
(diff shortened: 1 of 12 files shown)


rickwierenga:
implementing https://labautomation.io/t/single-dispense-undefined-offset-issue/3074

PR #188, opened 2024-07-08
Add Shaker Carrier (a MFX Carrier) Information

BioCam:
Hi everyone,

a mini-PR to add some missing information into the `pylabrobot/resources/ml_star/README.md` regarding the `MFX_CAR_L4_SHAKER` aka the "shaker carrier".

Also updated the nomenclature of `PLT_CAR_L4_SHAKER` to `MFX_CAR_L4_SHAKER` because the shaker carrier can hold a lot more than just plates and is therefore a sub-type of multi-flex carriers.
This also appears to be the reason why the shaker carrier is included in Hamilton's Labware Assistant 3D software with the other multiflex carriers: 

image

However, when ordering the shaker carrier from Hamilton it is usually called `PLT_CAR_L4_SHAKER`.
This is why I originally contributed it to the PLR Resource Library under this name.
This appears to be a remnant from a time before MFX carriers were properly established, but in PLR we can rename the carrier appropriately to what it is.

No functionality has been broken because `PLT_CAR_L4_SHAKER` is redirecting to `MFX_CAR_L4_SHAKER` giving a warning about `PLT_CAR_L4_SHAKER` being deprecated and being removed soon.


rickwierenga:
thanks!

To be consistent with how deprecations are handled elsewhere, I think we should make this an Exception, not a warning.

rickwierenga:
does this work for you? if not, feel free to revert f25.

BioCam:
Yes, that is fine with me.

Actually better, because it forces users into action, and since it will be revealed in the import statement there is no danger to potential samples. :)

PR #192, opened 2024-07-12
Fix Plate Stacking in ResourceStack

BioCam:
Hi everyone,

In this PR I enabled accurate stacking of plates with a lid inside a `ResourceStack`.

This is to enable full access to a deck's z storage capabilities and increase run throughputs by multiples of its current (mostly) single-plate use.

## Summary of the Problem
`ResourceStack` is a "management" `Resource` class. That means a subclass of `Resource` which does not represent a particular physical object. Instead it is designed to facilitate handling of stacking / placing `Resource`s on top of each other, similar to how `list`s can store elements in Python. In analogy to Python's `list` `ResourceStack` is also mutable.

However, while adapting some of the changes I proposed in [PR#173](https://github.com/PyLabRobot/pylabrobot/pull/173) to enable accurate stacking of `Plate`s with lids, I encountered a couple of issues with the current implementation of `ResourceStack`:
- `ResourceStack` so far ignored whether a `Plate` has a `Lid`, and as a result placing more than 2-3 plates on top of one another raised a Z-drive error / was impossible; this was ignored on 3 separate implementation levels:
  - `ResourceStack.get_size_z()` ignored it, giving the wrong height of the `ResourceStack`
  - `ResourceStack.assign_child_resource()` ignored it, giving wrong coordinates for `Plate`s inside the `ResourceStack`,
  - `liquid_handler.move_plate()` ignored it, leading to incorrect physical movement of  `Plate`s onto a `ResourceStack, creating crashes;
- due to the current `Plate` placement issue, a `ResourceStack` has to be placed on top of a site (i.e. `PlateCarrierSite`, `MFXModule.child_location`) with a `pedestal_z_height = 0` (i.e. which does not have a pedestal) - this can only be fixed once we have fixed `Plate`

plate_stacking_render

## PR Content
In this PR the 3 implementation levels mentioned above:
  - `ResourceStack.get_size_z()`
  - `ResourceStack.assign_child_resource()`
  - `liquid_handler.move_plate()` 
...to work together and enable automated placement of plates (with lids) onto an empty and non-empty `ResourceStack`.

Specifically, all of these move an incoming plate if the topmost plate of the `ResourceStack` has a lid on top of that lid (but at the same time does not become a child of that lid, retaining the deck tree organization we discussed PR#173).

This offset is calculated to be 
```python
top_plate.lid.get_size_z() - top_plate.lid.nesting_z_height
```
lid_nesting_height_explainer

## Next Steps
This PR does not cover all possible stacking scenarios, both in terms of changes regarding plates and sites.

Importantly, we should clearly distinguish between the following plate placements on top of one another:
- plate "stacking" ... placing one plate directly on top of another, meaning there is physical contact between the two - however, in most instances we mean that the two plates have a lid in between them and **do not overlap** in the z-dimension, for that case we have...
  - plate "nesting" ... a subtype of stacking, meaning plates **do overlap** in the z-dimension; likely to be the case with e.g. PCR plates (this is not yet implemented)
- plate "hotel" ... plates might be placed on top of one another but do not touch one another; instead a plate hotel is more like a shelf or a vertical carrier, plates can be placed into specified locations and moved in and out of that position freely using an arm like the iSWAP (this is not yet implemented in PLR).

I think we should also add a conditional statement in the future which checks that the top plate in a `ResourceStack` (if it exists) is of the same model than the incoming plate. This is to prevent wrong placements.

I also think we should add a warning to users when their `ResourceStack` crosses the "minimal safety height" / "safe traverse height" of 145 mm above deck / absolute z-coordinate 245 mm.

### DIFF
--- a/pylabrobot/liquid_handling/liquid_handler.py
+++ b/pylabrobot/liquid_handling/liquid_handler.py
@@ -1766,25 +1766,26 @@ class LiquidHandler(Machine):
 
     if isinstance(to, ResourceStack):
       assert to.direction == "z", "Only ResourceStacks with direction 'z' are currently supported"
-      top_plate = to.get_top_item()
-      if top_plate.lid is not None:
-        to_location = top_plate.lid.get_absolute_location()
-        to_location = Coordinate(
-          x=to_location.x,
-          y=to_location.y,
-          z=to_location.z  + top_plate.lid.get_size_z())
+      if len(to.children) != 0:
+        print(F"First plate added to ResourceStack '{to.name}'")
+        # TODO: add pedestal handling here once Plate is fixed
+      
+      if len(self.children) != 0:
+        top_plate = to.get_top_item()
+        if top_plate.lid is not None:
+          to_location = top_plate.get_absolute_location()
+          to_location = Coordinate(
+            x=to_location.x,
+            y=to_location.y,
+            z=to_location.z + top_plate.get_size_z() - top_plate.lid.nesting_z_height + top_plate.lid.get_size_z())
+        else:
+          to_location = to.get_absolute_location()
+          to_location = Coordinate(
+            x=to_location.x,
+            y=to_location.y,
+            z=to_location.z  + to.get_size_z())
       else:
-        to_location = to.get_absolute_location()
-        to_location = Coordinate(
-          x=to_location.x,
-          y=to_location.y,
-          z=to_location.z  + to.get_size_z())
-        
-      # to_location = to.get_absolute_location()
-      # to_location = Coordinate(
-      #   x=to_location.x,
-      #   y=to_location.y,
-      #   z=to_location.z  + to.get_size_z())
+        to_location = to # if ResourceStack is empty
     elif isinstance(to, Coordinate):
       to_location = to
     elif isinstance(to, MFXModule):
--- a/pylabrobot/liquid_handling/liquid_handler.py
+++ b/pylabrobot/liquid_handling/liquid_handler.py
@@ -1774,7 +1774,7 @@ class LiquidHandler(Machine):
         if not isinstance(top_plate, Plate):
           raise ValueError("Cannot move a plate to a ResourceStack with a non-plate top item")
 
-        if top_plate.has_lid:
+        if top_plate.lid is not None:
           to_location = top_plate.get_absolute_location()
           top_plate_lid_offset = top_plate.lid.get_size_z() - top_plate.lid.nesting_z_height
           to_location = Coordinate(
@@ -1793,6 +1793,8 @@ class LiquidHandler(Machine):
       # Calculate location adjustment of Plate based on PlateAdapter geometry
       adjusted_plate_anchor = to.compute_plate_location(plate)
       to_location = to.get_absolute_location() + adjusted_plate_anchor
+    elif isinstance(to, Coordinate):
+      to_location = to
     else:
       to_location = to.get_absolute_location()
 
--- a/pylabrobot/liquid_handling/liquid_handler.py
+++ b/pylabrobot/liquid_handling/liquid_handler.py
@@ -1689,11 +1689,7 @@ class LiquidHandler(Machine):
         z=to_location.z  + to.get_size_z() - lid.nesting_z_height)
     elif isinstance(to, ResourceStack):
       assert to.direction == "z", "Only ResourceStacks with direction 'z' are currently supported"
-      to_location = to.get_absolute_location()
-      to_location = Coordinate(
-        x=to_location.x,
-        y=to_location.y,
-        z=to_location.z  + to.get_size_z())
+      to_location = to.get_absolute_location(z="top")
     elif isinstance(to, Coordinate):
       to_location = to
     else:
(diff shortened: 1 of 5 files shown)


rickwierenga:
Wow, this is really neat and tiny.

What I don't understand is why `LiquidHandler.move_plate` and `ResourceStack.assign_child_resource` both need to do the stack z-height correction, when this is already correctly implemented in `ResourceStack.get_size_z`. It seems to me that we only require one place for the z-top of the stack to be correctly computed, and in other places can just call this implementation when adding stuff on top of an existing stack.

In `LiquidHandler`, for example, we have

```python
if top_plate.has_lid:
  to_location = top_plate.get_absolute_location()
  top_plate_lid_offset = top_plate.lid.get_size_z() - top_plate.lid.nesting_z_height
  to_location = Coordinate(
    x=to_location.x,
    y=to_location.y,
    z=to_location.z + top_plate.get_size_z() + top_plate_lid_offset)
else:
  to_location = to.get_absolute_location()
  to_location = Coordinate(
    x=to_location.x,
    y=to_location.y,
    z=to_location.z + to.get_size_z())
```

But isn't the `z` coordinate here just the same as `get_absolute_location(z="top")`?

> plate "stacking" ... placing one plate directly on top of another, meaning there is physical contact between the two - however, in most instances we mean that the two plates have a lid in between them and do not overlap in the z-dimension, for that case we have...
> plate "nesting" ... a subtype of stacking, meaning plates do overlap in the z-dimension; likely to be the case with e.g. PCR plates (this is not yet implemented)

These can be implemented numerically instead of by binary yes/no overlap, I think. I have seen `plate.nesting_height` as an attribute in venus.

> I also think we should add a warning to users when their ResourceStack crosses the "minimal safety height" / "safe traverse height" of 145 mm above deck / absolute z-coordinate 245 mm.

As I think I mentioned on a call, writing it here for reference and future readers, the `DidAssignResourceCallback` on HamiltonDeck is a good place for that. Shall I add this?

BioCam:
> Wow, this is really neat and tiny.

Haha thank you 😅
I spent quite a bit of time cleaning up my previous implementation and transferring it to `ResourceStack`.
You might notice that he refactores `assign_chile_resource` uses presets, and conditional variable resets couples with early method return statements to make it more efficient.

> What I don't understand is why `LiquidHandler.move_plate` and `ResourceStack.assign_child_resource` both need to do the stack z-height correction, when this is already correctly implemented in `ResourceStack.get_size_z`. It seems to me that we only require one place for the z-top of the stack to be correctly computed, and in other places can just call this implementation when adding stuff on top of an existing stack.

Yes, there is a little bit of redundancy here.
This is mostly due to the order in which I discovered the bugs: liquid_handler.move_plate(), resource_stack.assign_child_resource(), resource_stack.get_size_z(),
And that I was time-pressurint myself to get this out before the weekend.

But it's important to note that...
-  requires relative / anchor coordinates while
- liquid_handler.move_plate() requires absolute coordinates.

This is because a resource_stack can be generated without being physically assigned to anything (a really useful feature actually).
But we can definitely simplify move_plate command now that get_size_z is fixed.


> But isn't the `z` coordinate here just the same as `get_absolute_location(z="top")`?

Yes, I think you are right :)

> > plate "stacking" ... placing one plate directly on top of another, meaning there is physical contact between the two - however, in most instances we mean that the two plates have a lid in between them and do not overlap in the z-dimension, for that case we have...
> > plate "nesting" ... a subtype of stacking, meaning plates do overlap in the z-dimension; likely to be the case with e.g. PCR plates (this is not yet implemented)
> 
> These can be implemented numerically instead of by binary yes/no overlap, I think. I have seen `plate.nesting_height` as an attribute in venus.

Yes, completely agree!
Plate doesn't have a nesting_z_height yet, and neither does tip_rack, but when I designed this attribute for lid I wanted it to transfer its meaning to these resources as well... Next couple of PRs?

> > I also think we should add a warning to users when their ResourceStack crosses the "minimal safety height" / "safe traverse height" of 145 mm above deck / absolute z-coordinate 245 mm.
> 
> As I think I mentioned on a call, writing it here for reference and future readers, the `DidAssignResourceCallback` on HamiltonDeck is a good place for that. Shall I add this?

Yes please! I remember the conversation we had but haven't had time yet to look into implementing it.
If you do, please feel free to add it and I'll learn from your PR :)

As we discussed, it shouldn't raise and error until the plate is actually impossible for the machine to physically grab it, that is at z coordinate 285 mm.
But between [245 - 284] it should just raise a warning (including in the visualiser for in silico testing first testing purposes - which I understand to be a pain but just saying it would be incredibly useful).

BioCam:
Doesn't self.direction != "z" mean that resources are stacked next to one another in x or y?

If so, there is an assertment that states this is not yet implemented.

Once implemented we would have to figure out how to define what the max height then is.
E.g. if there are 4 z stacks grouped together in a 2x2 grid ResourceStack, is the topmost z height across all 4 bundled stacks the topmost z height of the entire 2x2 ResourceStack?

rickwierenga:
> Yes, there is a little bit of redundancy here.

I think I removed the redundancy.

> Plate doesn't have a nesting_z_height yet, and neither does tip_rack, but when I designed this attribute for lid I wanted it to transfer its meaning to these resources as well... Next couple of PRs?

Please! On the one hand, I love these easy-to-measure quantities. On the other, they aren't really physical units. With plates on top of pedestals, the solution is a lot prettier imo. But it seems a bit harder to approach it in the same way for plate>lid and plate>plate and tip_rack>tip_rack stacking. `stacking_z_height` seems a fine compromise.

> Yes please! I remember the conversation we had but haven't had time yet to look into implementing it.

check this out: https://github.com/PyLabRobot/pylabrobot/pull/194

rickwierenga:
in the end all that really changed in this PR (to enable fancy {plate, lid} stacking) is we fixed `ResourceStack.get_size_z` to incorporate lid stacking. The rest are tests & using the new get_absolute_location.

rickwierenga:
good to merge?

BioCam:
> good to merge?

Just verified the behaviour of the refactored code on my machine. Everything working perfectly.

We're good to merge from my end.

I will make some more visualisations over the next couple of days and then write up a labautomation forum post for "PLR Feature Announcement: Plate Stacking in a ResourceStack" to summarise this work for a wider audience and we can then pick whatever we find most useful for the docs from there.

PR #195, opened 2024-07-14
full xyz rotations

rickwierenga:
support full rotations on the xy, xz and yz planes.

rotations, like locations, are with respect to the parent. introduce `get_absolute_rotation`.

origin is fixed wrt the resource, which is also the origin of rotation. (before we updated the resource's origin after the rotation to be the lfb always & we re-located the direct children. this was all silly)

implement two simple lin alg functions matrix_multiply_3x3, matrix_vector_multiply_3x3 to avoid numpy dependency.


rickwierenga:
Visualizer not supported yet. tracking in https://github.com/PyLabRobot/pylabrobot/issues/196

PR #198, opened 2024-07-18
Add pedestal_size_z to PLT_CAR_L5MD_A00

fderop:
Ideally I would add this measurement to all carriers in the file, but I don't have access to the other ones.


rickwierenga:
thanks!

PR #191, opened 2024-07-11
make units in STAR method parameters consistent

rickwierenga:
**problem** The issue is some backend_kwargs in STAR use 0.1 units whereas other parameters use 1-units. (eg 0.1mm vs 1mm). The reason for this, not the excuse, is that PLR-level parameters, that is on LiquidHandler, (eg flow_rates), use 'normal' units. STAR, providing an implementation for the firmware doc, uses their units. I initially designed it so that backend_kwargs, which sort of "fall through", would use the STAR-native units, but we have to come to realization that this is dumb.

**solution** This PR will standardize the PLR-level methods on STAR, those that make it a valid LiquidHandlerBackend implementation (`pick_up_tips`, `aspirate`, etc.), all use the default PLR units (mm, uL, mg, etc.). STAR-level methods like `aspirate_pip` will continue using Hamilton methods.

This PR will also replace all int conversions with `int(x)` to `round(x)` for increased accuracy.

PR #199, opened 2024-07-18
Fix Aspirate & Dispense Arguments

BioCam:
Hi everyone,

It appears that two arguments have slipped through making all liquid-handler commands take standardised units (i.e. mm and uL) in #191.

- `minimum_traverse_height_at_beginning_of_a_command` &
- `min_z_endpos`

for both `lh.aspirate()` and `lh.dispense()`

In this mini-PR I added a round(argument*10) statement if the argument is not `None`.

This rescued the correct behaviour of these arguments.


rickwierenga:
thank you for catching these!!

PR #200, opened 2024-07-19
[lawbare] add basic height functions and example

fderop:
Added some basic height/volume functions and example plate definitions that use lambda functions to call these height/volume functions. I think this is a better way of working than having a function per plate.


pylabrobot/resources/celltreat/plates.py
```diff
@@ -0,0 +1,100 @@
...
+            dy=5,
+            dz=3,
+            item_dx=38.5,
+            item_dy=38.5,
+            size_x=38.5,
+            size_y=38.5,
+            size_z=18.5,
+            bottom_type=WellBottomType.FLAT,
+            compute_volume_from_height=lambda liquid_height: _compute_volume_from_height_cylinder(
+                liquid_height, WELL_RADIUS
```
fderop, line 82: Here I use a lambda function call on the general volume function of a cylinder + some basic parameters to calculate volume

pylabrobot/resources/celltreat/plates.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+
+
+def CellTreat_6_WP_Flat(name: str, lid: Optional[Lid] = None) -> Plate:
+    WELL_RADIUS = 17.5
+
+    return Plate(
+        name=name,
+        size_x=127.0,
+        size_y=86.0,
```
rickwierenga, line 64: have you verified these are perfect? in https://github.com/PyLabRobot/pylabrobot/pull/180 we discovered the plate dimensions were wrong in VENUS (127.0x86.0) and I have spread this issue by copying it everywhere.

the difference is tiny, but it's good to have high standards for labware moving forward because we have the opportunity to build the best library in the world.

the official guide, which might be wrong, says 127.8 by 85.38 (https://www.celltreat.com/wp-content/uploads/6-Well-Plate.pdf)

we should obviously deviate from docs if reality measurements are different, but numbers have to be based on either docs or measurements, and not simply copy paste :)

pylabrobot/resources/height_functions.py
```diff
@@ -259,3 +259,62 @@ def calculate_liquid_height_container_1segment_round_fbottom(
...
+#         name=name,
+#         ...
+#         items=create_equally_spaced_2d(
+#             Well,
+#             ...
+#             compute_height_from_volume=lambda liquid_volume: _compute_height_from_volume_rectangle(
+#                 liquid_volume, WELL_LENGTH, WELL_WIDTH
+#             ),
+#         ),
+#     )
```
rickwierenga, line 281: love the usage example!

rickwierenga:
actually maybe we should consolidate into height_volume_functions.py 🤔

pylabrobot/resources/celltreat/plates.py
fderop: great find, I'll change this.

fderop:
> actually maybe we should consolidate into height_volume_functions.py 🤔

Makes sense, I will push

fderop:
I consolidated the two files, added depreciation warnings, and fixed the imports where vscode could detect them.

pylabrobot/resources/celltreat/plates.py
```diff
@@ -0,0 +1,114 @@
...
+    model=CellTreat_6_WP_Flat.__name__,
+    items=create_equally_spaced_2d(
+      Well,
+      num_items_x=3,
+      num_items_y=2,
+      dx=6.19, # from plate specs/drawing
+      dy=3.65, # from plate specs/drawing
+      dz=3, # manually calibrated
+      item_dx=39.04, # from plate specs/drawing
+      item_dy=39.04, # from plate specs/drawing
```
fderop, line 100: I updated these numbers from the diagram you found @rickwierenga - where did you find this btw?

rickwierenga: google :)

rickwierenga:
thanks @fderop!

if it's ok with you, i'd like to tweet a link to this pr from the plr twitter so people can see what's changing. what's your @ / how should I give credit?

fderop:
> thanks @fderop!
> 
> if it's ok with you, i'd like to tweet a link to this pr from the plr twitter so people can see what's changing. what's your @ / how should I give credit?

Oh wow, I'm flattered, you can use @fvderop

PR #207, opened 2024-07-31
enable dispense blowout on STAR backend

ben-ray:
I noticed this one line was triggering a `BlowOutVolumeError("No blowout volume was aspirated.")` error, and this seems to fix things


rickwierenga:
SMH thanks for fixing

but i think we need to move this to the end of the method, instead of removing it, because the blow out air is gone after a dispense.

PR #212, opened 2024-08-06
Make PlateAdapter z-offset Manually Adjustable + 1st Semi- & Non-skirted Plates

BioCam:
Hi everyone,

In this PR I have extended the capabilities of the `PlateAdapter` class to enable adjustment of the z-offset of plates at run-time, enabling use of different plates on the same `PlateAdapter` in the same run.

Furthermore, I added two plate definitions:

- `Thermo_AB_96_wellplate_300ul_Vb_EnduraPlate` ... PLR's 1st semi-skirted `Plate` (to my knowledge)
- `Eppendorf_96_wellplate_250ul_Vb` ... PLR's 1st non-skirted `Plate` and 1st non ANSI/SLAS 1-2004 conforming `Plate` :O

Finally, I added some additional information to the `README.md` for the `pylabrobot/resources/thermo_fisher/` folder, that details a possible way to handle PLR Resource Library contributions from Thermo Fisher Scientific (the umbrella company) and its various brands (e.g. Thermo Scientific, Fisher Scientific, Invitrogen, Applied Biosystems, ...).

# Summary of the Problem

At the moment `PlateAdapter` only adjusts the placement of a `Plate` onto itself in the x- and y-dimensions (i.e. aligning the `Plate`'s well-grid with the `PlateAdapter`'s hole-grid.

However, the `PlateAdapter.dz` value, i.e. where the hole-origin z-coordinate is, had to be manually set to a specific value based on the plate that has been used for testing.
This is not transferrable to other plates.

This is because different `Plate.Well`s have different x-z profile geometries (i.e. some are more cone than cylinder, some have different cone angles), while at the same time different `PlateAdapter` holes have different x-z profile geometries as well.
As a result, different plates will be able to be placed differently deep into the same `PlateAdapter`.

Since this difference is usually cone-shaped, and sometimes rounded, I deem creating a solution based on algorithmic condition checking is unlikely to be successful.

# PR content

Instead, I added the attribute...
-`PlateAdapter.adapter_hole_size_z` to represent the maximal depth of a hole,
-`PlateAdapter.dz` to represent the anchor z-coordinate of the hole and
 -`PlateAdapter.plate_z_offset: float = 0.0` to represent additional z-offset in relationship to `PlateAdapter.dz`

If `PlateAdapter.plate_z_offset = 0.0` the plate sits on the hole bottom.
If `PlateAdapter.plate_z_offset = 2.0` the plate sits 2 mm above the hole bottom (probably due to a plate-hole profile difference).

As an attribute of `PlateAdapter` `PlateAdapter.plate_z_offset` can be modified at runtime.
This means that one has to identify the correct `PlateAdapter.plate_z_offset` for each plate they are using (requiring manual testing but adjustable to all geometry relationships between adapter and plate).
Then one can use different plates on the same `PlateAdapter` in the same script 🚀 

Testing what the correct `PlateAdapter.plate_z_offset` is can a bit complicated due to the profile relationships and friction-fitting between adapter and plate, and I will make some visualizations in the future. 

---

Testing this new feature I added two new plates:
- `Thermo_AB_96_wellplate_300ul_Vb_EnduraPlate` 
...a semi-skirted plate, naming which required adjusting the PLR Resource Library naming convention to include "brands" (like Applied Biosystems, which is not a standalone company and therefore should never be placed as the manufacturer of a plate at the beginning of PLR plate function names)
- `Eppendorf_96_wellplate_250ul_Vb`
...a non-skirted plate, and to my shock a plate with x-y dimensions 123x81mm^2 - but it appears to be the plate in which NEB ships (all?) its chemically competent _E. coli_ cells ...
Testing this plate proofed difficult because `move_plate` always places a `Plate` into the origin of a `PlateCarrierSite`, which only works well with plates of x-y dimensions 127.76x85.48 mm^2 (aka ANSI/SLAS 1-2004 standard). But it can still be done.
I recommend designing and printing your own `PlateAdapter` for these plates though to make life easier.

I would also like to make a shoutout to Eppendorf here (something I have never done before):
Eppendorf provides absolutely amazing resources for their plates, including engineering diagrams, certifications and property sheets. This massively facilitates integration into automated workflows and makes purchasing from them super easy and fast... If I could I would source everything from them.
(Disclaimer: I am not affiliated with nor sponsored by Eppendorf and I'm quite certain they have no awareness of my existence 😅 )

# Next steps

Making the z-coordinate of a `Plate` placed on a `PlateAdapter` modifiable is not ideal because it does require testing of the correct `PlateAdapter.plate_z_offset` for each `PlateAdapter`- `Plate` relationship.

We should have a discussion on how and whether it makes sense to isolate certain relationships that can be automatically/algorithmically handled, while the implementation presented in this PR could serve as the "base behaviour".

Please let me know what you think :)


BioCam:
@fderop I thought about your problem, mentioned in #205, some more and believe this new feature (i.e. the ability to modify the z-offset of a `Plate` placed upon a `PlateAdapter` at runtime) can solve your problem.

fderop:
> @fderop I thought about your problem, mentioned in #205, some more and believe this new feature (i.e. the ability to modify the z-offset of a `Plate` placed upon a `PlateAdapter` at runtime) can solve your problem.

Looks like it can!! Thanks Camillo. One thing I saw is that you have volume functions for every plate, e.g. `_compute_height_from_volume_Thermo_AB_96_wellplate_300ul_Vb_EnduraPlate`, and the only addition there is the maximum volume/height warnings. I think it would be a bit more elegant if we had functions for the different geometries which throw an error/warning when max vol/height are exceeded, what do you think?

BioCam:
> the only addition there is the maximum volume/height warnings

That is not accurate. For every `Plate`'s specific volume/height function the specific volume/height warning AND the `_compute_height/volume` arguments change.
One might assume that the `_compute_height/volume` arguments are just straight up taken from the `Plate` definition but that is not true:
Instead I empirically measure the height of liquid at a range of volumes, and then run an optimiser to fit the `_compute_height/volume` functions to the real-world data.
This almost always changes the arguments from what the `Plate` definition gives.

The reason for that is that these functions oversimplify the geometry of a well, e.g. a 2-segment well with a cone at the bottom and a cylinder at the top actually ignores that the cylinder might not have linear slants but they might actually be round.

The alternatives are, to my knowledge, (a) what I recently did, to just use a polynomial fit to this data, or (b) use the "2nd segment" arguments in all the aspirate and dispense functions (part of their firmware), which is (I believe) what VENUS users are stuck with.
(b) sounds like an absolute programming nightmare, essentially requiring hardcoding of every move, while (a) is ultra-precise but loses readability.

Hence, why I have to modify the parameters to the `_compute_height/volume` to keep relative readability and get to a tolerable level of accuracy, no matter whether one aspirates/dispenses in the "1st" or "2nd segment" of a well (because PLR automatically calculates the liquid-level following needed anyway).

I have also started to add the results of empirical vs calculated data as dictionaries to these definitions, into the `plates.py` file next to the plate definition:

e.g.: `Eppendorf_96_wellplate_250ul_Vb_Lid`
```python
# results_testing_optimal_arguments_dict = {
#     "Volume (ul)": [0, 4, 8, 20, 70, 120, 170, 220, 260],
#     "Observed Height (mm)": [0, 0.45, 1.45, 2.55, 6.45, 9.05, 11.6, 13.15, 14.349],
#     "Predicted Height (mm)": [0.0, 2.91, 3.66, 4.97, 7.54, 9.06, 10.46, 11.86, 12.98],
#     "Relative Deviation (%)": [None, 545.56, 152.42, 94.80, 16.93, 0.056, 9.86, 9.82, 9.55]
# }
```

I agree though with you: ideally we would not need any of these functions at all, and have the correct heights and volumes be automatically calculated from the known geometry of the well.
But that is not (currently) feasible.
Partly, I'd say, because PLR just doesn't have enough information about the geometry of the well to allow this to happen.

I've got a couple of ideas of how to get around this but they'd all take some time :)

BioCam:
> However, it doesn't seem like the polynomial fitting solves these problems since the resulting error at low volumes is very high.

Apologies, I believe I didn't make my meaning clear enough: the above results...
```python
results_testing_optimal_arguments_dict = { # example fit for rational height/volume function
    "Volume (ul)": [0, 4, 8, 20, 70, 120, 170, 220, 260],
    "Observed Height (mm)": [0, 0.45, 1.45, 2.55, 6.45, 9.05, 11.6, 13.15, 14.349],
    "Predicted Height (mm)": [0.0, 2.91, 3.66, 4.97, 7.54, 9.06, 10.46, 11.86, 12.98],
    "Relative Deviation (%)": [None, 545.56, 152.42, 94.80, 16.93, 0.056, 9.86, 9.82, 9.55]
}
``` are **not** for a polynomial fit for the `Eppendorf_96_wellplate_250ul_Vb` plate.
These results are for a **fit of the rational** `_compute_height/volume` functions to empirical data.

I have not performed a polynomial fit for `Eppendorf_96_wellplate_250ul_Vb` nor any `Plate` yet.

But when performing a polynomial fit to `Hamilton_1_trough_200ml_Vb` the accuracy is very good:

```python
results_testing_optimal_arguments_dict = { # example polynomial fit
    "Volume (ul)": [0, 6000, 10000, 20000, 50000, 100000, 150000, 200000, 240000, 300000],
    "Observed Height (mm)": [0, 5.8, 7.4, 10.1, 18.5, 32.9, 47.8, 61.7, 72.6, 88.4],
    "Predicted Height (mm)": [2.7, 4.68, 5.99, 9.24, 18.69, 33.64, 47.84, 61.54, 72.33, 88.55],
    "Relative Deviation (%)": [None, 19.31, 19.05, 8.51, 1.03, 2.25, 0.08, 0.26, 0.37, 0.17]
}
```

I would also like to point out that I calculate relative deviations in % for analysis of prediction accuracies. That might not be the best way to assess the utility because for low volumes one will inevitably approach the dead_volume of the container anyway. Discarding a height/volume function because of high relative inaccuracies inside the dead_volume doesn't make sense. :)

> What do you think is the best way to measure the liquid levels?

My process:
1. Measure the `absolute_well_cavity_bottom` (if conductive use #69, if non-conductive one has to move channel+tip manually to z-position until plate cannot be moved up anymore).
2. Add increasing volumes of liquid to well (with a highly trusted manual pipette), use #69 to measure liquid level -> generates empirical data of "Volume (ul)"& "Observed Height (mm)".
3. Use data for fitting (either polynomial or rational).

This has been working very well in my hands, but yes, there is quite a bit of work involved in getting this done with high precision. 



> I added some examples [here](https://github.com/PyLabRobot/pylabrobot/blob/4c385cd834e43c6bf65492d450a68ecc3a583e15/pylabrobot/resources/celltreat/plates.py#L77). Maybe this approach can also be used to parametrize your polynomial?

Could you please elaborate on this approach? I am not sure I understand it.

How does this scale to wells that are not a simple geometry (cylinder, frustum of an inverted cone) but are bowed, multi-segmented geometries (potentially with asymmetries along both x-z and y-z planes, e.g. the Hamilton trough)?

pylabrobot/resources/thermo_fisher/README.md
```diff
@@ -1,18 +1,37 @@
...
+├── Ion Torrent (IT; brand)
+├── Gibco (GIB; brand)
+├── Thermo Scientific (TS; brand)
+│   ├── Nalgene
+│   ├── Nunc
+│   └── Pierce
+├── Unity Lab Services (brand, services)
+├── Patheon (brand, services)
+└── PPD (brand, services)
+```
```
rickwierenga, line 27: I think this might the first clear public explanation of what is going on here :)

BioCam: Hahaha thanks...
To be honest, all these mergers and acquisitions and what TFS then actually does with the companies it sucked up is very hard to see through, but I'm quite confident in this structure now.

I will similarly reorganise Corning resources... Similar behemoth to TFS

BioCam: I left out Life Technologies because it's not a brand anymore and is usually just found in old docs as a relic (please correct me if you find evidence to the contrary)

pylabrobot/resources/eppendorf/plates.py
```diff
@@ -0,0 +1,108 @@
...
+  return round(calculate_liquid_height_in_container_2segments_round_vbottom(
+    d=6.74,
+    h_cone=8.54,
+    h_cylinder=10.96,
+    liquid_volume=liquid_volume),3)
+
+# results_testing_optimal_arguments_dict = {
+#     "Volume (ul)": [0, 4, 8, 20, 70, 120, 170, 220, 260],
+#     "Observed Height (mm)": [0, 0.45, 1.45, 2.55, 6.45, 9.05, 11.6, 13.15, 14.349],
+#     "Predicted Height (mm)": [0.0, 2.91, 3.66, 4.97, 7.54, 9.06, 10.46, 11.86, 12.98],
```
rickwierenga, line 41: there is quite a big difference between predicted and observed height. do you think we should use your polynomial fit again?

pylabrobot/resources/thermo_fisher/plates.py
```diff
@@ -87,16 +113,111 @@ def ThermoScientific_96_wellplate_1200ul_Rd(name: str, with_lid: bool = False) -
...
+  return round(calculate_liquid_height_in_container_2segments_round_vbottom(
+    d=4.76,
+    h_cone=1.9,
+    h_cylinder=18.2,
+    liquid_volume=liquid_volume),3)
+
+# results_testing_optimal_arguments_dict = {
+#     "Volume (ul)": [0, 4, 8, 20, 70, 120, 170, 220, 260],
+#     "Observed Height (mm)": [0, 0.17, 0.77, 2.27, 6.57, 9.17, 11.17, 13.17, 15.17],
+#     "Predicted Height (mm)": [0.0, 2.16, 2.73, 3.70, 6.42, 9.13,11.84, 14.54, 16.71],
```
rickwierenga, line 158: same question as above

rickwierenga:
> I recommend designing and printing your own PlateAdapter for these plates though to make life easier.

Do you think it would be possible to have one possible definition + STL file in PLR available for everyone to use?

> That is not accurate. For every Plate's specific volume/height function the specific volume/height warning AND the _compute_height/volume arguments change.

Given a known geometry, we should have a parameterized function, of which we can fit the parameters (physical sizes) based on measurements.

BioCam:
> Do you think it would be possible to have one possible definition + STL file in PLR available for everyone to use?

Yes... I've designed one that covers probably 70-8-% of al use cases; the rest of 96-well plates I expect to have geometries that require specialised holders. Let me have a chat with our IP office about sharing our designs and come back to you one this topic :)



> Given a known geometry, we should have a parameterized function, of which we can fit the parameters (physical sizes) based on measurements.

That is what I have done above.
The problem is that we _model_ well geometries in a simplistic way: e.g. no PLR function goes beyond 2 segment divisions yet, and even beyond that a lot of geometries are curved or maybe even have asymmetric planes and rational modelling is therefore limited.

The question is just how precise do these functions have to be.

In the above example of a rational function fitted to measurements you, @rickwierenga and @fderop, appear to think that it is not accurate enough at the low end of the possible volumes?

rickwierenga:
Ah, so these parameterized functions are not the right shape because of too simplistic models or maybe manufacturing imperfections. That makes sense. To deal with this, perhaps it's good to use a fitted polynomial instead like you did for `Hamilton_1_trough_200ml_Vb`? You have already done the measurements ;)

Let it be explicit that the volume<>height functions are just arbitrary python float->float functions and you're not constrained to using the predefined functions in height_volume_functions.py.

The level of accuracy required obviously depends on use-case, but if it's easy, and it seems to be easy here because you have the measurements, I think higher precision is better.

BioCam:
ohh well...
rational_vs_polynomial_fit


...comes at the price of

```python
import numpy as np
import matplotlib.pyplot as plt

# Define the polynomial functions for predicting height
def polynomial_predict_volume(height):
    return max(0.9617 + 10.2590 * height - 1.3069 * height**2 + 0.26799 * height**3 - 0.01003 * height**4, 0)


def polynomial_predict_height(volume):
    return max(-0.1823 + 0.1327 * volume - 0.000637 * volume**2 + 1.6577e-6 * volume**3 - 1.1487e-9 * volume**4, 0)


# Provided dictionary
rational_predicted_heights = {
    "Volume (ul)": [0, 4, 8, 20, 70, 120, 170, 220, 260],
    "Observed Height (mm)": [0, 0.17, 0.77, 2.27, 6.57, 9.17, 11.17, 13.17, 15.17],
    "Predicted Height (mm)": [0.0, 2.16, 2.73, 3.70, 6.42, 9.13, 11.84, 14.54, 16.71],
    "Relative Deviation (%)": [0.0, 1172.65, 253.32, 63.00, 2.24, 0.41, 5.95, 10.43, 10.14]
}

# Generate the polynomial predicted heights
volume_ul = rational_predicted_heights["Volume (ul)"]
polynomial_predicted_height_mm = [polynomial_predict_height(v) for v in volume_ul]

# Calculate the relative deviations for the polynomial fit
observed_height_mm = rational_predicted_heights["Observed Height (mm)"]
relative_deviation_polynomial = [100 * (p - o) / o if o != 0 else 0 for p, o in zip(polynomial_predicted_height_mm, observed_height_mm)]

# Create the plot
plt.figure(figsize=(14, 8))

# Subplot 1: Observed vs Predicted Heights
plt.subplot(2, 1, 1)
plt.plot(volume_ul, polynomial_predicted_height_mm, 'ro-', label="Polynomial Predicted Height")
plt.plot(volume_ul, rational_predicted_heights["Predicted Height (mm)"], 'go-', label="Rational Predicted Height")
plt.plot(volume_ul, observed_height_mm, marker='x', linestyle='', color='purple', markersize=10, markeredgewidth=2, label="Observed Height")
plt.xlabel('Volume (ul)')
plt.ylabel('Height (mm)')
plt.title('Observed vs Predicted Heights')
plt.legend()
plt.grid(True, alpha=0.2)
plt.gca().spines['right'].set_visible(False)
plt.gca().spines['top'].set_visible(False)

# Subplot 2: Relative Deviations
plt.subplot(2, 1, 2)
plt.plot(volume_ul, relative_deviation_polynomial, 'ro-', label="Polynomial Relative Deviation")
plt.plot(volume_ul, rational_predicted_heights["Relative Deviation (%)"], 'go-', label="Rational Relative Deviation")
plt.xlabel('Volume (ul)')
plt.ylabel('Relative Deviation (%)')
plt.title('Relative Deviations')
plt.legend()
plt.grid(True, alpha=0.2)
plt.gca().spines['right'].set_visible(False)
plt.gca().spines['top'].set_visible(False)

plt.tight_layout()
plt.show()

```

rickwierenga:
you make the call! (maybe after Florian shares his view)

rickwierenga:
(i hope chat or copilot wrote that code ;))

BioCam:
> (i hope chat or copilot wrote that code ;))

hahaha

It was a collaborative effort (with time-consuming, exhaustive prompting to make it work as it should)

(and since this is all open-source work that is all fine :) )

pylabrobot/resources/eppendorf/plates.py
BioCam: Done

pylabrobot/resources/thermo_fisher/plates.py
BioCam: Done

BioCam:
Please let me know whether there are any more comments/feedback.

Otherwise good to merge from my side :)

rickwierenga:
Thank you!

BioCam:
Thank you @rickwierenga and @fderop for the peer-review and not being this bear...

image
(credit to the depths of reddit)

fderop:
> My process:
> 
> 1. Measure the `absolute_well_cavity_bottom` (if conductive use [Probe z height using channel #69](https://github.com/PyLabRobot/pylabrobot/pull/69), if non-conductive one has to move channel+tip manually to z-position until plate cannot be moved up anymore).
> 2. Add increasing volumes of liquid to well (with a highly trusted manual pipette), use [Probe z height using channel #69](https://github.com/PyLabRobot/pylabrobot/pull/69) to measure liquid level -> generates empirical data of "Volume (ul)"& "Observed Height (mm)".
> 3. Use data for fitting (either polynomial or rational).
> 
> This has been working very well in my hands, but yes, there is quite a bit of work involved in getting this done with high precision.
> 
> >

Initially I thought this wouldn't be super accurate, but the polynomial fitting seems to produce great results.  
  
>> I added some examples [here](https://github.com/PyLabRobot/pylabrobot/blob/4c385cd834e43c6bf65492d450a68ecc3a583e15/pylabrobot/resources/celltreat/plates.py#L77). Maybe this approach can also be used to parametrize your polynomial?
> 
> Could you please elaborate on this approach? I am not sure I understand it.

If the polynomials are of the same order/shape, you could parametrize them in the manner above. If the polynomial is not generalizable, then we can stick to a well-specific function.  

Either way, looks great!

pylabrobot/resources/plate_adapter.py
```diff
@@ -81,9 +83,11 @@ def __init__(
     self.dz = dz
     self.adapter_hole_size_x = adapter_hole_size_x
     self.adapter_hole_size_y = adapter_hole_size_y
+    self.adapter_hole_size_z = adapter_hole_size_z
     self.adapter_hole_size_z = self.get_size_z() - self.dz
```
ben-ray, line 87: > self.adapter_hole_size_z = adapter_hole_size_z
    self.adapter_hole_size_z = self.get_size_z() - self.dz
    
I am making the definition for a different Alpaqua magnet plate (the 96S) and I am trying to understand why adapter_hole_size_z is overwritten by self.get_size_z() - self.dz.
    
   Is the intended use defining all of these as best as I can, then use plate_z_offset to empirically test what height I need my pip to go?

BioCam: > adapter_hole_size_z is overwritten by self.get_size_z() - self.dz

I think this is a bug - I cannot say why this is the case, it doesn't change anything about the behaviour of the PlateAdapter.
The entire class variable `adapter_hole_size_z` is actually meant to enable more sophisticated behaviour in a future iteration (e.g. what if the hole_size_z > well_size -> plate "floats" in the wells... ).
The line `self.adapter_hole_size_z = self.get_size_z() - self.dz` should in my opinion be deleted.

BioCam: > Is the intended use defining all of these as best as I can, then use plate_z_offset to empirically test what height I need my pip to go?

Yes exactly.

For the x and y PlateAdapter positions this can be done very accurately, even with just the technical drawing.
I've made this and just need find some time to start the PlateAdapter docs pages to place it there:

example_alpaqua_magnet_dimension_calculations

For the z position this is a lot more difficult and can only be measured empirically.

This is because every plate-plate_adapter combo requires a different plate position due to how the plate's well's geometry interacts with the plate_adapter's hole.

But there _is_ an absolute `Plate_adapter.dz` which is the bottom of the plate adapter hole.
To measure this you can move a channel with a tip or teaching needle above the hole, marked here by the yellow arrow, and perform a z-probing on that position:
250927_Moschner_Alpaqua_MAGNUM_FLX_render

_This_ is then the default position that a plate would try to be moved to (if `PlateAdapter.plate_z_offset==0) and gives a _starting point_.

During teaching PLR knows where the plate should be -> z_value_0

Then you place the plate onto the plateadapter and probe the real position -> z_value_1

Then -> `PlateAdapter.plate_z_offset = z_value_1 - z_value_0`

This process is automatable (i.e. can dynamically be performed at runtime) :)

PR #213, opened 2024-08-07
[resource] fix to deep copy on rotation

jkhales:
without this, the functions, e.g. `_compute_height_from_volume` aren't preserved.

Perhaps there is a different way you want to approach this?

### DIFF
--- a/pylabrobot/resources/resource.py
+++ b/pylabrobot/resources/resource.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import copy
 import itertools
 import json
 import logging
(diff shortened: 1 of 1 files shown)


rickwierenga:
I think this also copies .parent, and the whole tree with it. That is undesired imo. We should copy down the tree, but not up. One bad way to fix this is to simply set `parent = None` after the update.

But the bigger issue here is these functions do not get serialized (the current copy function is based on serialization), which is also bad when saving to a file/etc. I will close this PR and write a new one where we (de)serialize functions.

PR #216, opened 2024-08-13
[Resource] Add slim 300 uL tips

fderop:
Hamilton sells [300 uL CORE-II tips with a slim diameter](https://www.hamiltoncompany.com/automated-liquid-handling/disposable-tips/300-%C2%B5l-slim-conductive-filter-tips). `STF_L` failed to pick these tips up, I assume because of the collar type. `HTF_L` worked but does not handle maximum volume properly. I added a definition for the slim 300 uL tips: `slim_standard_volume_tip_with_filter`. Let us know if there's a better way to do this.  
  
I also transferred the comments to the rack creator docstrings, feel free to leave that commit out.

### DIFF
--- a/pylabrobot/resources/ml_star/tip_creators.py
+++ b/pylabrobot/resources/ml_star/tip_creators.py
@@ -119,7 +119,7 @@ def standard_volume_tip_no_filter() -> HamiltonTip:
   )
 
 def standard_volume_tip_with_filter() -> HamiltonTip:
-  """ Low volume tip without a filter (`tt01` in venus) """
+  """ Standard volume tip without a filter (`tt01` in venus) """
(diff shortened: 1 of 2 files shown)


pylabrobot/resources/ml_star/tip_creators.py
```diff
@@ -128,8 +128,18 @@ def standard_volume_tip_with_filter() -> HamiltonTip:
     pickup_method=TipPickupMethod.OUT_OF_RACK
   )
 
+def slim_standard_volume_tip_with_filter() -> HamiltonTip:
```
fderop, line 131: Here is the slim tip

rickwierenga:
thanks!

PR #219, opened 2024-08-15
- added setup method that calls LiquidHandlerBackend and USBBackend

Robert-Keyser-Calico:
I think that the changes to the Hamilton backend caused the setup cascade to break for the EVO backend.


rickwierenga:
Thanks!

PR #220, opened 2024-08-16
[serializer] handle closure

jkhales:


jkhales:
@rickwierenga when you merge commits I would suggest editing the message body to only relevant lines

rickwierenga:
Thanks!

> @rickwierenga when you merge commits I would suggest editing the message body to only relevant lines

(removed all commit messages from commit message)

PR #226, opened 2024-08-22
Added z and z nesting heights to Costar flat bottom 360 µL plates

maraxen:
Also added support for using ResourceStack with PlateCarrier. Current implementation of ResourceStack usage with PlateCarrier only checks that the "top" (messily assumes that the resource is stacked in the z dimension) resource is in fact a plate. This aspect should probably be made more robust before merging, but the plate lid specs should be accurate.


rickwierenga:
the TODO is: plates will "sink" a little bit on some carriers*. This behavior is implemented when assigning a plate to the PlateCarrierSite. But when assigning a Plate to a ResourceStack, where the RS happens to be on a PlateCarrierSite, sinking is not implemented yet. The location of the ResourceStack wrt the PlateCarrierSite probably shouldn't change, meaning the location of the Plate wrt ResourceStack should change.

*from #205:
image

rickwierenga:
>  The location of the ResourceStack wrt the PlateCarrierSite probably shouldn't change, meaning the location of the Plate wrt ResourceStack should change.

or actually, maybe keeping the Plate wrt ResourceStack at 0,0,0 is easier to implement & reason about. That means changing the location of the ResourceStack wrt PlateCarrierSite. I think we can use callbacks for that. (When the lowest item in the ResourceStack changes, we should reposition the ResourceStack wrt PlateCarrierSite)

maraxen:
okay, I'll take a closer look at the existing logic for sinking and try to adapt that to Plates in a ResourceStack. I'll also modify the logic so that PlateAdapter is not included as a possible type allowed in a stack.

maraxen:
> > The location of the ResourceStack wrt the PlateCarrierSite probably shouldn't change, meaning the location of the Plate wrt ResourceStack should change.
> 
> or actually, maybe keeping the Plate wrt ResourceStack at 0,0,0 is easier to implement & reason about. That means changing the location of the ResourceStack wrt PlateCarrierSite. I think we can use callbacks for that. (When the lowest item in the ResourceStack changes, we should reposition the ResourceStack wrt PlateCarrierSite)

Just saw this, I think that makes sense. The actual use case for me is a lid glued to the bottom of a plate to help stabilize it, which I'm realizing it may be good to have logic that checks what is at the bottom of the stack and adjusts the ResourceStack position in reference to that. Does that sound reasonable?

rickwierenga:
> adjusts the ResourceStack position in reference to that. Does that sound reasonable?

Yep yep, changing the ResourceStack location wrt PlateCarrierSite sounds easiest to implement and reason about.

Technically it should be possible to implement the Lid-on-top-of-pedestal location, just like skirted plates are placed on pedestals, using the Lid.nesting_z_height attribute. Before that, it seems necessary to merge #205 into your branch.

I would add Lid as a case in `PlateCarrierSite._get_child_location`. And then use `PlateCarrierSite._get_child_location` called with `ResourceStack.children[0]` to determine what the location of ResourceStack wrt PlateCarrierSite should be. The code for repositioning ResoureStack based on its lowest item should be adaptive as its lowest child changes, so should be written using a callback.

Issue #179, opened 2024-06-26
nested tip racks


rickwierenga:
Following discussion in https://github.com/PyLabRobot/pylabrobot/pull/173, we will use `ResourceStack` for stacking resources. Nested tip racks are a special case and will get a `NestedTipRackStack(ResoruceStack)` class, exposing convenience methods like `count_tips`, and in the future a nice async generator. We will also introduce a `NestedTipRack(TipRack)` class that has a `nesting_height: float` property.

![image](https://github.com/PyLabRobot/pylabrobot/assets/36956818/6d0e7b24-c640-4836-94a1-5fed84738a08)

@BioCam will be working on this

PR #228, opened 2024-08-28
Create NestedTipRack Class

BioCam:
Hi everyone,

in this little PR I created a new PLR `Resource` class, the `NestedTipRack`, a direct subclass of `TipRack`.

`NestedTipRack` is almost identical to `TipRack` but distinguishes itself with the required specification of the attribute `stacking_z_height`.

NTR_stacking_explainer

This is needed to establish a new management system for stacking of NTRs and bigger-picture tip (access) management in a subsequent PR.
For this purpose, we should discuss whether we want to establish the management capabilities inside the existing management `Resource` `ResourceStack` or in a new management `Resource` which we could call `TipRackStack`.


BioCam:
Small note:

I have added  the ability to assign one NTR to another NTR at the moment.

- This is so I can already use NTR stacking right now (i.e. before we agreed upon a management system).
- This goes **against** what we previously discussed in terms of the deck tree: one resource having children of the same resource type (and will likely have to be removed once a management system has been put in place).

rickwierenga:
backref: https://github.com/PyLabRobot/pylabrobot/issues/179

rickwierenga:
sorry for delay and thank you for taking initiative & implementing this!

PR #229, opened 2024-08-29
[Labware] Add support for 3.20 mm ultra wide bore tips

fderop:


rickwierenga:
thanks

pylabrobot/resources/ml_star/plate_carriers.py
```diff
@@ -109,7 +109,7 @@ def PLT_CAR_L5AC_A00(name: str) -> PlateCarrier:
       ],
       site_size_x=127.0,
       site_size_y=86.0,
-      pedestal_size_z=-5.0
+      pedestal_size_z=4.78
```
rickwierenga, line 112: just saw this. why is this necessary?

tagging @BioCam because you might be impacted, i merged before having discussion sorry

fderop: Sorry, should have made this a separate PR, but basically this number did not work for the L5ACs we have in house. I have attached some pictures of an ultra wide bore tip at z=0 on top of an L5AC carrier and the same tip+channel on top of an adjacent L5MD to show that my calibration is good.

### L5AC:
IMG_20240829_165007_869

### L5MD:
IMG_20240829_165236_280

BioCam: To understand, @fderop:

1. You made the `pedestal_size_z` positive - this doesn't change anything about the computation (because for plate placement adjustment I made it take the absolute value anyway) but for consistency and helping others understand that this is a "sinking" z_value, could you please make it a negative value again?
2. You are saying that the `pedestal_size_z` is 220 microns smaller? I will measure this using cLLD_probing on Monday and share the measured data here. How many different tips have you tested this with and what method did you use for the measurements?
3. Regarding your explanation: I don't understand, you are only modifying `PLT_CAR_L5AC_A00`, what does this have to do with `L5MD`?

fderop: > 1. You made the `pedestal_size_z` positive - this doesn't change anything about the computation (because for plate placement adjustment I made it take the absolute value anyway) but for consistency and helping others understand that this is a "sinking" z_value, could you please make it a negative value again?

I will revert the change.

> 2. You are saying that the `pedestal_size_z` is 220 microns smaller? I will measure this using cLLD_probing on Monday and share the measured data here. How many different tips have you tested this with and what method did you use for the measurements?

I measured this on the carrier itself using calipers. I didn't know that the `pedestal_size_z` was converted to absolute.

> 3. Regarding your explanation: I don't understand, you are only modifying `PLT_CAR_L5AC_A00`, what does this have to do with `L5MD`?

The calibrations are relative. If the tip touches the surface of `L5MD` exactly when instructed to touch at `z=0` relative to `L5MD`, then we can assume that a touch at `z=0` on the surface of `L5AC` should be accurate as well assuming that both definitions are accurate.

PR #230, opened 2024-08-30
[Fix] Fix carrier PLT_CAR_L5AC_A00 pedestal_size_z

fderop:
Revert change introduced in #229


rickwierenga:
sorry, i missed this because of travel

PR #234, opened 2024-09-04
[rotations] fix angles computed by grip directions

jkhales:


rickwierenga:
for explicitness, these are now counter clockwise per convention

PR #235, opened 2024-09-05
clearly differentiate between local and absolute size

rickwierenga:


pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -2476,8 +2478,17 @@ async def iswap_release_picked_up_resource(
...
-    # Get center of source plate. Also gripping height and plate width.
-    center = location + resource.rotated(z=rotation).center() + offset
+    # Get center of source plate in absolute space.
+    # The computation of the center has to be rotated so that the offset is in absolute space.
+    center_in_absolute_space = Coordinate(*matrix_vector_multiply_3x3(
+      resource.rotated(z=rotation).get_absolute_rotation().get_rotation_matrix(),
+      resource.center().vector()
+    ))
+    # This is when the resource is rotated (around its origin), but we also need to translate
+    # so that the lfb of the plate is lfb in absolute space, not local.
```
jkhales, line 2488: what is lfb? worth spelling it out. abbreviations are generally a code smell

rickwierenga: left front bottom, used quite often in PLR but will spell out if better

rickwierenga: actually that's a lie, we dont write that in the repo. but i use it on gh often. will spell out

BioCam:
Hi everyone,

Could you please define what local and absolute size means?

rickwierenga:
local size is within a resource's own coordinate system (where origin is LFB), and obviously the rotation (wrt self) is 0.

global size is the size of the resource in the absolute (deck) coordinate system.

for example: if a resource is rotated by 90 degrees, its local x size will remain the same, but the global x size will now be the local y size.

BioCam:
Thank you that makes a lot of sense.

I recommend making some visualisations for this in the docs.

PR #236, opened 2024-09-07
fixed read_timeout for vantage

1dkone1:
I was trying to pick up the row of pipet tips and then place. There was the timeout error so the timeout was extended from 30 to 60 seconds.


rickwierenga:
thanks!

PR #239, opened 2024-09-10
[rotations] add ResourceHolderMixin to fix placement in a number of resources

jkhales:


jkhales:
@rickwierenga this fails for `ResourceStack` curious how you want to handle that

jkhales:
> why does it fail

my guess is that because `ResourceStack` doesn't use the automatic location calculation in `assign_child_resources` that `PlateCarrierSite` does. Wondering if this should be added to a parent class somewhere instead of playing whack a mole.

rickwierenga:
yes, i guess in that case we should also automatically find the best location of the plate, similar to what we do with CarrierSite

pylabrobot/liquid_handling/liquid_handler_tests.py
```diff
@@ -249,6 +254,50 @@ async def test_move_plate_onto_resource_stack_with_plate(self):
     self.assertEqual(plate2.location.z, 15)
     self.assertEqual(stack.get_absolute_size_z(), 30)
 
+  async def test_move_plate_rotation(self):
+    rotations = [0, 90, 270, 360]
+    grip_directions = [
+        (GripDirection.LEFT, GripDirection.RIGHT),
+        (GripDirection.FRONT, GripDirection.BACK),
+    ]
+    site_classes = [ResourceStack, PlateCarrierSite]
```
jkhales, line 263: anything else I should put in here?

jkhales: Yes. I like that.

jkhales: Will do it. Anything else you can think of? Tilter?

rickwierenga: tilter, heater shaker, shaker, temperature controllers, all these classes. potentially a lot easier to just create a new class SinglePlateTakingResource, override assign_child_resource there, and then have these on-deck/off-deck devices inherit from that

jkhales: Yea agreed

pylabrobot/plate_reading/plate_reader.py
```diff
@@ -48,12 +48,16 @@ def __init__(
...
       raise ValueError("There already is a plate in the plate reader.")
     if not isinstance(resource, Plate):
       raise ValueError("The resource must be a Plate.")
-    super().assign_child_resource(resource, location=location)
+
+    if location is not None:
+      # Directly call Machine's assign_child_resource method to allow for custom location
+      Machine.assign_child_resource(self, resource, location=location, reassign=reassign)
+    else:
+      super().assign_child_resource(resource, reassign=reassign)
```
jkhales, line 60: this mimics the previous behavior, but not sure that is truly the behavior we desire.

pylabrobot/resources/utils.py
```diff
@@ -199,21 +199,3 @@ def query(
       z=z,
     ))
   return matched
-
```
jkhales, line 202: moved to the mixin file

pylabrobot/resources/resource_holder.py
```diff
@@ -0,0 +1,36 @@
...
+    raise ValueError("Resource rotation must be a multiple of 90 degrees on the z axis")
+  location = {
+    0.0: Coordinate(x=0, y=0, z=0),
+    90.0: Coordinate(x=resource.get_size_y(), y=0, z=0),
+    180.0: Coordinate(x=resource.get_size_y(), y=resource.get_size_x(), z=0),
+    270.0: Coordinate(x=0, y=resource.get_size_x(), z=0),
+  }[resource.rotation.z % 360]
+  return location
+
+class ResourceHolderMixin:
```
jkhales, line 23: using a mixin avoids complexity arising from "diamond inheritance"

rickwierenga: how do we make sure `ResourceHolderMixin.assign_child_resource` has the same method header as `Resource`?

jkhales: i'm not sure i follow. isnt this the same as with a normal subclassing?

jkhales: bumping this

jkhales: any more thoughts on this @rickwierenga?

pylabrobot/machines/machine.py
```diff
@@ -38,6 +38,7 @@ def __init__(
     backend: MachineBackend,
     category: Optional[str] = None,
     model: Optional[str] = None,
+    **_kwargs
```
rickwierenga, line 41: kwargs seems not needed here, but maybe an artifact of using mixins?

pylabrobot/plate_reading/plate_reader.py
rickwierenga: I think we should implement this check in the mixin: it shouldn't update the location if one is explicitly passed

pylabrobot/resources/plate.py
```diff
@@ -44,7 +46,7 @@ def serialize(self) -> dict:
     return {**super().serialize(), "nesting_z_height": self.nesting_z_height}
 
 
-class Plate(ItemizedResource[Well]):
+class Plate(ResourceHolderMixin, ItemizedResource[Well]):
```
rickwierenga, line 49: why does plate need this? i thought this class was specifically for resources that hold plates

pylabrobot/resources/resource_holder.py
```diff
@@ -0,0 +1,36 @@
...
+  """
+  A mixin class for resources that can hold other resources, like a plate or a lid
+  """
```
rickwierenga, line 26: maybe specify that this applies translations on rotation

pylabrobot/resources/carrier.py
```diff
@@ -241,7 +240,7 @@ def get_plate_sinking_depth(plate: Plate):
       resource.register_did_assign_resource_callback(self._update_resource_stack_location)
       self.register_did_unassign_resource_callback(self._deregister_resource_stack_callback)
 
-    return get_child_location(resource) - Coordinate(z=z_sinking_depth)
```
rickwierenga, line 244: _get_child_location would mean the location is explicitly specified, which should mean that the mixin doesn't apply (that's only for default values). so I think we should keep get_child_location here

pylabrobot/machines/machine.py
jkhales: When you have a bunch of subclasses with different args, they get passed up the MRO chain. This is the standard way to "ignore them"

pylabrobot/resources/carrier.py
jkhales: seems like the opposite case in line 209?
```
    location = location or self._get_child_location(resource)
```

jkhales:
bumping this

pylabrobot/machines/machine.py
rickwierenga: this feels like bad design. why is it necessary?

pylabrobot/resources/carrier.py
```diff
@@ -240,8 +238,10 @@ def get_plate_sinking_depth(plate: Plate):
         z_sinking_depth = get_plate_sinking_depth(first_child)
       resource.register_did_assign_resource_callback(self._update_resource_stack_location)
       self.register_did_unassign_resource_callback(self._deregister_resource_stack_callback)
+    return -Coordinate(z=z_sinking_depth)
```
rickwierenga, line 241: _get_sinking_depth should not handle callbacks

rickwierenga: made an issue for this: https://github.com/PyLabRobot/pylabrobot/issues/246

pylabrobot/resources/resource_stack.py
```diff
@@ -95,26 +96,42 @@ def get_actual_resource_height(resource: Resource) -> float:
       return max(get_actual_resource_height(child) for child in self.children)
     return sum(get_actual_resource_height(child) for child in self.children)
 
-  def assign_child_resource(self, resource: Resource, location: Optional[Coordinate] = None,
-    reassign: bool = False):
+
+  def get_resource_stack_size(self) -> Coordinate:
     if self.direction == "x":
       resource_location = Coordinate(self.get_size_x(), 0, 0)
```
rickwierenga, line 102: potentially just put this in get_default_child_location. It seems incorrect to have a get_size method return a coordinate

pylabrobot/resources/resource_stack.py
```diff
@@ -95,26 +96,42 @@ def get_actual_resource_height(resource: Resource) -> float:
...
+    if len(self.children) > 0:
+      top_item = self.get_top_item()
+      if isinstance(resource, Lid) and isinstance(top_item, Plate):
+        resource_location = self.get_resource_stack_size()
+        resource_location.z -= resource.nesting_z_height
+        top_item.assign_child_resource(
+          resource,
+          location=get_child_location(resource) + resource_location
+        )
+        return
```
rickwierenga, line 128: I thought we were gonna remove this

jkhales: that's def a separate PR in my opinion. It's also a breaking change.

rickwierenga: ok

pylabrobot/resources/resource_stack.py
jkhales: it's used in the assign lid to resource stack special case. I can rename to something if you don't like the `size`  word.

rickwierenga: for this particular function i don't like it :)

pylabrobot/resources/carrier.py
jkhales: i'll add the issue to the code in a comment as well

pylabrobot/resources/resource_stack.py
jkhales: https://github.com/PyLabRobot/pylabrobot/issues/247

pylabrobot/machines/machine.py
jkhales: I agree, it feels weird, but I believe it's just how python does things

jkhales: 🤖 also agrees image

PR #245, opened 2024-09-17
add rotation around origin for Rotation class

ben-ray:
not sure rotation.py is the right home for the test
@fderop for reference, i think u were thinking about this too?


pylabrobot/resources/rotation.py
```diff
@@ -1,36 +1,47 @@
...
-    self.z = z  # around z-axis, yaw
+    self.x = x 
+    self.y = y
+    self.z = z
 
-  def get_rotation_matrix(self):
-    # Create rotation matrices for each axis
-    Rz = ([
+  def get_rotation_and_translation(self, origin: Coordinate = Coordinate.zero()):
+    # rotation matrices for each axis
```
rickwierenga, line 14: I was thinking about storing the origin as an attribute on the class itself

pylabrobot/resources/rotation.py
```diff
@@ -1,36 +1,47 @@
...
+    rotated_origin = [
+      sum(rotation_matrix[i][j] * origin_vector[j] for j in range(3)) for i in range(3)
+    ]
+    translation = [
+      rotated_origin[0] - origin.x,
+      rotated_origin[1] - origin.y,
+      rotated_origin[2] - origin.z
+    ]
+
+    return rotation_matrix, translation
```
rickwierenga, line 44: we need to return the rotation because otherwise it would break get_absolute_rotation and other functions elsewhere

pylabrobot/resources/rotation.py
ben-ray: good idea

pylabrobot/resources/rotation.py
ben-ray: hmm idk how to do this without a 4x4 matrix or a second translation matrix in addition to the rotation matrix

4x4 would just add the translation as a column, but would then have to be implemented in get_absolute_rotation and others?
```
rotation_matrix_4x4 = [
            [rotation_matrix_3x3[0][0], rotation_matrix_3x3[0][1], rotation_matrix_3x3[0][2], translation.x],
            [rotation_matrix_3x3[1][0], rotation_matrix_3x3[1][1], rotation_matrix_3x3[1][2], translation.y],
            [rotation_matrix_3x3[2][0], rotation_matrix_3x3[2][1], rotation_matrix_3x3[2][2], translation.z],
            [0, 0, 0, 1]  # Last row for homogeneous transformation
        ]
```

Issue #225, opened 2024-08-22
remove Hamilton liquid classes


rickwierenga:
Liquid classes in the Hamiltonian sense are confusing, especially in PLR: https://labautomation.io/t/how-does-pylabrobot-handle-liquid-classes/977

A lot of code in STAR looks like this:

```python
flow_rates = [
      op.flow_rate or (hlc.aspiration_flow_rate if hlc is not None else 100)
        for op, hlc in zip(ops, hamilton_liquid_classes)]
```

Hamiltonian liquid classes are effectively dicts of parameters.

We should do the following instead:

```python
lh.aspirate(..., **hamilton_liquid_class.generate_kwargs())`
```

```python
>>> hamilton_liquid_class.generate_kwargs()
{
  "flow_rate": 42,
  "clot_detection_height": 69,
  ...
}
```

so that we can remove all references to Hamilton liquid classes in `STAR`/`Vantage` backends, thus simplifying them a lot. `HamiltonLiquidClass` can simply be a [TypedDict](https://peps.python.org/pep-0589/).

(I vaguely recall discussing this before, possibly in a private message, can't find a public issue/forum post now)

rickwierenga:
one issue with making it a TypedDict is that volume correction becomes slightly more awkward, and the user needs to explicitly say

```python
lh.aspirate(vols=[hlc.correct(original_volume)])
```

but i think for sensitive liquid handling you'll want to fine-tine the volumes anyway, and this method is a lot more transparent than doing it somewhere deep in STAR where it's hidden

rickwierenga:
first PR: https://github.com/PyLabRobot/pylabrobot/pull/248

BioCam:
> one issue with making it a TypedDict is that volume correction becomes slightly more awkward, and the user needs to explicitly say
> 
> ```python
> lh.aspirate(vols=[hlc.correct(original_volume)])
> ```

I agree; this puts enormous technical debt on the end user.

This way every user must constantly think about the actual piston_volume, i.e. the "corrected_volume", one has to move to achieve the target_liquid_volume.

The excellent idea behind standard liquid classes is to not have worry about this.
It makes reading and writing complex protocols much harder.

Ideally we want a system/advanced_liquid_class that is **more** than a dictionary, i.e. an object that corrects not just for volume but corrects all aspirate/dispense attributes... because why would the same blow_out_volume for 1 ul of 50% glycerol work the same way for 1000 of 50% glycerol.

[...]

PR #248, opened 2024-09-18
hamilton liquid classes make kwargs

rickwierenga:


rickwierenga:
[more obvious link to the original issue](https://github.com/PyLabRobot/pylabrobot/issues/225)

new behavior:

```python
>>> from pylabrobot.liquid_handling.liquid_classes.hamilton.star import StandardVolumeFilter_Water_DispenseSurface
>>> hlc = StandardVolumeFilter_Water_DispenseSurface
>>> hlc.make_asp_kwargs(num_channels=2)
{'flow_rates': [100.0, 100.0], 'mix_speed': [100.0, 100.0], 'transport_air_volume': [5.0, 5.0], 'blow_out_air_volume': [0.0, 0.0], 'swap_speed': [2.0, 2.0], 'settling_time': [1.0, 1.0], 'clot_detection_height': [0.0, 0.0]}
>>> hlc.make_asp96_kwargs()
{'flow_rate': 100.0, 'mix_speed': 100.0, 'transport_air_volume': 5.0, 'blow_out_air_volume': 0.0, 'swap_speed': 2.0, 'settling_time': 1.0}
```

can be used like

```python
await lh.aspirate(..., **hlc.make_asp_kwargs(3))
```

⚠️ backends no longer automatically find these values ⚠️

rickwierenga:
for volume correction using the curve:

```python
>>> hlc.compute_corrected_volume(42)
46.593333333333334
```

BioCam:
> ```python
> corrected_volume = hlc.compute_corrected_volume(42)
> await lh.aspriate(..., vols=[corrected_volume])
> ```

How does this deal with multi-channels, i.e. lists of volumes?

Does this mean the corrected_volume has to be handed to the aspirate function directly?

rickwierenga:
> How does this deal with multi-channels, i.e. lists of volumes?

It's still just a list of floats. different ways to handle it, eg:

```py
corrected_volumes = [hlc.compute_corrected_volume(x) for x in [42, 21]]
await lh.aspirate(..., vols=corrected_volumes)
```

> Does this mean the corrected_volume has to be handed to the aspirate function directly?

yes

BioCam:
Does this mean the previous handling of HamiltonLiquidClass is completely absolute with this PR?
i.e. all protocols developed in PLR are now not functional anymore?

Can you please elaborate on this?
```python
await lh.aspirate(..., vols=corrected_volumes)
```
What does a complete aspiration dispensation procedure look like?

I didn't like the HamiltonLiquidClass either but I am wondering what is the advantage of this much longer implementation?

rickwierenga:
> Does this mean the previous handling of HamiltonLiquidClass is completely absolute with this PR?

The previous automatic handling of liquid classes was removed.

If you use liquid classes, as do we, there is simply a new api for them: instead of passing `hlc=something`, you use `**hlc.make_asp_kwargs()`, to preserve the parameter sets people have already created.

> What does a complete aspiration dispensation procedure look like?

It depends on whether you want to use a liquid class, or specify parameters yourself, or take backend defaults.

1. use a liquid class

```python
from pylabrobot.liquid_handling.liquid_classes.hamilton.star import StandardVolumeFilter_Water_DispenseSurface
hlc = StandardVolumeFilter_Water_DispenseSurface
corrected_volumes = [hlc.compute_corrected_volume(x) for x in [42, 21]]
await lh.aspirate(some_wells, vols=corrected_volumes, **hlc.make_asp_kwargs(2))
```

2. specify yourself

```python
def my_correction_curve(x): return x + 1
corrected_volumes = [my_correction_curve(x) for x in [42, 21]]
await lh.aspirate(some_wells, vols=corrected_volumes, flow_rates=[21, 21], mix_speed=[1, 1]) # etc.
```

3. use backend default

```python
await lh.aspirate(some_wells, vols=[42, 21])
```

> I didn't like the HamiltonLiquidClass either but I am wondering what is the advantage of this much longer implementation?

In PLR, the implementation is much shorter and simpler. This PR removed 198 lines of code. Before, we had four sources of values, in order or precedence:
1. user specified
2. user specified liquid class
3. automatically inferred liquid class
4. backend default

Now it's all user-specified (either explicitly or through hlc.make_asp_kwargs()), or the backend gives a default.

Had the liquid classes been accurate, this would have been nice. But users still needed to fine tune their liquid transfers, and so the automatic handling does not make a substantial difference.

Another benefit is the code is now explicit. Before, the behavior was hidden and not obvious. (for example, when aspirating 100uL twice and dispensing 200uL, the dispensing drive would not reset, because the correction curve is non linear, so `2*f(100)!=f(2*100)`.

PR #264, opened 2024-09-27
Revert "hamilton liquid classes make kwargs"

rickwierenga:
Reverts PyLabRobot/pylabrobot#248


rickwierenga:
i will work on a slightly better replacement before merging this

PR #269, opened 2024-09-28
Fix iswap rotation issue

rickwierenga:
using iswap to grip ƒrom front (or back) and put to back (or front) was putting in the wrong location before


jkhales:
seems like worth figuring out why previous tests didn't catch this and fixing it

rickwierenga:
is what i am doing

rickwierenga:
because it is a square

rickwierenga:
LOL

fderop:
> because it is a square

when float rotations are added, make sure to change the test to a circle 😂

rickwierenga:
float/continuous rotations are already supported https://github.com/PyLabRobot/pylabrobot/pull/195

PR #276, opened 2024-10-11
skip setup completely if centrifuge is already initialized 

Ph1so:
Still needs to be checked on actual centrifuge


pylabrobot/centrifuge/vspin.py
```diff
@@ -39,110 +39,111 @@ async def setup(self):
...
+      await self.send(b"\xaa\x01\x13\x20\x34")
+      await self.send(b"\xaa\x00\x21\x02\xff\x22")
+      await self.send(b"\xaa\x02\x13\x20\x35")
+      await self.send(b"\xaa\x00\x21\x03\xff\x23")
+      await self.send(b"\xaa\xff\x1a\x14\x2d")
+      self.dev.baudrate = 57600
+      self.dev.ftdi_fn.ftdi_setrts(1)
+      self.dev.ftdi_fn.ftdi_setdtr(1)
+      self.get_status()
+    except:
```
rickwierenga, line 56: what kind of error is raised? / why is `except` needed?

Ph1so: In status_check, an error is raised when nothing is returned from the status check. So if nothing is returned the centrifuge isn't init and the error will trigger the except, which is the actual init

rickwierenga:
i had to change the check from 0x00 to b''.

after that, i get the following error when i use setup:

```py
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[2], [line 4](vscode-notebook-cell:?execution_count=2&line=4)
      [2](vscode-notebook-cell:?execution_count=2&line=2) cf = Centrifuge(name = 'centrifuge', backend = VSpin(bucket_1_position=240, device_id="FTDHY54N"), size_x= 1, size_y=1, size_z=1)
      [3](vscode-notebook-cell:?execution_count=2&line=3) # cf = Centrifuge(name = 'centrifuge', backend = AgilentCentrifuge(bucket_1_position=10), size_x= 1, size_y=1, size_z=1)
----> [4](vscode-notebook-cell:?execution_count=2&line=4) await cf.setup()

File ~/retro/phi-plr/pylabrobot/machines/machine.py:63, in Machine.setup(self, **backend_kwargs)
     [62] async def setup(self, **backend_kwargs):
---> [63]   await self.backend.setup(**backend_kwargs)
     [64]   self._setup_finished = True

File ~/retro/phi-plr/pylabrobot/centrifuge/vspin.py:115, in VSpin.setup(self)
    [113]   await self.send(b"\xaa\x02\x0e\x10")
    [114]   stat = await self.send(b"\xaa\x01\x0e\x0f")
--> [115]   resp = stat[0]
    [117] await self.send(b"\xaa\x01\x0e\x0f")
    [118] await self.send(b"\xaa\x01\x0e\x0f")

IndexError: index out of range
```

this is after a machine power cycle

Ph1so:
Hmmm.Let me test this out at the lab.

PR #287, opened 2024-10-22
remove mystery resources and add docs

rickwierenga:
similar to https://github.com/PyLabRobot/pylabrobot/pull/253

implementing https://github.com/PyLabRobot/pylabrobot/issues/170


rickwierenga:
after this, every resource in PLR (with the exception of some Hamilton carriers and the Tecan resources) should have part numbers and documentation w/ a picture

PR #300, opened 2024-11-10
add teaching needle block

rickwierenga:
```python
teaching_tip_rack = lh.deck.get_resource("teaching_tip_rack")
await lh.pick_up_tips(teaching_tip_rack["A1:H1"])
```

located on rails 31 on STARLet, rail 55 on STAR (both past last submarine)

### DIFF
--- a/pylabrobot/resources/hamilton/hamilton_deck_tests.py
+++ b/pylabrobot/resources/hamilton/hamilton_deck_tests.py
@@ -62,6 +62,9 @@ class HamiltonDeckTests(unittest.TestCase):
           │   ├── <empty>
           │   ├── <empty>
           │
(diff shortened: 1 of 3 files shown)


BioCam:
Beautifully implemented, thank you @rickwierenga!

I was about to start this today to facilitate the new ztouch_probing 😅 

To ensure I understand everything in here:

You extended the rail system to include a "teaching carrier", which is the panel at the righthandside of the machine that houses the
- teaching needle rack,
- the place&shift metal strip for tips discarding, and
- the the gripper parking position.
You also modelled the metal teaching needles as 300ul_tips/standard_tips.

Is that correct?

rickwierenga:
yes. the 'carrier' only includes the teaching needle rack atm, but we can possibly have it be the parent of the trash and grippers as well.

PR #306, opened 2024-11-15
Better typehint for need_setup_finished

jrast:
The typehint on `need_setup_finished` lead to incomplete information on autocomplete in VSCode:
2024-11-15_12h42_46

With this change, the sugestion is correct / complete:
2024-11-15_12h45_20

If PyLabRobot still supports Python < 3.10, the ParamSpec will not be available in the typing package and typing extensions need to be used:
```python
if sys.version_info < (3, 10):
    from typing_extensions import ParamSpec
else:
    from typing import ParamSpec
```

Note: `python_requires` is missing in `setup.py`.


jrast:
The typehint is not correct, will look at it...

jrast:
Ok, with the latest commit the typehint should be fine. `mypy` now reports some issues in other places as the typehint of decorated functions is now correct.

rickwierenga:
thank you!

PR #310, opened 2024-11-19
git precommit configuration

rickwierenga:
- dont use typecheck
- use repo instead of command line tool


rickwierenga:
thanks for comments

> Running pre-commit run --all-files leads to some changes in the tools directory.

ah, the Makefile only does the pylabrobot module. i will make both run on both.

> available (0.7.4 vs 0.2.1

yes, will update in a separate commit because a lot of files will change

PR #315, opened 2024-11-21
add nest_1_troughplate_195000uL_Vb and nest_1_troughplate_185000uL_Vb

rickwierenga:


ashah03:
Wow, that was fast, thank you! Part numbers should be 360103 and 360104. Will try to test this out in the next two days and let you know if there are any issues.

By the way for clarity, I would say that it's not necessarily 96/384 "holes" but instead there are 96 trapesoidal shapes at the bottom.
image

mackenziekormann:
Hi, Rick! I'm Mackenzie, working with Adit on our Hamilton automation setup. I tested out the trough today (sorry for the delay) and the two furthest tips away from the front (I would say channels 0 and 1) aren't entering the trough, only 2 through 7.

rickwierenga:
could you add offsets?

```
lh.aspirate(wells, vols, offsets=[Coordinate(x0, y0, z0), Coordinate(x1, y1, z1), ...])
```

x,y,z in mm, per channel starting at 0

are there any offsets for which it works?

mackenziekormann:
Alrighty, I tried setting offsets, but the error I'm getting is `ValueError: Minimum distance between two y positions is <9mm` regardless of the offset I enter.

mackenziekormann:
Tried a few iterations of offsets and nothing worked. Still getting the error along the lines of `ValueError: Minimum distance between two channels is <9mm` or something akin to that

mackenziekormann:
Wanted to check in here again about the error I was getting. It's happening with the 12 channel trough as well, with the same error. Confused as to why the channels can dispense close together but can't aspirate that closely.

rickwierenga:
when i use the STARChatterboxBackend (same as STAR, but prints commands instead of sending them), it works:

```python
from pylabrobot.liquid_handling.backends.hamilton.STAR_chatterbox import STARChatterboxBackend
from pylabrobot.resources import PLT_CAR_L5PCR, Hamilton_96_adapter_188182, Coordinate, STARLetDeck
lh = LiquidHandler(backend=STARChatterboxBackend(), deck=STARLetDeck())
await lh.setup()

from pylabrobot.resources import HTF
tr = HTF(name="tr")
lh.deck.assign_child_resource(tr, location=Coordinate(0, 0, 0))
await lh.pick_up_tips(tr["A1:H1"])

from pylabrobot.resources import nest_1_troughplate_195000uL_Vb
trough = nest_1_troughplate_195000uL_Vb(name="trough")
lh.deck.assign_child_resource(trough, location=Coordinate(300, 0, 100))
await lh.aspirate(trough["A1"]*8, vols=[100]*8)

from pylabrobot.resources import nest_1_troughplate_185000uL_Vb
trough1 = nest_1_troughplate_185000uL_Vb(name="nest_1_troughplate_185000uL_Vb")
lh.deck.assign_child_resource(trough1, location=Coordinate(600, 0, 100))
await lh.aspirate(trough1["A1"]*8, vols=[100]*8)
```

(need to sync because of https://github.com/PyLabRobot/pylabrobot/commit/974073c759660aa85afdb5171c8710486eedbd36)

rickwierenga:
this was dumb. i think i fixed it

mackenziekormann:
Thanks for the fix! Unfortunately now running into the issue of the tips running into the bottom of the trough when pipetting and causing an error.

mackenziekormann:
I can make a separate issue for this, but `ValueError: Minimum distance between two y positions is <9mm: 2369, 2280 (channel 1 and 2)` is the error I get when I attempt to use 7 tips in one channel of the 12 channel trough

mackenziekormann:
Similarly to the error with pipetting into the 96 wellplate, I'm having issues with tips running into the bottom of the single channel trough. Setting Coordinate=(0, 0, 6) changes the absolute location when checked, but does not effect the tips hitting the bottom of the trough.

rickwierenga:
there were two issues above causing the 'min position error': 1) nest_1_troughplate_195000uL_Vb, like all plates, is defined in landscape orientation. This means that the robot will try to fit all channels into the well on the narrow side. To fix, we need to rotate the plate:

```python
rom pylabrobot.resources import nest_1_troughplate_195000uL_Vb
trough = nest_1_troughplate_195000uL_Vb(name="trough").rotated(z=90)  # <---
lh.deck.assign_child_resource(trough, location=Coordinate(300, 0, 100))
await lh.aspirate(trough["A1"]*8, vols=[100]*8)
```

second, LH didn't compute the spacing in rotated resources correctly. fixed with #326

ashah03:
Not sure I understand why it needed to be rotated - it is the same orientation as most plates (landscape) on the deck. The video Mackenzie sent is going sideways (facing x direction of the robot).

ashah03:
The goal is to fit the 8 channels on the narrow side, which we thought would work since it's about the same dimensions as a plate. But maybe it's a bit too small to fit 8 channels in?

rickwierenga:
so you're aspirating from different troughs? in that case you need to use `["A1:H1"]` to get all troughs instead of a single `["A1"]`

ashah03:
It’s all one trough so this shouldn’t be applicable.

rickwierenga:
oh oh this is about the 1 trough plate smh

the y size of the 96-grid version is 72 mm (per drawing - can you caliper reality?), which just isn't big enough to fit 8 channels at 9 mm spacing

rickwierenga:
mm, it should be big enough actually. each channel would have 9mm of space. we were trying to split into 9mm spacing on the edges as well

rickwierenga:
sorry for delay, end of year is busy. does https://github.com/PyLabRobot/pylabrobot/pull/337 fix? may need to tweak the `min_spacing_edge` variable to 0 for 72mm to work. Just hack around with the value and see what works physically.

PR #324, opened 2024-12-05
add STAR.position_channels_in_y_direction

rickwierenga:


pylabrobot/liquid_handling/backends/hamilton/STAR.py
```diff
@@ -7465,6 +7466,29 @@ async def rotate_iswap_wrist(self, orientation: WristOrientation):
       tp=orientation.value,
     )
 
+  @staticmethod
+  def channel_id(channel_idx: int) -> str:
+    """channel_idx: plr style, 0-indexed from the back"""
+    channel_ids = "123456789ABCDEFG"
```
ben-ray, line 7472: generate this programmatically with: chr(55 + i)

ben-ray: not sure there is any hamilton with >16pip, so possibly useless

rickwierenga: more than 16 pip is impossible because the board has 16 connectors (im pretty sure)

PR #327, opened 2024-12-09
Add `{Well,TipSpot}.get_identifier`

rickwierenga:
https://discuss.pylabrobot.org/t/how-do-i-get-a-well-coordinate-a1-row-col-etc-from-a-resource/95


rickwierenga:
some files were importing Well from plate.py. I had to make the well import in plate to be type-checking-only to avoid circular imports. -> i also had to fix other files to import Well from well.py instead of plate.py

PR #328, opened 2024-12-09
Split LiquidHandlerBackend.move_resource

rickwierenga:
Move resource was not a very atomic command, as it can always be split up in
- pick_up_resource
- move_picked_up_resource
- drop_resource

in some protocols, it is required to execute actions while a resource is being picked up (such as opening a centrifuge door), but not to proceed with the drop until the other action is completed. By splitting, this is supported in the front end LiquidHandler.

Using the backend this was obviously already possible, but then you don't get state management.

The existing methods keep working (I am renaming some parameters).


rickwierenga:
renamed:
- resource_offset -> pickup_offset
- get_direction -> pickup_direction
- put_direction -> drop_direction

PR #331, opened 2024-12-10
move_resource can use iswap hotel mode on STAR

rickwierenga:
`lh.move_plate(plate, dest, use_arm="iswap_hotel_unsafe")` to use commands PO and PI instead of PP and PR when using the STAR backend


ben-ray:
based!

ben-ray:
it is nice to be able to set open_gripper_position precisely, it can help avoid crashes on a tight hotel

rickwierenga:
added

PR #332, opened 2024-12-10
add cytomat

jkhales:


rickwierenga:
todos:

- [ ] tests
- [ ] user guide
- [ ] defining racks from the plr model

PR #337, opened 2024-12-20
fix channel spacing in single resource aspirations/dispenses

rickwierenga:
before when aspirating or dispensing with multiple channels in a single container, we always divided the available space in a container by num_channels+1 to space channels evenly. This works when the container is sufficiently large, but does not work when channel is barely big enough to fit the number of channels.

For example, aspirating with 8 channels from a 80 wide mm resource would have meant a space of 80/(8+1)<9 mm between channels. In reality, there is enough space if you take 9 mm between channels, 80-7*9 leaves 17 mm of space, which is 8.5mm per side.

This PR fixes such cases, by spacing channels >9mm apart and seeing if there is enough space left on the sides. I set the minimum space on the side to 2mm per side, the edge of the container to the center of the tip, which may need to be adjusted based on feedback. If there is enough space left, the aspiration or dispense will happen. If there is more space than 9mm left on either side, we will split up the channels equally across the resource, similar to the behavior that we had before. For example, aspirating with 4 channels from the 80mm resource will space all channels 20mm apart, and 20mm from the side.


mackenziekormann:
Tested this on our STAR today. Aspiration works great! Dispensing unfortunately does not.

rickwierenga:
smh i forgot to implement for dispense hold on

rickwierenga:
try again?

mackenziekormann:
works, thanks!

rickwierenga:
thanks for testing!

PR #343, opened 2024-12-23
more iswap backend kwargs, hotel

rickwierenga:


rickwierenga:
changes:

added to STAR.pick_up_resource:
- `iswap_grip_strength: int = 4`
- `minimum_traverse_height_at_beginning_of_a_command: Optional[float] = None`
- `z_position_at_the_command_end: Optional[float] = None`
- `plate_width_tolerance: float = 2.0`
- `hotel_open_gripper_position: Optional[float] = None`
- `hotel_depth=160.0`
- `hotel_clearance_height=7.5`
- `high_speed=False`
- `plate_width: Optional[float] = None`
- `use_unsafe_hotel: bool = False`
- `iswap_collision_control_level: int = 0`
- `iswap_fold_up_sequence_at_the_end_of_process: bool = True`

added to STAR.drop_resource:
- `minimum_traverse_height_at_beginning_of_a_command: Optional[float] = None`
- `z_position_at_the_command_end: Optional[float] = None`
- `hotel_open_gripper_position: Optional[float] = None`
- `hotel_depth=160.0`
- `hotel_clearance_height=7.5`
- `hotel_high_speed=False`
- `use_unsafe_hotel: bool = False`
- `iswap_collision_control_level: int = 0`

the resource-tree update logic was moved form LiquidHandler.move_{plate,lid} into LiquidHandler.{pick_up,drop}_resource so that resources are put in the correct spots and the tree is accurate, also when the user doesn't use the highest levels methods. I am tempted to remove LiquidHandler.move_{plate,lid} because they only provide a default value for the parameter `pickup_distance_from_top` at this point (the default value can be in LiquidHandler.pick_up_resource instead.

I also deleted STAR.{iswap_pick_up_resource,iswap_release_picked_up_resource} and put the logic there into STAR.pick_up_resource and STAR.drop_resource. STAR.pick_up_resource and STAR.drop_resource now call the lowest-level firmware-command-implementations `iswap_get_plate` and `iswap_put_plate`, removing one layer (simple is better). This also allows the sharing of that logic with the hotel-variants of iswap commands.

PR #344, opened 2024-12-28
add lock, unlock, and temp to shaker

ben-ray:


rickwierenga:
For temp stuff: there's actually a HeaterShaker class for this already ;)

Will merge lock

PR #349, opened 2025-01-03
fix lh summary for incubator

ben-ray:


rickwierenga:
fixed with https://github.com/PyLabRobot/pylabrobot/commit/2ac1c0ad70ba296817900d4909cbaf881c8b9474

PR #350, opened 2025-01-03
Fix teaching needle block positions on STARlet

BioCam:
Hi everyone - Happy New Year,

I tried out 

```python
teaching_tip_rack = lh.deck.get_resource("teaching_tip_rack")
await lh.pick_up_tips(teaching_tip_rack["A1:H1"])
```
from [PR#300 - add teaching needle block](https://github.com/PyLabRobot/pylabrobot/pull/300)

but it crashed:

### Problem & Proposed Fixes

1. The `TipSpot` positions were 9mm too low in the y-dimension -> added a `+9` to the y-location.
2. The indexing was mixed up:
    a. Indexing was completely "row A"-based, i.e. ['A8', 'A7', 'A6', 'A5', 'A4', 'A3', 'A2', 'A1']. This is inconsistent with the y-dimension being typically seen as a column, and the example code calls for `teaching_tip_rack["A1:H1"]`
    b. Indexing was reversed: the front-most `TipSpot` was `0`, meaning that `await lh.pick_up_tips(teaching_tip_rack["A1:H1"])` is impossible because the back-most channel (i.e. 0) tries to pick up the front-most tip (because it is indexed as 0), while channel_1 is called to pickup the 2nd-front-most tip, ...

-> I fixed both issues with 
```python
ordered_items={f"{letter}1": tip_spots[idx] for idx, letter in enumerate("HGFEDCBA")},
```

This assigns the entire teaching_tip_rack the row-index positions  ["A1", "B1", "C1", "D1", "E1", "F1", "G1", "H1"], in a back-to-front orientation that enables pickup of all 8 teaching needles in parallel.

### Next Steps

Could someone please check on their machine whether these changes yield the above mentioned behaviour on their machine?

I wonder whether these positions could be different on different machines (I tested them on a 2021-built STARlet).
If yes, this could explain why the positions were frontwards shifted by 9mm/1 tip in the y-dimension.

If no, then I'll remove the glaring `+9` addition, integrate the value directly into the code, and we can update this on PLR:main.

Happy automation :)


ben-ray:
there are indeed more than one version of the waste block, and we can take note to add in the (much) older waste block versions if the teaching needle position is also different. thanks camillo!

rickwierenga:
i will test Camillo's values in lab. i'm pretty sure these values were working before, but we have multiple robots to test on and i will see if there are obvious differences/versions in the teaching racks we have

BioCam:
I should also have another little tool ready to help us identify these differences between machines automatically by Wednesday 👀

rickwierenga:
yes, your values worked flawlessly on the machine i just tested this on. it seems most of our STARs/STARLets use this same waste block, except the STARLet which has the 2004-iswap which I must have used for the initial calibration. I think it's fine to just use the recent wasteblock, and revisit this if we ever need multiple versions to work. thanks!

rickwierenga:
>  identify these differences between machines automatically by Wednesday 

even before wednesday

PR #351, opened 2025-01-04
remove init cytomat from setup, increase timeout on init when called directly

ben-ray:
cytomats are typically configured to self-initialize on power-up. this means we don't typically need to call init() on startup. plus, this unnnec. init adds 1-2 minutes every time the jupyter kernel is refreshed and the device needs to be connected

on some cytomats, init can take longer, so i doubled the default timeout to 2 minutes


rickwierenga:
thanks!

PR #352, opened 2025-01-05
Merge ml_star and hamilton resource folders

BioCam:
Hi everyone,

A small cleanup PR:
Currently there are two separate `pylabrobot/resources/` folders for storing resource information from the manufacturer Hamilton:
- `ml_star`
- `hamilton`

`ml_star` is...
1. a remnant of the early days of PLR,
2. inconsistent with the folder nomenclature for manufacturer's resources (i.e. confusing to beginners),
3. contains Hamilton resource information that is not just used by the "Microlab STAR" / ml_star series of machines but also by other Hamilton machines (e.g. Vantage, MLPrep, Nimbus, ...).

### PR Content

I moved all information from `ml_star` into the `hamilton` folder, and updated the necessary imports


rickwierenga:
thanks Camillo!

> a remnant of the early days of PLR,

as background: more specifically, i used this name because it's used in venus like that. (venus has ml_star and hamilton in separate folders, which the conversion script copied)

BioCam:
I haven't updated the docs yet to reflect the resource folder structure.

Should we do this in this PR as well?

rickwierenga:
yes, i can do that easily

right now i am adding deprecation warnings to the deleted files

BioCam:
> right now i am adding deprecation warnings to the deleted files

Do you think this is needed?
...I guess I don't know how many people `from pylabrobot.resources import resource_x` vs  `from pylabrobot.resources.ml_star import resource_x`

rickwierenga:
thanks!

rickwierenga:
> Do you think this is needed?

jj tells us we have to be respectful of all users 🙄

PR #359, opened 2025-01-15
sort kwargs in lh.move_resource

rickwierenga:
both `LiquidHandlerBackend.{pick_up,drop}_resource` take can kwargs. These are often different (eg channels for core-gripper pickup in STAR - not an argument to drop). In the past, lh would send the full set of backend_kwargs to both mehods which caused an error for some backend methods if only one of {pick_up,drop} accepted a kwargs. This PR fixes that by removing kwargs not accepted by the backend method before passing them.

One complication is truly unnecessary params are removed in both cases, without an error to the user. I might add a special error case in lh.move_resource, but this isn't the prettiest. tbd

ref: i split move_resource as a backend method into separate methods in https://github.com/PyLabRobot/pylabrobot/pull/328


BioCam:
This seems to fix the broken CO-RE gripper movement.
All my move commands have been working again in this branch.

Thank you, @rickwierenga!

PR #361, opened 2025-01-17
make `head` accesible on lh backends

rickwierenga:
we often need to know which tips are mounted on which channels

for example:

- https://github.com/PyLabRobot/pylabrobot/pull/353
- https://github.com/PyLabRobot/pylabrobot/pull/357
- https://github.com/PyLabRobot/pylabrobot/pull/260

in #260, @BioCam added request_tip_len_on_channel. This has a side effect of moving all channels to z-safety as a way to determine tip lengths. With this pr, we can just query this information from the tips we know are on the channel head.


BioCam:
Why was the `STAR.request_tip_len_on_channel()` method removed?

Even if the tip tracker records the tip length accurately, a machine that can check its tip length at the firmware level is quite useful to troubleshoot potential issues with the PLR-level-implemented tip tracking.

rickwierenga:
brought it back with https://github.com/PyLabRobot/pylabrobot/pull/380

PR #366, opened 2025-01-19
[fix] prevent automatic channel spacing in ops with a single channel

naikymen:
Added a condition to the if clause that matches the logic described in the comment above it:

```
    # If the user specified a single resource, but multiple channels to use, we will assume they
    # want to space the channels evenly across the resource. Note that offsets are relative to the
    # center of the resource.
```

### DIFF
--- a/pylabrobot/liquid_handling/liquid_handler.py
+++ b/pylabrobot/liquid_handling/liquid_handler.py
@@ -1005,10 +1005,6 @@ class LiquidHandler(Resource, Machine):
       ValueError: If all channels are `None`.
     """
 
-    # If the user specified a single resource, but multiple channels to use, we will assume they
-    # want to space the channels evenly across the resource. Note that offsets are relative to the
-    # center of the resource.
-
     self._check_containers(resources)
 
     use_channels = use_channels or self._default_use_channels or list(range(len(resources)))
@@ -1028,7 +1024,7 @@ class LiquidHandler(Resource, Machine):
     # If the user specified a single resource, but multiple channels to use, we will assume they
     # want to space the channels evenly across the resource. Note that offsets are relative to the
     # center of the resource.
-    if len(set(resources)) == 1:
+    if len(set(resources)) == 1 & len(use_channels) > 1:
       resource = resources[0]
       resources = [resource] * len(use_channels)
       center_offsets = self._get_single_resource_liquid_op_offsets(
--- a/pylabrobot/liquid_handling/liquid_handler.py
+++ b/pylabrobot/liquid_handling/liquid_handler.py
@@ -841,7 +841,7 @@ class LiquidHandler(Resource, Machine):
     # If the user specified a single resource, but multiple channels to use, we will assume they
     # want to space the channels evenly across the resource. Note that offsets are relative to the
     # center of the resource.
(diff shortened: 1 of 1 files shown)


naikymen:
A couple comments:

- It would be nice to have tests for these cases.
- There are a pair of hard-coded values I'm not sure I understand.

```python
min_spacing_edge = (
  2  # minimum spacing between the edge of the container and the center of channel
)
min_spacing_between_channels = 9
```

Where do these come from? I can guess where the 9 comes from, but not the 2. Is this a hardware-specific thing? Perhaps it should be set by backends.

Thanks as always!

rickwierenga:
`_get_single_resource_liquid_op_offsets` should just return the center of the resource if only one channels is used. Aspirating from/dispensing to the center seems like desired behavior imo. The comment should probably be updated to reflect that

rickwierenga:
> It would be nice to have tests for these cases.

i will add some

> There are a pair of hard-coded values I'm not sure I understand.

You're right, these are quite STAR/Vantage-specific. I will make a pr making them backend attrs. thanks for flagging

naikymen:
Hi Rick! how are you doing?

> `_get_single_resource_liquid_op_offsets` should just return the center of the resource if only one channels is used. Aspirating from/dispensing to the center seems like desired behavior imo. The comment should probably be updated to reflect that

Hmm ok. The issue I was seeing actually comes from `_get_centers_with_margin`. It raises an error when a single channel is used and the resource has a small diameter (e.g. the diameter of a small tip).

Here is where the exception I saw comes from:

```python
def _get_centers_with_margin(dim_size: float, n: int, margin: float, min_spacing: float):
  """Get the centers of the channels with a minimum margin on the edges."""
  if dim_size < margin * 2 + (n - 1) * min_spacing:
    raise ValueError(f"Resource is too small to space {n} channels: {dim_size} mm")
```

The condition gives `2 * 2 + (0) * 9 = 4`. Any resource smaller than 4 will trigger the condition if one channel is used.

In my case I had a 3.3 mm resource, the width of a well in my 384 well-plates.

I'm not sure what to do now. Perhaps I should close this PR and open another one when its clearer.

Best and thanks as always!

naikymen:
Would this work?

```python
  if dim_size < margin * 2 + (n - 1) * min_spacing and n > 1:
    raise ValueError(f"Resource is too small to space {n} channels: {dim_size} mm")
```

Adding `and n > 1` would prevent the exception, but I'm not sure if it would break something else.

rickwierenga:
if the tip can't with with `margin` on each side the container is too small. the `n-1` term would obviously be 0 with n=1.

the problem is the margin seems to wide if you get an error while an operation is physically possible.

naikymen:
> the problem is the margin seems to wide if you get an error while an operation is physically possible.

The margin is too wide in this case yes. What do you suggest?

rickwierenga:
yes, i think it would be good to make it a static backend property as you suggested

rickwierenga:
or the tip! yesterday, i added the tip_bottom_diameter to HamiltonTip (not merged yet) which we could extend to other tips as well. https://github.com/PyLabRobot/pylabrobot/pull/353/commits/4aedcf10ffb2600d9895569259c7d901b1d5b603

PR #367, opened 2025-01-20
ThermoFisher Proflex thermocycler backend

btu6626:
- Implemented basic methods needed to run protocols and perform single step operations on thermofisher thermocyclers. 
- Tested on Proflex and SimpliAmp


ben-ray:
chadded implementation

PR #368, opened 2025-01-21
add init back into cytomat

ben-ray:
some cytomats are configured in the .exe service software to automatically init on startup, others are not. adding back init makes this more robust across different configs


rickwierenga:
undoing https://github.com/PyLabRobot/pylabrobot/pull/351 🙃

PR #370, opened 2025-01-23
Update using-the-visualizer.ipynb

bbbjiangggg:
Fixed the set_tip_state() commands in the user guide visualizer notebook for tip rack 2, 3, and 4. The boolean values should be stored in an one dimensional list, filling the rack column by  column. This fixed the visualization problem that I was encountering.


rickwierenga:
thanks!

PR #372, opened 2025-01-24
add retries to cytomat query overview register

ben-ray:
once every 5 or 6 tries, our cytomat 6000 & cytomat 2C will report it is "busy" for longer than an await function takes to complete. this change intends to fix this issue


rickwierenga:
done in https://github.com/PyLabRobot/pylabrobot/pull/373

PR #374, opened 2025-01-24
add retry logic to cytomat mv:st and mv:ts

ben-ray:
received `CytomatCommandUnknownError` on mv:st command after 49 successful calls, this should robustly fix random cytomat blackpills


rickwierenga:
hopefully addressed with https://github.com/PyLabRobot/pylabrobot/pull/385

PR #377, opened 2025-01-29
single container aspirate/dispense

rickwierenga:
fixing https://discuss.pylabrobot.org/t/correct-way-to-aspirate-dispense-to-from-a-1-well-reservoir-using-the-96-head/45/3


harley-pioneer:
Tested today with `await lh.aspirate96(pbs_res.get_item("A1"), 200, offset=Coordinate(0,0,2))` and was able to run aspirate96 and dispense96 command with STARlet

PR #383, opened 2025-02-01
fix resource rotation on move in star

ben-ray:


rickwierenga:
hopefully this fixes all iswap movements, especially when rotating resources, moving resources that are rotated themselves, moving resources of which a parent is rotated, and moving resources onto other rotated resources.

PR #387, opened 2025-02-03
ztouch probing uses request_tip_len_on_channel again

rickwierenga:
lh.backend.head sometimes is incorrect. Presumably because autoreload breaks the reference of lh.backend.head to the dictionary pointed to by lh (the value is copied and they point to different dictionaries after autoreload). We don't know for sure. Will be fixed in the future. For now, revert to the previous implementation that works.


rickwierenga:
following https://github.com/PyLabRobot/pylabrobot/pull/380

PR #421, opened 2025-03-09
fix rotation degrees on step off foil

ben-ray:
looks like we confused degrees with `% 90`


rickwierenga:
fixed with https://github.com/PyLabRobot/pylabrobot/pull/424. %90 would always be 0 for the supported rotations

PR #425, opened 2025-03-12
make hamilton heater shaker async

ben-ray:
tested with 8x hhs box


pylabrobot/heating_shaking/hamilton.py
```diff
@@ -51,19 +52,21 @@ def serialize(self) -> dict:
...
-    return self.io.read()
+  async def _send_command(self, command: str, **kwargs):
+      assert len(command) == 2, "Command must be 2 characters long"
+      
+      args = "".join([f"{key}{value}" for key, value in kwargs.items()])
+      await asyncio.to_thread(self.io.write, f"T{self.shaker_index}{command}id{str(self.command_id).zfill(4)}{args}".encode())
+      
+      self.command_id = (self.command_id + 1) % 10_000
+      response = await asyncio.to_thread(self.io.read)
+      return response
```
rickwierenga, line 63: actually i don't think the read or write themselves take so long and introducing multithreading leads to potential complications

i think it's better to do many small reads and asyncio.sleep in between

ben-ray: good idea!

ben-ray:
just found out this also breaks .decode on get_temperature

PR #428, opened 2025-03-15
“custom” offset option for `spread`

rickwierenga:
_applies only to single resource, multi channel operations_

previously it was just wide or tight

this allows custom offsets, all wrt center. with wide or tight, offsets are wrt those points

wide + tight background: https://github.com/PyLabRobot/pylabrobot/pull/378


ben-ray:
#441 eliminated my need for "custom" offsets argument, but this may still be useful for others

PR #441, opened 2025-03-19
adjustments to pipette positioning utils to support rotated plates

ben-ray:
this is tested to work fine on both 90deg rotated plates & non-rotated plates, but might move the exact location of the pips from before

### DIFF
--- a/pylabrobot/liquid_handling/utils.py
+++ b/pylabrobot/liquid_handling/utils.py
@@ -53,8 +53,6 @@ def get_wide_single_resource_liquid_op_offsets(
     x_offset = resource.get_absolute_size_x() / 2
     center_offsets = [Coordinate(x=x_offset, y=c, z=0) for c in centers]
     offsets_relative_to_center = [c - resource.center() for c in center_offsets]
-  print(center_offsets)
-  print(offsets_relative_to_center)
 
   # offsets are relative to the center of the resource, but above we computed them wrt lfb
   # so we need to subtract the center of the resource
--- a/pylabrobot/liquid_handling/utils.py
+++ b/pylabrobot/liquid_handling/utils.py
@@ -11,11 +11,14 @@ def _get_centers_with_margin(dim_size: float, n: int, margin: float, min_spacing
   if dim_size - (n - 1) * min_spacing <= min_spacing * 2:
     remaining_space = dim_size - (n - 1) * min_spacing - margin * 2
     return [margin + remaining_space / 2 + i * min_spacing for i in range(n)]
-  return [(i + 1) * dim_size / (n + 1) for i in range(n)]
+  start = margin
+  end = dim_size - margin
+  step = (end - start) / (n - 1) if n > 1 else 0
+  return [start + i * step for i in range(n)]
 
 
 def get_wide_single_resource_liquid_op_offsets(
-  resource: Resource, num_channels: int
+  resource: Resource, num_channels: int,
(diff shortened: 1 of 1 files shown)


pylabrobot/liquid_handling/utils.py
```diff
@@ -66,14 +71,14 @@ def get_tight_single_resource_liquid_op_offsets(
...
+    min_y = (resource.get_absolute_size_y() - channel_space) / 2
     if min_y < min_spacing_edge:
       raise ValueError("Resource is too small to space channels.")
     offsets = [
       Coordinate(0, min_y + i * min_spacing_between_channels, 0) for i in range(num_channels)
     ][::-1]
   elif resource.get_absolute_rotation().z % 90 == 0:
-    min_x = (resource.get_size_x() - channel_space) / 2
+    min_x = (resource.get_absolute_size_x() - channel_space) / 2
     offsets = [
```
rickwierenga, line 82: does it work with just `resource.get_absolute_size_y`? why is the special %90 case necessary?

pylabrobot/liquid_handling/utils.py
```diff
@@ -45,13 +48,15 @@ def get_wide_single_resource_liquid_op_offsets(
...
-    y_offset = resource.get_size_y() / 2
-    center_offsets = [Coordinate(x=c, y=y_offset, z=0) for c in centers]
+    x_offset = resource.get_absolute_size_x() / 2
+    center_offsets = [Coordinate(x=x_offset, y=c, z=0) for c in centers]
+    offsets_relative_to_center = [c - resource.center() for c in center_offsets]
 
   # offsets are relative to the center of the resource, but above we computed them wrt lfb
   # so we need to subtract the center of the resource
-  return [c - resource.center() for c in center_offsets]
+  return offsets_relative_to_center
```
rickwierenga, line 59: effectively nothing changed here

pylabrobot/liquid_handling/utils.py
```diff
@@ -11,11 +11,14 @@ def _get_centers_with_margin(dim_size: float, n: int, margin: float, min_spacing
   if dim_size - (n - 1) * min_spacing <= min_spacing * 2:
     remaining_space = dim_size - (n - 1) * min_spacing - margin * 2
     return [margin + remaining_space / 2 + i * min_spacing for i in range(n)]
-  return [(i + 1) * dim_size / (n + 1) for i in range(n)]
+  start = margin
+  end = dim_size - margin
+  step = (end - start) / (n - 1) if n > 1 else 0
+  return [start + i * step for i in range(n)]
```
rickwierenga, line 17: since there is enough space in this case, we would rather space the channels equally than putting side channels exactly at the margin. margin is meant to be a minimum margin

ben-ray: because I can define margin, I expect channels to go narrower when I input a higher margin

pylabrobot/liquid_handling/utils.py
ben-ray: correct, we no longer need the %90 case when using get_absolute_size_x

pylabrobot/liquid_handling/utils.py
ben-ray: I can instead change the well definition for the same effect, but then the well definition is no longer accurate

rickwierenga: they should go narrower than margin in any case

PR #442, opened 2025-03-21
Set objective and imaging mode on cytation1

rickwierenga:


rickwierenga:
https://discuss.pylabrobot.org/t/cytation-1-objective-not-moving-with-y-p0e01-instruction/187/

PR #444, opened 2025-03-21
cytation 1 P0d instead of P0e for objective moving

vcjdeboer:
The Cytation 1 apparently uses P0d instead of P0e for moving the objectives


rickwierenga:
thanks!

PR #453, opened 2025-03-27
cytation 1 image size and overlap

vcjdeboer:


pylabrobot/plate_reading/biotek_backend.py
```diff
@@ -1190,8 +1202,14 @@ def image_size(magnification: float) -> Tuple[float, float]:
     if center_position is None:
       center_position = (0, 0)
     # Going in a snake pattern is not faster (strangely)
+
+    #overlap between images to have images stitched
+    overlap_fraction = 0.1  # 10% overlap
+    step_width = img_width * (1 - overlap_fraction)
+    step_height = img_height * (1 - overlap_fraction)
+
```
rickwierenga, line 1210: should we make this a parameter?

pylabrobot/plate_reading/biotek_backend.py
```diff
@@ -1164,13 +1164,25 @@ def image_size(magnification: float) -> Tuple[float, float]:
...
+        if magnification == 40:
+            return (216.5 / 1000, 161.9 /1000)  # pixel size ≈ 0.168 µm
+        raise ValueError(f"Don't know image size for magnification {magnification}")
+      else:
+        if magnification == 4:
+          return (3474 / 1000, 3474 / 1000)
+        if magnification == 20:
+          return (694 / 1000, 694 / 1000)
+        if magnification == 40:
+          return (347 / 1000, 347 / 1000)
```
rickwierenga, line 1184: do you know this is because of cytation 1? it seems to me tis would be determined by the sensor (since it's possible to share sensors between cytation1 and 5 & why would they record different sizes on a different machine?)

pylabrobot/plate_reading/biotek_backend.py
vcjdeboer: yes that is possible, maybe the stitching needs to be implemented then as well

rickwierenga: definitely, but i think that should be a separate PR imo

pylabrobot/plate_reading/biotek_backend.py
rickwierenga: also `else` after return/raise is not necessary :)

PR #454, opened 2025-03-31
ensure shaking before return

rickwierenga:
apparently sometimes it fails to start


ben-ray:
I noticed this will prevent the hhs from updating its shake speed if this cmd is called while the shaker is already shaking.

PR #456, opened 2025-04-11
Hamilton Mfx module 188049

linx5o:
Following the discussion about not yet implemented modules. [PLR Forum](https://discuss.pylabrobot.org/t/starlet-carrier-module-classifications/190/2).

I have attempted to create a definition for a module I have on hand.

There are no cat. no on the module. I also do not have access to the purchase history. 

From the `VENUS four Operators Manual.pdf` provided by Eric Sindelar (senior Hamilton employee) on the [labautomation.io](http://labautomation.io/) forum. There are two potential candidates.
image
image

This [ebay listing](https://www.ebay.com/itm/235714969973) suggests it is the prior. However, this other listing for [188042](https://www.ebay.com/itm/365462833597) more aligns with the image in the manual.
image

I therefore believe that this module should be 188049.

I have tested this module on the liquid handler. There are no errors and the height is correct upon visual inspection when using plates.

### DIFF
--- a/pylabrobot/resources/hamilton/mfx_modules.py
+++ b/pylabrobot/resources/hamilton/mfx_modules.py
@@ -63,7 +63,7 @@ def MFX_DWP_module_flat(name: str) -> PlateHolder:
   )
 
 def MFX_WP_module_flat(name: str) -> PlateHolder:
-  """Hamilton cat. no.: 188041 (uncertain)
+  """Hamilton cat. no.: 188049 (uncertain)
   Module to position a Well Plate. Taller version of MFX_DWP_module_flat.
   """
   return PlateHolder(
--- a/pylabrobot/resources/hamilton/mfx_modules.py
+++ b/pylabrobot/resources/hamilton/mfx_modules.py
@@ -64,12 +64,13 @@ def MFX_DWP_module_flat(name: str) -> PlateHolder:
 
 def MFX_WP_module_flat(name: str) -> PlateHolder:
   """Hamilton cat. no.: 188041 (uncertain)
+  Module to position a Well Plate. Taller version of MFX_DWP_module_flat.
   """
   return PlateHolder(
     name=name,
     size_x=135.0,
     size_y=94.0,
-    size_z=216 - 18.195 - 100,
+    size_z=216 - 18.195 - 100, # mesured using probe
     # probe height - carrier_height - deck_height
     child_location=Coordinate(3.0, 4.0, 209.9 - 18.195 - 100),
     pedestal_size_z=0,
--- a/pylabrobot/resources/hamilton/mfx_modules.py
+++ b/pylabrobot/resources/hamilton/mfx_modules.py
@@ -61,3 +61,16 @@ def MFX_DWP_module_flat(name: str) -> PlateHolder:
     model=MFX_DWP_rackbased_module.__name__,
     pedestal_size_z=0,
   )
+
+def MFX_WP_module_flat(name: str) -> PlateHolder:
+  """Hamilton cat. no.: 188041 (uncertain)
+  """
+  return PlateHolder(
+    name=name,
+    size_x=135.0,
+    size_y=94.0,
+    size_z=216 - 18.195 - 100,
+    # probe height - carrier_height - deck_height
+    child_location=Coordinate(3.0, 4.0, 209.9 - 18.195 - 100),
+    pedestal_size_z=0,
+  )


BioCam:
Hi @linx5o,

> I therefore believe that this module should be 188049.

I don't believe it is `188049` because...

image

...cat no `188049` seems to refer to a plate_holder without clamps or lips that has a plate_adapter on top of it?

BioCam:
Would you be okay with posting a picture of this multiflex module please?

Then we can assess:
- Does it have metal clamps?
- You've named it "flat", does it still have a pedestal (i.e. central elevation)? It _is possible_ to have a central elevation and it is still technically "flat" because the corners might also be raised: e.g. this is can be kinda defined as "flat":
shared image (8)

- Or is the plate_holder top actually completely flat?
- And most importantly, does it have these small nodules on the side of the pedestal (if there is a pedestal)?
image

linx5o:
Here are the images:
IMG20250416154846
IMG20250416154837
No clamps, just some guides. No possibility for central elevation either.
image
Some rough measurements of size. It is the taller module.

I do agree with your judgment. I have not been able to get LabwareAssistant3D.exe to work on my computer. There does not seem to be a lot of information online about these either. Sorry! I will keep trying to find a cat. no.

rickwierenga:
found by @ben-ray:

image

BioCam:
No worries @linx5o,

The pictures help a lot and we can figure this out :)

---

> found by @ben-ray:
> 
> image

...I don't think it is this one because @linx5o didn't mention that his MFX module is _sloped_, i.e. has a rigid top surface with a tilt?

---

## Investigation 🕵🏻‍♂️ 

This PR introduces the following resource definition function:

```python
from pylabrobot.resources import PlateHolder, Coordinate
    MFX_CAR_L5_base, 
)

def MFX_WP_module_flat(name: str) -> PlateHolder:
  """Hamilton cat. no.: 188049 (uncertain)
  Module to position a Well Plate. Taller version of MFX_DWP_module_flat.
  """

  # Resource size is carried over from MFX_DWP_module_flat.
  # Child location offset for x, y are measured with caliper.

  return PlateHolder(
    name=name,
    size_x=135.0,
    size_y=94.0,
    size_z=216 - 18.195 - 100,  # mesured using probe
    # probe height - carrier_height - deck_height
    child_location=Coordinate(3.0, 4.0, 209.9 - 18.195 - 100),
    pedestal_size_z=0,
  )

enruis_mfx_module = MFX_WP_module_flat(name="enruis_mfx_module")

mfx_carrier_test = MFX_CAR_L5_base(
  name="mfx_carrier_test",
  modules={
      0: enruis_mfx_module,
  }
)
lh.deck.assign_child_resource(mfx_carrier_test, rails=1)

print(enruis_mfx_module.get_absolute_location() + enruis_mfx_module.child_location)

>>>Coordinate(x=103.0, y=72.0, z=209.9)
# i.e. the origin (flb) of the plate_holder's child_location
```

Going through Labware Assistant 3D this seems like a good candidate:
image
(don't get thrown off by Labware Assistant 3D showing a metal top, I ordered sth I believed should have a metal top based on this .exe before and it came with a coated plastic top like your MFX module 🙃 )

**How to check for the difference?**

VENUS 4 is free to use and also shared publicly shared by Eric Sindelar on the labautomation.io forum.

One can use Labware Assistant 3D to generate an MFX_CAR_L5_base with `188228` in its front position, add it to a STARlet deck and check out the child_location like this:
image


--> difference between definitions for the plate_holder's child_location:
- @linx5o: `Coordinate(x=103.0, y=72.0, z=209.9)`
- VENUS, assuming cat no.: `188228`: `Coordinate(x=104.0, y=71.5, z=206.0)`

Now, this doesn't have to mean VENUS is right, in many instances we found that VENUS definitions are more of a guide than an actual "ground truth", but usually that applies more to VENUS' plate/tube/etc definitions, not so much towards fine-machined metal carriers + mfx_modules.

@linx5o, do you think this adjusted definition could work for your purposes / test it?:

```python
def MFX_WP_module_flat(name: str) -> PlateHolder:
  """Hamilton cat. no.: 188228
  Module to position a Well Plate. Taller version of MFX_DWP_module_flat.
  """

  # Resource size is carried over from MFX_DWP_module_flat.
  # Child location offset for x, y are measured with caliper.

  return PlateHolder(
    name=name,
    size_x=135.0,
    size_y=94.0,
    size_z=216 - 18.195 - 100,  # mesured using probe
    # probe height - carrier_height - deck_height
    child_location=Coordinate(4.0, 3.5, 206.0 - 18.195 - 100),
    pedestal_size_z=0,
  )
```

I believe your definition is close enough to the VENUS "ground truth" to make your definition refer to Hamilton cat. no. `188228` :)

BioCam:
Note:

Might I recommend renaming `MFX_WP_module_flat()` -> `Hamilton_MFXmodule_tall_flat()`

(We haven't figured out a PLR naming convention for MFX Modules yet, but the first identifier should always be the manufacturer. So far we tried to avoid ambiguous terms like MTP [micro-tube-plate], DWP [deep-well-plate], WP [well-plate?]. Though a couple are still remnant from old times. I'm making this recommendation here in anticipation of what the naming convention might look like)

linx5o:
@BioCam Could sloped be referring to the positioning guides along the side (The things that stick up to help guide the plate)? I can't really tell with 188228, but they don't seem to have a slope in them.

I also remeasured the sizes. My x for off. It should be 134.

BioCam:
> @BioCam Could sloped be referring to the positioning guides along the side (The things that stick up to help guide the plate)? I can't really tell with 188228, but they don't seem to have a slope in them.

No, I believe you are referring to what some people call plate_holder "lips", this stuff:
image

"Sloped" refers to the fact that the top surface is higher on the right than on the left, usually by ~2 mm.
This gives every well a tiny tilt, helping with liquid recovery.
You can achieve the same thing by buying a "tilter mfx module" (which Rick, Ben & the Retro team did a lot of PLR work on).
But a tilter module costs a couple thousands, a sloped mfx plate holder cost max a couple hundreds.




> @BioCam Could sloped be referring to the positioning guides along the side (The things that stick up to help guide the plate)? I can't really tell with 188228, but they don't seem to have a slope in them.
> 
> I also remeasured the sizes. My x for off. It should be 134.
> 
> I have used this definition for the last few days. No issues so far.

I don't understand what you mean by this. Could you please elaborate.
Your definition generates plate_holder's child_location a `Coordinate(x=103.0, y=72.0, z=209.9)` based on the code I provided above (i.e. MFX carrier on rail=1, your MFX module in the frontmost position on that MFX carrier).



> I have used this definition for the last few days. No issues so far.

There are many reasons why these slight differences between VENUS' "ground truth" might not have caused any issues:

linx5o:
@BioCam 
> No, I believe you are referring to what some people call plate_holder "lips"
> "Sloped" refers to the fact that the top surface is higher on the right than on the left, usually by ~2 mm. 

I see. Thank you for clarifying. I'm not particularly familiar with these naming schemes.


> I don't understand what you mean by this. Could you please elaborate?

I have change the size of the x axis in the definition:
`return PlateHolder(
    name=name,
    size_x=134.0,
    size_y=94.0,
    size_z=216 - 18.195 - 100,`

Have not tested yet.

With regards to iSwap. I only have these modules defined. I don't have any generic plate holder. So any movements with iSwap wouldn't really show the effects of incorrect dimensions. They move fine between two instances of this module. I do need to use the pickup from top parameter to have it pickup lower down from the top.

I will try to do some more tests.

BioCam:
> `return PlateHolder( name=name, size_x=134.0, size_y=94.0, size_z=216 - 18.195 - 100,`

I see, so you didn't change the `child_location` but instead changed the `PlateHolder`'s `size_x=135.0` -> `size_x=134.0`

I recommend you not do that:
The carrier you are using is `MFX_CAR_L5_base` which has a `size_x=135.0` mm, which is...
1. the VENUS "ground truth"
2. confirms the fact that this carrier (like most carriers) is 6T i.e. "6 tracks"; each track has a confirmed `size_x=22.5` -> 6 tracks * 22.5 mm/track = 135

Almost all MFX modules have the same `size_x` & `size_y` (based on their need to use the screw holes available for its fixtures.
Most MFX modules are completely flush with the base MFX carrier on each side of `x`.
Therefore, the MFX modules' `size_x` is the same as the MFX carrier's `size_x` `=135`mm.

If you use 134 mm you will see that the MFX Module doesn't sit flush on the carrier when using PLR's Visualizer.

But, tbf... the MFX Module's `size_x` doesn't _actually_ matter - we are only discussing it here because PLR aims to model physical reality as accurately and consistent as possible.

**Why does it not matter?**

Because PLR uses a front-left-bottom (flb) parent-child relationship.
This means removing 1 mm of the MFX module `size_x` removes it from the **right**, which has no influence on the model's functionality.

---

What does matter a lot is the `child_location`, for which VENUS says you should be using...
```python
def Hamilton_MFXmodule_tall_flat(name: str) -> PlateHolder:
  """Hamilton cat. no.: 188228
  Module to position a short / 'MTP' wellplate, i.e. wellplate size_z < 39 mm before
  breaching minimal safety height of 245 mm.
  Taller MFX module version of MFX_DWP_module_flat.
  """

  # Resource size is carried over from MFX_DWP_module_flat.
  # Child location offset for x, y are measured with caliper.

  return PlateHolder(
    name=name,
    size_x=135.0,
    size_y=94.0,
    size_z=216 - 18.195 - 100,  # Did you measure this using clld probing?
    # probe height - carrier_height - deck_height
    child_location=Coordinate(4.0, 3.5, 206.0 - 18.195 - 100), <== CHANGES MADE HERE
    pedestal_size_z=0,
  )
```

linx5o:
I see. I measure the height of the lips and pedestal with clld probing. For x, y offsets. I measured the tabs with a caliper.

BioCam:
Hi @linx5o, just checking in whether you have been able to verify the above plate_holder / "MFX Module" definition so we can update/merge this PR? :)

rickwierenga:
any update here @linx5o ?

would be great to confirm exact geometry and merge this PR soon!

linx5o:
@rickwierenga 
I am really sorry, no updates. I don't have any resources to confirm the modules I have. Best I can do is take measurements. Sorry!

rickwierenga:
>  I don't have any resources to confirm the modules I have. 

what does this mean? do you not have access to the labware anymore?

rickwierenga:
please reopen this in the future

PR #471, opened 2025-04-18
make hhs truly async

ben-ray:


jrast:
Nice!

BTW: I observed that many classes are not truly async and it would be nice to add support for true async to more classes. Looking at the code it would be nice to move the relevant code into the io layer.

ben-ray:
IO layer sounds like a good idea! the firmware logs for async devices have nondeterministic ordering of commands, which currently breaks our CI (that checks to ensure all firmware commands are identical for our production protocols we don't want changing). the ideal async solution at IO layer could help resolve our CI issue as well

rickwierenga:
should be solved with https://github.com/PyLabRobot/pylabrobot/pull/479

PR #474, opened 2025-04-30
make io methods async

rickwierenga:
while not all of the methods here are truly async, some operations are actually fundamentally synchronous, it is possible for some operations (like in http) to be asynchronous (eg https://docs.aiohttp.org/en/stable/). PLR is fundamentally an async framework. so, it makes sense to have the io abc be async. this forces all implementations to also have async methods.

to make it easy to write async backends, i agree that implementing some async-ification directly in the io layer is very convenient (as suggested by @jrast here: https://github.com/PyLabRobot/pylabrobot/pull/471#issuecomment-2816757239). where applicable, we could run a background thread that e.g. reads until a certain condition is met (e.g. new line character is read, a certain number of bytes, etc.). all backends should be async, and this makes that easy. the alternative is like each  usb-backend async-ifies usb itself.

we already have some intelligence in the io layer. For example, `USB.read` calls the pyusb read method continuously as long as we read the max packet size (so that only one `USB.read` call is required from the PLR-backend).

### DIFF
--- a/pylabrobot/centrifuge/vspin.py
+++ b/pylabrobot/centrifuge/vspin.py
@@ -30,7 +30,7 @@ class Access2Backend(LoaderBackend):
     r = None
     start = time.time()
     while r != b"" or x == b"":
-      r = self.io.read(1)
+      r = await self.io.read(1)
       x += r
       if r == b"":
         await asyncio.sleep(0.1)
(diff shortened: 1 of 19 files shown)


rickwierenga:
the actual async-ification of specific backends will be done in future PRs. This PR is just to introduce the async io abc and make the necessary follow-up changes.

PR #485, opened 2025-05-02
add legacy Heraeus Cytomat backend, which may be identical to liconic backend

ben-ray:


pylabrobot/incubators/cytomat/cytomat_legacy_backend.py
```diff
@@ -0,0 +1,189 @@
...
+  async def _send_command(self, command: str) -> str:
+    """
+    Send an ASCII command (without CR) and return the raw response string.
+    """
+    print(f"send: {command}")
+    cmd = command.strip() + "\r"
+    logger.debug("Sending Cytomat command: %r", cmd)
+    self.io.write(cmd.encode(self.serial_message_encoding))
+    resp = self.io.read(128).decode(self.serial_message_encoding)
+    if not resp:
```
rickwierenga, line 160: io is async so needs await

PR #495, opened 2025-05-10
New vilhelm evo branch

rickwierenga:
cherry picked commits from https://github.com/PyLabRobot/pylabrobot/pull/472 onto plr:main


pylabrobot/liquid_handling/backends/tecan/EVO_backend.py
```diff
@@ -1100,15 +1224,291 @@ async def get_disposable_tip(self, tips, z_start, z_search):
...
+  pass
+  # """
+  # Provides firmware commands for the MCA TEST
+  # """
+
+  # async def initialize_plunger(self, tips):
+  #   """Initializes plunger and valve drive
+
+  #   Args:
+  #     tips: binary coded tip select
```
rickwierenga, line 1269: should this code be un-commented?

pylabrobot/liquid_handling/backends/tecan/EVO_backend.py
```diff
@@ -172,15 +184,15 @@ class EVOBackend(TecanLiquidHandler):
...
   def __init__(
     self,
     diti_count: int = 0,
-    packet_read_timeout: int = 120,
-    read_timeout: int = 300,
-    write_timeout: int = 300,
+    packet_read_timeout: int = 12,  # VIKMOL CHANGED FROM 120 THIS CHANGE INCREASES INITIALIZATION SPEED
+    read_timeout: int = 60,  # VIKMOL CHANGED FROM 300 THIS CHANGE INCREASES INITIALIZATION SPEED
+    write_timeout: int = 60,  # VIKMOL CHANGED FROM 300 THIS CHANGE INCREASES INITIALIZATION SPEED
   ):
```
rickwierenga, line 196: @VilhelmKMoller  why? it is just the maximum time that an operation can take. if it speeds it up, that means we are getting time out errors somewhere. (if that is intended, for example to check whether a module exists, we should be using a really short time out)

pylabrobot/liquid_handling/backends/tecan/EVO_backend.py
```diff
@@ -485,22 +566,64 @@ async def drop_tips(self, ops: List[Drop], use_channels: List[int]):
...
+
     await self.liha.set_z_travel_height([self._z_range] * self.num_channels)
     await self.liha.position_absolute_all_axis(
       x, int(y - ys * 3.5), ys, [self._z_range] * self.num_channels
     )
 
-    # discard tips
-    await self.liha.discard_disposable_tip(self._bin_use_channels(use_channels))
+    # place tips back in rack
+    await self.liha.place_tip_back(self._bin_use_channels(use_channels))
```
rickwierenga, line 626: this is a front end method on LH. I don't think a specific backend method is needed, `drop_tips` should support it

pylabrobot/resources/tecan/plate_carriers.py
```diff
@@ -118,7 +118,7 @@ def MP_3Pos_PCR(name: str) -> TecanPlateCarrier:
       resource_size_x=127.0,
       resource_size_y=85.5,
       name_prefix=name,
-      pedestal_size_z=0,  # ?
+      pedestal_size_z=0,
     ),
```
rickwierenga, line 122: is it actually 0 for all of them?

pylabrobot/liquid_handling/liquid_classes/tecan - 副本.py
rickwierenga, line 1: how does this file compare to just `tecan.py`? which should we use?

rickwierenga:
`place_tip_back` and `discard_disposable_tip` call the same firmware command (`"AST"`). Otherwise the implementations for `drop_tips` and `return_tips` are similar. can they be merged into one function?

rickwierenga:
the offsets now get computed for `z_positions`, but `z_positions` is not used for `pick_up_tips` and `drop_tips`. it is always `self._z_range`. is that correct? should we test with different positions?

rickwierenga:
it would be great to get the commands `pick_up_tips96` etc. working now that you have implemented the MCA. perhaps a future PR

pylabrobot/liquid_handling/backends/tecan/EVO_backend.py
```diff
@@ -1100,15 +1224,291 @@ async def get_disposable_tip(self, tips, z_start, z_search):
...
+  async def place_tip_back(self, tips, discard_hight):  # added by VIKMOL z_start, z_search
+    """Places tips back into the original tip rack positions.
+
+    Args:
+      tips: Binary coded tip select
+      discard_hight: binary. 0 above tip rack, 1 in tip rack
+
+    """
+
+    await self.backend.send_command(module=self.module, command="AST", params=[tips, discard_hight])
```
rickwierenga, line 1255: `return_tips` calls this without the `discard_hight`. perhaps `return_tips` is not used anymore?

pylabrobot/liquid_handling/backends/tecan/EVO_backend.py
VilhelmKMoller: This is probably not the best solution.
When you initialize the EVO all arms move sequentially to the right of the robot if the operater as placed the arms to the far right then this code will cause an error. If that happens you can run the code again and then it will be fine. 
I changed the number as it changes the initialization time from 5min to 30sec-2min.

pylabrobot/resources/tecan/plate_carriers.py
VilhelmKMoller: yes it is. I have not found any where we need to add a pedestal height yet.

pylabrobot/liquid_handling/liquid_classes/tecan - 副本.py
VilhelmKMoller: This is my students trying to change the liquid classes for testing purposes. This has been added by mistake. The file can be deleted for now.

VilhelmKMoller:
> `place_tip_back` and `discard_disposable_tip` call the same firmware command (`"AST"`). Otherwise the implementations for `drop_tips` and `return_tips` are similar. can they be merged into one function?

Wait with merging or removing. The return tips to tip rack function works but discard tips in trash does not. The reason is that i need to make a trash container before i can through out the tips. I would like your advice on that. When that is done then we can change the discard tips command to the correct discard firmware command.

pylabrobot/liquid_handling/backends/tecan/EVO_backend.py
VilhelmKMoller: I was unsure what the difference between the LH.drop_tips, LH.return_tips and LH.discard_tips. 
I assumed that drop_tips contained the two other options where the others where explicit. So drop_tips could be used to discard tips or return tips. 
Is this correct?

pylabrobot/liquid_handling/backends/tecan/EVO_backend.py
rickwierenga: call:
- this copy pasted from liha
- delete and implement in the feature

pylabrobot/liquid_handling/backends/tecan/EVO_backend.py
rickwierenga: call:
- bug: it seems to wait for timeout to pass before it's done. fix
- if you decrease it too much, it will raise a timeout error

in evoware: initialization flag, check if it's already initialized.

pylabrobot/liquid_handling/backends/tecan/EVO_backend.py
```diff
@@ -485,22 +545,63 @@ async def drop_tips(self, ops: List[Drop], use_channels: List[int]):
...
+    # Get positions including offsets
+    x_positions, y_positions, z_positions = self._liha_positions(ops, use_channels)
+
+    # Apply offsets
+    for i, op in enumerate(ops):
+      x_positions[i] += op.offset.x
+      y_positions[i] += op.offset.y
+
+    for key in z_positions:
+      z_positions[key][i] += op.offset.z  # Apply the offset to all z position types
```
rickwierenga, line 561: can be removed

pylabrobot/liquid_handling/backends/tecan/EVO_backend.py
```diff
@@ -485,22 +545,63 @@ async def drop_tips(self, ops: List[Drop], use_channels: List[int]):
...
+    await self.liha.set_z_travel_height([self._z_range] * self.num_channels)
+    await self.liha.position_absolute_all_axis(
+      x,
+      y - yi * ys,
+      ys,
+      [self._z_range] * self.num_channels,
+    )
+
+    # TODO check channel positions match resource positions for z-axis
+    await self.liha.discard_disposable_tip(self._bin_use_channels(use_channels), discard_hight=0)
```
rickwierenga, line 577: binary:
- discard_hight = 0: into the box
- discard_hight = 1: above the box

@VilhelmKMoller will test

maybe in the future:

```
if resource is trash:
  above the box
else:
  in the box
```

pylabrobot/liquid_handling/backends/tecan/EVO_backend.py
rickwierenga: https://github.com/PyLabRobot/pylabrobot/issues/498

pylabrobot/liquid_handling/backends/tecan/EVO_backend.py
rickwierenga: LH.drop_tips, LH.return_tips and LH.discard_tips: all of them call LH.drop_tips, which is the general method to drop it. LH.drop_tips is calling the backend method LHB.drop_tips. return_tips means putting the tips back where they were picked up; discard_tips put them in the default trash location. Every tip drop operation is basically the same, so it makes sense to have just one method on the backend. In this way the backends can be as simple as possible, easy to write new backends.

rickwierenga:
the last issue now is that the tests for EVO.pick_up_tips don't pass:

```
E               AssertionError: Calls not found.
E               Expected: [call(module='C5', command='SHZ', params=[2000, 2000, 2000, 2000, 2000, 2000, 2000, 2000]),
E                call(module='C5', command='RPX', params=[0]),
E                call(module='C5', command='PAA', params=[2380, 1991, 90, 2000, 2000, 2000, 2000, 2000, 2000, 2000, 2000]),
E                call(module='C5', command='PVL', params=[0, None, None, None, None, None, None, None]),
E                call(module='C5', command='SEP', params=[420, None, None, None, None, None, None, None]),
E                call(module='C5', command='PPR', params=[30, None, None, None, None, None, None, None]),
E                call(module='C5', command='AGT', params=[1, 768, 210, 0])]
E               Actual: [call(module='C5', command='SHZ', params=[2000, 2000, 2000, 2000, 2000, 2000, 2000, 2000]),
E                call(module='C5', command='RPX', params=[0]),
E                call(module='C5', command='PAA', params=[2380, 1991, 90, 2000, 2000, 2000, 2000, 2000, 2000, 2000, 2000]),
E                call(module='C5', command='PVL', params=[0, None, None, None, None, None, None, None]),
E                call(module='C5', command='SEP', params=[420, None, None, None, None, None, None, None]),
E                call(module='C5', command='PPR', params=[30, None, None, None, None, None, None, None]),
E                call(module='C5', command='AGT', params=[1, 1023, 210, 0])]
```

the second parameter to C5AGT is differnt: 768 -> 1023.

there was a comment `# TODO: check z params` in the original method, and i'm pretty sure the evo on which this backend was initially developed had fixed tips so this method may never have been tested.

with this in mind, i am updating the tests and trusting this new implementation is tested and correct

PR #500, opened 2025-05-15
await loop.run_in_executor

ben-ray:


rickwierenga:
oversight, thanks for fixing

Issue #477, opened 2025-04-30
make io.FTDI async


rickwierenga:
https://github.com/PyLabRobot/pylabrobot/pull/474

PR #504, opened 2025-05-19
make io.FTDI async

rickwierenga:
https://github.com/PyLabRobot/pylabrobot/issues/477

### DIFF
--- a/pylabrobot/centrifuge/vspin_backend.py
+++ b/pylabrobot/centrifuge/vspin_backend.py
@@ -47,7 +47,7 @@ class Access2Backend(LoaderBackend):
     logger.debug("[loader] setup")
 
     await self.io.setup()
-    self.io.set_baudrate(115384)
+    await self.io.set_baudrate(115384)
 
     status = await self.get_status()
     if not status.startswith(bytes.fromhex("1105")):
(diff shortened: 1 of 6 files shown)


rickwierenga:
similar to USB, HID and Serial, the FTDI library we use (called pylibftdi) has synchronous apis. This means that when calling the read/write/etc. methods, we actually block. This means it is not possible to send multiple commands to multiple machines at the same time. For example: pick up a plate using the iswap while opening the incubator door. If you wait for a response to the first before sending the second command, they will happen in serial.

in https://github.com/PyLabRobot/pylabrobot/pull/474 i made all io methods async. but this is essentially just a python-level syntax change. it does not mean the methods will suddenly run in the background. to actually do things simultaneously, we have to run the code elsewhere (or, for example, have `asyncio.sleep` which allows other async methods to run - different from `time.sleep` which actively blocks)

in these PRs (this, https://github.com/PyLabRobot/pylabrobot/pull/479, https://github.com/PyLabRobot/pylabrobot/pull/505) i am running the read/write/etc. methods in a separate thread, allowing the main thread to do other things (like start new threads for other devices). it's implemented using a `ThreadPoolExecutor` so that we are able to use the same thread for each backend (in between its `setup` and `stop` calls). The alternative could be spawning new threads for each method, which would be slower.

PR #516, opened 2025-05-22
Add `itemized_resource.traverse(direction="down_left")`

BioCam:
missing option in `itemized_resource.traverse(direction="down_left")`

+ renamed accordingly
`itemized_resource.traverse(direction="down")` 

-> `itemized_resource.traverse(direction="down_right")`

### DIFF
--- a/pylabrobot/resources/itemized_resource.py
+++ b/pylabrobot/resources/itemized_resource.py
@@ -332,6 +332,9 @@ class ItemizedResource(Resource, Generic[T], metaclass=ABCMeta):
       indices = [(8 * y + x) for x in range(8) for y in range(0, 12)]
     elif direction == "left":
       # Start at the top right, and go left in each row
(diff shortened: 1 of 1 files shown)


rickwierenga:
wouldn't you want `left_up` and the other ones as well?

- {up, down} x {left, right}
- {left, right} x {down up}

should be 8 variants

BioCam:
> wouldn't you want `left_up` and the other ones as well?

For completeness-sake maybe, but I couldn't image a case in which the others would be useful.

I added "down_left" because I have a use case right now and am missing this option :)

rickwierenga:
@BioCam says: we should split into
- start_position, e.g. left
- direction, e.g. down

raise error when impossible

rickwierenga:
i extended to support all directions and starting locations. i rewrote the entire method to make this easier. it also fixed the situation where this method was actually hardcoded to 96-item resources before.

PR #521, opened 2025-05-24
OT2: enable manual channel selection & auto-cancel run in stop()

hazlamshamin:
- Remove use_channels assertions from pick_up, aspirate, dispense, drop
- Enhance stop() to cancel any live run via HTTP-API


rickwierenga:
(sorry for late response) thanks for flagging + PR!

PR #522, opened 2025-05-24
Integrate Optional Barcode Information Into `Resource`

BioCam:
### Hi everyone,

Many automation applications require barcodes but at the moment PLR does not have a simple way of storing barcode information.
This PR proposes adding three new attributes to the `Resource` class:

```python
    barcode: Optional[Union[str, int]] = None,
    barcode_symbology: Optional[str] = None,
    barcode_position_on_resource: Optional[
      Literal["right", "back", "left", "front", "bottom", "top"]
```
---

#### Classic Usage: Resource Tracking
Barcodes are essential for tracking resources in automation workflows, especially for items like plates and tubes. For example:

- **Diagnostics**: Samples may arrive with barcoded tubes or well plates, and the system needs to verify them.
- **Tracking Across Runs**: Barcodes allow resources to be tracked across different runs, such as identifying a plate processed on day 0, moved to an incubator, and later retrieved for day 1 processing.

#### Classic Usage: Verification
Barcodes help verify that resources are correctly placed. For example, a barcode scanner on a Hamilton STAR can verify correct tip rack placement and alert the operator if something is wrong (see [PR#64: Exposing autoload carrier load, unload & barcode reading](https://github.com/PyLabRobot/pylabrobot/pull/64): .

---

While these are classic use cases, there are additional challenges in barcode reading:

- Misaligned self-printed barcodes
- Barcode symbology not supported by the hardware
- Issues with the combination of plates, plate holders, and the specific height at which the scanner can read barcodes

Python can address these problems by using cameras and barcode libraries for more flexible, custom solutions.

### The proposed attributes in this PR will allow:

- **Dynamic Barcode Updates**: The `barcode` can be updated at runtime based on measurements or spreadsheet inputs.
- **Barcode Symbology**: The `barcode_symbology` stores the barcode's encoding scheme, which is crucial for compatibility with scanners and for custom search implementations.
- **Barcode Position**: The `barcode_position_on_resource` defines the face of the resource where the barcode is located, relative to the resource’s "front-left-bottom" origin.

#### Why at the `Resource` Level?
Barcodes can be applied to any physical object with x, y, and z dimensions. Therefore, it makes sense to include this functionality in the `Resource` class, covering **all** physical objects that might be placed in a workcell. The exception is `ResourceStack`, which is a dynamic aggregation and could be handled separately if needed.

#### Why Not Use the `name` Attribute?
The `name` attribute:
1. Is often human-generated and cannot be relied upon for barcode-like precision.
2. Should remain human-readable.
3. May be generated dynamically (e.g., a tip rack’s tips' names are created when the rack is generated, but a potential barcode physically added would be different).
4. Does not store symbology or position information.

---

#### Auto-Identifying Resource Orientation
With barcode information, the orientation of resources can be automatically identified.
For example, scanning multiple sides of a plate + _knowing_ the side the barcode _is_ position allows the system to autodetect the orientation of a resource, useful for workflows involving multiple rotations of resources or when correcting mistakes in resource placement.

#### Simplified Computer Vision
By storing barcode information directly in the resource, computer vision tasks can be simplified, especially with 2D codes (like fiducial markers). These can be used to:
- Quickly identify resources in images with higher robustness due to in-built error correction encoding (e.g. AprilTags)
- Determine the relative position of the barcode for easier scanning and processing.

The following barcode types are examples of what could be supported via integration of this information (but in reality there is not limit):

```go
Barcodes
├── 1D (Linear)
│   ├── UPC (Universal Product Code)
│   │   ├── UPC-A (12 digits)
│   │   └── UPC-E (6 digits, zero-suppressed)
│   ├── EAN (European Article Number)
│   │   ├── EAN-13 (13 digits)
│   │   └── EAN-8 (8 digits)
│   ├── ISBT 128 (ISBT Standard)
│   ├── Code 39 (alphanumeric)
│   ├── Code 93 (enhanced alphanumeric)
│   ├── Code 128
│   │   ├── Subset B (alphanumeric & ASCII)
│   │   └── Subset C (numeric pairs, high density)
│   ├── Interleaved 2 of 5 (numeric, pair-encoded)
│   ├── Codabar (Codebar) (numeric + a few symbols)
│   └── MSI (Modified Plessey, numeric)
└── 2D (Matrix)
    ├── General-Purpose 2D Codes
    │   ├── QR Code
    │   │   └── Micro QR (mini version)
    │   ├── Data Matrix
    │   ├── PDF417
    │   │   └── Micro PDF417
    │   ├── Aztec Code
    │   └── MaxiCode
    └── Fiducial Markers (specialized 2D)
        ├── AprilTags (tag36h11, tag25h9, …)
        ├── ArUco markers
        ├── ARToolKit markers
        └── ChArUco / AprilGrid setups
```

#### Movement Correction with Fiducial Markers
Using fiducial markers stored in the barcode data, robotic arms can correct their movement in 3D space. This longer-term goal allows more precise movement correction during automation tasks.

### Next Steps
This PR only proposes a streamlined approach to storing and dynamically updating barcode information within resources. To bring these use cases to production, further testing and iterations will be needed.

### DIFF
--- a/pylabrobot/resources/barcode.py
+++ b/pylabrobot/resources/barcode.py
@@ -3,8 +3,8 @@ This module defines the Barcode class, which represents a barcode associated wit
 It includes attributes for the barcode data, symbology, and its position on the resource.
 """
 
-from typing import Literal
 from dataclasses import dataclass
+from typing import Literal
 
 BarcodePosition = Literal["right", "front", "left", "back", "bottom", "top"]
 


BioCam:
> 1. should this be on the Resource level? This means every individual Well etc would have a barcode. On the other hand, adding it below would mean we change Plate, TipRack, PlateCarrier, TipCarrier, etc.
>    a. this could be neatly implemented as mixin: `Barcoded` or something like that

I would definitely say yes to adding the barcode attribute at the resource level:
Wellplates with every well having a different barcodes exist: https://www.azenta.com/products/individual-access-96-well-skirted-flat-bottom-pcr-plate-2d-coded

I also think we should not constraint the usage of barcodes to what we believe exists:
e.g. I could imagine using specialised microfluidic devices or glass slides -> they can easily be modelled in PLR in a generic sense, as a `Resource`.
Having the barcode attribute at the highest level now gives us the ability to encode specific barcodes into the models of microfluidic device or glass slide, physically represented as small stickers or laser engravings :)

> 2. since all these barcode attributes go together: how about a `barcode: Barcode` parameter?
> 
> ```python
> @dataclass
> class Barcode:
>   barcode: Union[str, int]
>   barcode_symbology: str
>   barcode_position_on_resource: Literal["right", "back", "left", "front", "bottom", "top"]
> ```

I really like this, and completely agree: bundling up the connected information is very useful!
I would make it...

 ```python
@dataclass
class Barcode:
   data: Union[str, int] = None
   symbology: str = None
   position_on_resource: Literal["right", "front", "left", "back", "bottom", "top"] = None
```
...though.

This way the "data" a barcode carrier can be added dynamically using, e.g. ...

```python
plate.barcode.data ="WELL-A1-AND-BEYOND"

tube.barcode.data = "GENE-YOLO-42"

incubator_shaker.barcode.data = "E.COLI.EXPRESS.YOURSELF"
```

But this raises the question: 
should each resource evaluate `resource.barcode=None` by default, or should every resource "have" a barcode whose subattributes all evaluate to none?

rickwierenga:
> Wellplates with every well having a different barcode exist: https://www.azenta.com/products/individual-access-96-well-skirted-flat-bottom-pcr-plate-2d-coded

I was not aware such a thing exists! I'm convinced it should be Resource level.

> This way the "data" a barcode carries can be added dynamically using, e.g. ...

You mean added after initialization time? In any case, the data can be modified at runtime.

> should each resource evaluate resource.barcode=None by default, or should every resource "have" a barcode whose subattributes all evaluate to none?

Part of the motivation for moving everything into a single `barcode` attribute is the `None`-case would apply to all attributes together. Through the type checker, it prevents ghost-resources with for example `data` being not None but symbology being `None`. That is a very confusing situation. Making individual attributes of `Barcode` nullable actually nullifies this benefit! For location: it's debatable: should `Barcode.position_on_resource` be nullable?

BioCam:
> You mean added after initialization time? In any case, the data can be modified at runtime.
> 
> >

Yes, exactly. The power of this class attribute structure stems from both:
1. pre-define `Barcode`s
2. instantiate resources during run-time, _and then_ add barcode information to them: e.g. just stack new barcodes plates into a `ResourceStack` and make the script _indentify_ the plate it is using and adding specific liquid combinations into, and then save the barcoded plate information (maybe to track between automation runs)

-> part two enables i.a. one powerful feature: user **do not have to specify the barcode data** _before the run_! 
(which is always a time-consuming, error-prone step that PLR would now enable to eliminate)



> Part of the motivation for moving everything into a single `barcode` attribute is the `None`-case would apply to all attributes together. Through the type checker, it prevents ghost-resources with for example `data` being not None but symbology being `None`. That is a very confusing situation. Making individual attributes of `Barcode` nullable actually nullifies this benefit! For location: it's debatable: should `Barcode.position_on_resource` be nullable?

You convinced me: `Barcode` itself should be nullable but all of the `Barcode` attributes should be present when instantiating a `Barcode` and setting `Resource.barcode` to it.
`Barcode.position_on_resource` should not be nullable by default then: if a `Barcode` is instantiated it should always have this _barcode minimal information_ stored. Maybe we'll revise this when tested through real-world usage but the information we're asking for is really just bare minimum.



> I think the position should be more specific than currently envisioned. just consider a labeled carrier on a STAR: If I just say the BC is on the left, it's still quite a big surface where the BC could actually be.

Yeah this gave me a lot of headache when making this proposal:
Using simple...
```python
  Barcode.position_on_resource: Literal["right", "back", "left", "front", "bottom", "top"]
```
- Con: the big surface problem -> information is most likely not sufficient to _directly_ find the barcode
- Pro:
  - fast & easy to specify
  - opposite of the con: _does not restrict_ the barcode search to 1 particular location, but opens up barcode search to an entire parent's (sur)face
  - related: **does not restrict self-created barcodes to be absolutely perfect** !

The last point was really why I've proposed this approach:
self-created barcodes are an incredible hassle with 1D barcodes (not just Hamilton's :) ).
And that is not a surprise, the technology is ancient.
Even a slight tilt or off-target positioning of the physical barcode sticker leads to detection failure... and then a human robot-sitter to chase after it.

I was imagining more the use of 2D barcodes here tbh:
**Simply stick them anywhere on the resource**.
 The `Barcode.position_on_resource` is then plenty of information because the precise location of the barcode doesn't matter anymore -> the imaging can be easily set up to capture the entire face/side of the plate, and a 2D barcode - tilted or off-target  positioned, doesn't matter anymore.

You are making a good point though about _big_ resource surface like in the case of a Hamilton carrier.
I don't know what would be best here.

I could imagine some way of specifying the resource-relative center,center,center location of a barcode on that resource?
But this is a massive strain on human operators to get that location correct every time?

@jrast: For your specific Hamilton carrier example: Doesn't the firmware itself only allow searching of a barcode at a very precisely _firmware_-defined carrier location, i.e. the right-back in vertical format?
(`C0CI` if I remember correctly 👀 )

PR #523, opened 2025-05-24
Create Utility Function `chunk_list`

BioCam:
```python
def chunk_list(list_: List[T], chunk_size: int) -> List[List[T]]:
  """Divide a list into smaller chunks of a specified size.

  Args:
      list_: The list to be divided into chunks.
      chunk_size: The size of each chunk.

  Returns:
      A list of chunks, where each chunk is a list of elements.

  Example:
      >>> chunk_list([1, 2, 3, 4, 5, 6, 7, 8, 9], 3)
      [[1, 2, 3], [4, 5, 6], [7, 8, 9]]

      >>> chunk_list([1, 2, 3, 4, 5], 2)
      [[1, 2], [3, 4], [5]]
  """
  chunks: List[List[T]] = []
  for i in range(0, len(list_), chunk_size):
    chunks.append(list_[i : i + chunk_size])
  return chunks
```

very useful i.a. for y-independent multi-channel liquid handlers like the STAR, Vantage, Prep, EVO

### DIFF
--- a/pylabrobot/utils/list.py
+++ b/pylabrobot/utils/list.py
@@ -9,8 +9,8 @@ def assert_shape(list_: List[List[T]], shape: Tuple[int, int]):
   """Assert that a list has the correct shape.
 
   Args:
-      list_: The list to check.
-      shape: The expected shape.
+    list_: The list to check.
+    shape: The expected shape.
   """
 
   if len(list_) != shape[0]:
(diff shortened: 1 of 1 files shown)


rickwierenga:
todo for me: check that sphinx builds the doc strings correctly

PR #524, opened 2025-05-27
Build `lh.probe_tip_presence_via_pickup()`

BioCam:
Hi everyone,

In this PR I am proposing the addition of the "atomic command" for tip_probing:

```python
  async def probe_tip_presence_via_pickup(
    self,
    tip_spots: List[TipSpot],
    use_channels: Optional[List[int]] = None
) -> List[int]:
    """
    Probe tip presence by attempting pickup on each TipSpot.

    Args:
        tip_spots: TipSpots to probe.
        use_channels: Channels to use (must match tip_spots length).

    Returns:
        List[int]: 1 if tip is present, 0 otherwise.
    """
```

This is step 1 of a multi-feature generation process:

Build lh.probe_tip_presence_via_pickup()
Build measured_tip_inventory = lh.probe_tip_inventory(probing_mode=lh.probe_tip_presence_via_pickup)
Build lh.update_tip_inventory(measured_tip_inventory)
Build lh.consolidate_tip_inventory()

For further details please see the user & developer forum thread:
https://discuss.pylabrobot.org/t/liquid-handler-auto-detection-of-real-tip-presence/212/3


pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -2276,6 +2276,85 @@ def assign_child_resource(
...
+    """
+    Probe tip presence by attempting pickup on each TipSpot.
+
+    Args:
+        tip_spots: TipSpots to probe.
+        use_channels: Channels to use (must match tip_spots length).
+
+    Returns:
+        List[int]: 1 if tip is present, 0 otherwise.
+    """
```
rickwierenga, line 2293: wouldn't you want to return a bool?

pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -2276,6 +2276,85 @@ def assign_child_resource(
...
+                if ch in channel_subset:
+                    failed_local_idx = channel_subset.index(ch)
+                    presence_flags[index_subset[failed_local_idx]] = 0
+
+        # Step 3: Drop tips immediately after probing
+        successful = [
+            (spot, ch) for spot, ch, i in cluster
+            if presence_flags[i] == 1
+        ]
+        if successful:
```
rickwierenga, line 2344: maybe `if any(successful)` so it's more readable

pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -2276,6 +2276,85 @@ def assign_child_resource(
...
+            try:
+                await self.drop_tips(
+                    [spot for spot, _ in successful],
+                    use_channels=[ch for _, ch in successful],
+                    # minimum_traverse_height_at_beginning_of_a_command=z_height,
+                    z_position_at_end_of_a_command=z_height
+                )
+            except Exception as e:
+                print(f"Warning: drop_tips failed for cluster at x={cluster[0][0].location.x}: {e}")
```
rickwierenga, line 2353: how do you know it's this cluster?

pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -2276,6 +2276,85 @@ def assign_child_resource(
...
+
+    sorted_clusters = [clusters_by_x[x] for x in sorted(clusters_by_x)]
+
+    # Step 2: Probe each cluster
+    for cluster in sorted_clusters:
+        tip_subset, channel_subset, index_subset = zip(*cluster)
+
+        try:
+            await self.pick_up_tips(
+                list(tip_subset),
```
rickwierenga, line 2328: is list conversion necessary?

rickwierenga:
very nice and clever feature addition!

> atomic command

i wouldn't call this an atomic command since it uses the atomic command `pick_up_tips`. (it requires no changes to the backend! - any backend in PLR that raises the appropriate error will get this functionality for free)

rickwierenga:
why don't we support many tip spots? e.g. an entire tip rack, and this function figuring the optimal trial pattern (which you have already implemented)

BioCam:
> why don't we support many tip spots? e.g. an entire tip rack, and this function figuring the optimal trial pattern (which you have already implemented)

Main reasons: extendability & intuition

- This is the "lowest" level of tip presence probing (inside PLR), that means as a user I want to have a good understanding of what to expect when executing this command: expecting the same number of `tip_spots` and `channels` to be given is in alignment with all other lowest level commands (tip pickup itself, aspirate, dispense, drop).
Anything beyond that is a surprise, and I don't want to be surprised by my commands 😂 
- Extendability: using this command to probe an entire rack or all racks of a deck is now easy, and as highlighted in #519 is actually the next step of this feature development series:
image

Bonus: by having this lowest level of tip presence being so "atomic-ish" we can also extend it in a different sense: replace it with an alternative that does the exact same tip presence probing - but not via tip_pickup (general lh useful) but via sleeve_check (specific for Hamilton lh).
If we'd now have to add additional inter-tip_rack logic at the low-level `lh.probe_tip_presence_via_pickup()` we would have to copy the exact same logic again into `lh.backend.probe_tip_presence_via_sleeve_check()`, going against DRY.

rickwierenga:
thanks for the explanation, agreed!

as i dm'd you: this feature is gonna be so useful. i have had multiple instances since you opened these PRs where I wished I'd already merged it!

PR #532, opened 2025-05-29
add extra spacing on the top slot of cytomat racks

ben-ray:
IMG_0017
there is always extra room for taller reservoirs on the top slot of a cytomat. this change has been thoroughly tested in production


rickwierenga:
did this by actualy making the `PlateHolder`s taller rather than hacking it into the indexing function.

ben-ray:
> did this by actualy making the `PlateHolder`s taller rather than hacking it into the indexing function.

much better

PR #534, opened 2025-05-29
Remove callback support from LiquidHandler

rickwierenga:
## Summary
- drop all operation callback functionality from `LiquidHandler`
- update docs accordingly
- remove callback tests

## Testing
- `make test`


rickwierenga:
same as #535 

for some reason it opened 2

PR #538, opened 2025-05-30
Add cooling capability flag to temperature controller backends

rickwierenga:
## Summary
- add abstract property `supports_active_cooling` to `TemperatureControllerBackend`
- override property returning `False` on all existing backends

## Testing
- `make format`
- `make typecheck`
- `make test`


rickwierenga:
inheco does support it, i edited the code to reflect that

PR #549, opened 2025-06-05
Implement flow rate for Tecan EVO backend

rickwierenga:
## Summary
- add flow rate handling to aspirate and dispense for the Tecan EVO backend
- test custom flow rates in EVO backend tests

## Testing
- `pytest -q` *(fails: ModuleNotFoundError: No module named 'serial', 'responses', 'websockets', 'werkzeug', 'requests')*

------
https://chatgpt.com/codex/tasks/task_e_6840fd261fc0832b880c36b96367ad43

### DIFF
--- a/pylabrobot/liquid_handling/backends/tecan/EVO_backend.py
+++ b/pylabrobot/liquid_handling/backends/tecan/EVO_backend.py
@@ -797,8 +797,9 @@ class EVOBackend(TecanLiquidHandler):
       tlc = tecan_liquid_classes[i]
       z = zadd[channel]
       assert tlc is not None and z is not None
-      sep[channel] = int(tlc.aspirate_speed * 12)  # 6?
-      ssz[channel] = round(z * tlc.aspirate_speed / ops[i].volume)
+      flow_rate = ops[i].flow_rate or tlc.aspirate_speed
+      sep[channel] = int(flow_rate * 12)  # 6?
+      ssz[channel] = round(z * flow_rate / ops[i].volume)
       volume = tlc.compute_corrected_volume(ops[i].volume)
       mtr[channel] = round(volume * 6)  # 3?
       ssz_r[channel] = int(tlc.aspirate_retract_speed * 10)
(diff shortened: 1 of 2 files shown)


nedru004:
This change works for me.

rickwierenga:
@nedru004 does it have the correct flow rate?

on the forum we discussed perhaps the units are wrong

would you be able to confirm this? then we can fix and/or merge this

https://discuss.pylabrobot.org/t/subdevice-error-in-evo200-dispensing/166

nedru004:
The multiplier in the comments with the question mark seems to give the correct steps per uL. I changed to those values and checked the volume dispensed with a pipette which was close to the volume requested. I am running the liquid system with air right now. I assume I will have error in the volumes since so much air is compressed between the pipette and the diluters, but hopefully not much.

rickwierenga:
> The multiplier in the comments with the question mark seems to give the correct steps per uL. I changed to those values and checked the volume dispensed with a pipette which was close to the volume requested.

nice

> I am running the liquid system with air right now. I assume I will have error in the volumes since so much air is compressed between the pipette and the diluters, but hopefully not much.

in the tubes that connect the pumps to the channel heads?

nedru004:
> in the tubes that connect the pumps to the channel heads?

Yeah. I dont have the air LiHa and I dont have the wash/flush protocol, so I am just running the syringes/pumps dry at the moment. I am not sure if that is bad for the machine (lubrication?) and if that has an effect on volume at fast flow rates. That is why I slowed it down.

rickwierenga:
I do not know for sure but I think it would change the liquid handling characteristics.

Do you want to make a PR with your changes into this branch? Then we can have a verified implementation. Don't worry about fixing the unit tests, I can do that if you want.

rickwierenga:
updated the tests to reflect the new behavior

@nedru004 can you please confirm this last version is working as expected for you? then i will merge this branch

nedru004:
This version works for me. Thanks

rickwierenga:
thanks for confirming!

rickwierenga:
and congrats on your first commit being merged into plr:main! :D

PR #562, opened 2025-06-05
Update `waste_block` Name & y-Origin Coordinate

BioCam:
Hi everyone,

## Issues
1. the "waste block" as Hamilton names it has so far been called "teaching_carrier"
2. its y-origin coordinate has been protruding into the autoload area, i.e. was too low

## PR Content 

I updated the name.

I updated the y-origin location - while ensuring there is absolutely no change to the tip_spots inside the "teaching_tip_rack" which the "waste_block" hosts.

image

-> is now...

image


rickwierenga:
great example of in-silico development for PLR using the Visualizer :D

PR #564, opened 2025-06-07
Visualizer Coloring by Resource Subclass

BioCam:
image

### DIFF
--- a/pylabrobot/visualizer/lib.js
+++ b/pylabrobot/visualizer/lib.js
@@ -239,16 +239,7 @@ class Resource {
     if (RESOURCE_COLORS.hasOwnProperty(this.constructor.name)) {
       return RESOURCE_COLORS[this.constructor.name];
     }
-    if (
-      this.constructor.name === "Resource" &&
-      this.name.toLowerCase().includes("workcell")
-    ) {
-      return "lightgrey";
-    }
-    if (RESOURCE_COLORS["Resource"]) {
-      return RESOURCE_COLORS["Resource"];
-    }
-    return "#eab676";
+    return RESOURCE_COLORS["Resource"];
   }
 
   // Properties influenced by mode
--- a/pylabrobot/visualizer/lib.js
+++ b/pylabrobot/visualizer/lib.js
@@ -881,41 +881,12 @@ class Trash extends Resource {
   }
 }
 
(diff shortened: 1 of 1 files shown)


pylabrobot/visualizer/lib.js
```diff
@@ -646,61 +734,133 @@ class Container extends Resource {
...
+}
+
+class Tube extends Container {
+  drawMainShape() {
+    // exactly the same pattern as Well (circle background + liquid overlay)
+    const volume = this.getVolume();
+    const alpha = this.maxVolume > 0 ? volume / this.maxVolume : 0;
+    const liquidColor = `rgba(239, 35, 60, ${alpha})`;
+
+    const radius = (1.25 * this.size_x) / 2;
```
rickwierenga: why 1.25?

pylabrobot/visualizer/lib.js
```diff
@@ -646,61 +734,133 @@ class Container extends Resource {
...
+    const group = new Konva.Group();
+
+    // white circular background
+    const background = new Konva.Circle({
+      radius: radius,
+      fill: "#E0EAEE",
```
rickwierenga: shouldn't this be white for fill?

pylabrobot/visualizer/lib.js
BioCam: That is an error - thank's for catching this!

pylabrobot/visualizer/lib.js
BioCam: No, I think it is clearer if the "white" of a Container and the "white" of a tip_spot are different shades of "white"; 
I found it a lot easier to visually distinguish them (e.g. empty tip_rack vs empty wellplate)

pylabrobot/visualizer/lib.js
```diff
@@ -825,10 +943,41 @@ class Trash extends Resource {
...
+  getColor() {
+    return RESOURCE_COLORS["TroughCarrier"];
+  }
+}
+
+class TubeCarrier extends Carrier {
+  getColor() {
+    return RESOURCE_COLORS["TubeCarrier"];
+  }
+}
```
rickwierenga: Considering Resource dynamically loads the color based on the constructor name, I think these are not needed.

PR #577, opened 2025-06-13
Change flowrate multipliers

nedru004:
Change multiplier for flowrate conversion to steps. Multiplier was 2x higher than it should be. Used values in comments. Commands SEP and SPP are both flowrates and set in half-steps per second. MTR is a volume and set in full steps.


rickwierenga:
thanks!

commit message was "Change multiplier for flowrate conversion to steps. Multiplier was 2x higher than it should be. Used values in comments. Commands SEP and SPP are both flowrates and set in half-steps per second. MTR is a volume and set in full steps."

PR #586, opened 2025-06-20
Update default values in STARBackend to use floats instead of integers

linx5o:
For values which are rounded in the `aspirate_pip` parameter input, changed all `int` to `float`.

Address issue: [Float being converted to Int in Star backend](https://discuss.pylabrobot.org/t/float-being-converted-to-int-in-star-backend/254)

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -1763,7 +1763,7 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
     clot_detection_height = _fill_in_defaults(
       clot_detection_height,
       default=[
-        hlc.aspiration_clot_retract_height if hlc is not None else 0
+        hlc.aspiration_clot_retract_height if hlc is not None else 0.0
         for hlc in hamilton_liquid_classes
       ],
     )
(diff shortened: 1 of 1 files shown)


rickwierenga:
thank you so much for finding and fixing this!

i am a bit confused about changing the default 0 parameters, they shouldn't be affected by this right?

rickwierenga:
... as opposed to fixing the `_fill_in_defaults` function :)

linx5o:
I thought about it as well. Some parameters require ints for inputs and have no other formatting, fill_default would hence act as a last check ensuring correct type. 

For the values I changed they would be multiplied by 10 and rounded in the parameter input of `aspirate_pip` changing them back to int. This just ensures that precision is maintained a little longer before becoming int.

linx5o:
I will check in the lab today if the fix is ok. 

My understanding is that the default of 0 an int cause `val` inputs to `_fill_in_defaults` to all be cast to int.

Changing default of 0.0 of type float would hence cast to float.

linx5o:
Can confirm this addresses the issue. Unsure if it is the optimal solution though.

rickwierenga:
i just pushed a fix in the `_fill_in_defaults` method. while your fix works and is compatible with the old function implementation, i found that we're probably gonna run into the same issue again in the future because it's kind of non-obvious that default has to be a float (just bad api design). it's a little hard to maintain. i.e., you would have to read the entire star backend to find if we missed any defaults that should be a float instead of an int.

lmk if this fix still works :)

linx5o:
I don't think this solution will work. We cannot cast all types to `float`. If we do parameters like this: `immersion_depth_direction=immersion_depth_direction,` will become incorrect. Such values are not rounded in parameter input of `aspirate_pip` which will mean they remain `float` when the `aspirate_pip` requests `int`.

rickwierenga:
you are right, i was being stupid

what do you think about removing the automatic conversion? forcing people to pass a float when a float is expected? it is kind of annoying, but might highlight if we forgot one (the user will pass a float and then we raise an error because the default is an int).

unfortunately, mypy accepts "int" values as float (e.g. `x: float = 1`), meaning we can't automatically type check this part of the code

but i guess also for that reason, we should allow people to still pass what in python is an `int` object?

and then be very careful with having correct 'default' types because as you point out it will lead to incorrect rounding otherwise

linx5o:
I changed the type check to limit to `float` and `int`. I don't think it should be limited to the default type. I think this solution of will work.

rickwierenga:
but what if the required type is neither?

linx5o:
You are right. Maybe best to remove this check as well.

rickwierenga:
wait why do that? isn't it beneficial to flag to the user we might have the wrong type? (i guess sending malformed firmware commands doesn't really hurt, it's just not super nice in terms of api design. harder to debug etc)

linx5o:
If the default type was `int` and the value we gave was `float` this would fail the test right? However, we it should not fail and return the `float` value. 
```
if not all(isinstance(v, t) for v in val):
    raise ValueError(f"Value must be a list of {t}, but is {val}")
```

linx5o:
Side note:
```
if not isinstance(val, list):
    return [val] * len(default)
```
How is this possible given:
```
def _fill_in_defaults(val: Optional[List[T]], default: List[T]) -> List[T]:
```

If possible, this doesn't do any of the checks either.

rickwierenga:
> `if not isinstance(val, list):`

right, that can be removed. it's a leftover from an older version i think

rickwierenga:
> If the default type was int and the value we gave was float this would fail the test right? However, we it should not fail and return the float value.

yes, this is the issue you discovered.

i was considering, should we not cast the user-passed values if they are of the incorrect type? this would be more annoying in the api (the user can't enter 0 instead of 0.0). i think for that reason maybe revert that change

linx5o:
Sorry I'm not following. What to revert to?

Currently the user is responsible for input type if they input values.

rickwierenga:
let's maybe define what the function needs to do:
- if `val` is None, return default
- for every value that is `None` in the `val` list, use the default value instead
- use `default` as validation for `val`:
  - make sure length matches
  - make sure the types match
    - if the user passes an int and the default is float, we should accept that
      - but not the other way around
      - in this case, we can automatically convert it

this is redefined from the original implementation, which as you pointed out was buggy.

linx5o:
> make sure the types match

I think we just drop this check. Have the user be responsible for types.

We could also write a separate function for values which should be int.

Rounding all values before passing to aspirate_pip is also an option. But it brings us back to here where these bugs would be hard to find and debug.

Move the conversion of units from plr to Hamilton to input of `fill in defaults` then all output should be int.

rickwierenga:
yes, i don't like the idea of rounding/converting units in this function.

maybe just drop the check (and automatic conversion)

rickwierenga:
good to go?

linx5o:
Looks good

rickwierenga:
thanks for flagging thoughts and PR!

PR #589, opened 2025-06-28
`use_channels` must be unique

rickwierenga:


rickwierenga:
http://discuss.pylabrobot.org/t/timeouterror/122/48

PR #590, opened 2025-06-29
quick calibration for 4x and 20x objectives on cytation1 w/ 1288x964 pixel Blackfly BFLY-U3-13S2M sensor

ben-ray:
beware not all cytations may be built equally, so if you have a different sensor, it might be worth calculating these values for your particular setup

you can find the mapping between pixels <> microns by exporting an image with scalebar from gen5.exe using your cytation & counting the pixels of the scalebar

### DIFF
--- a/pylabrobot/plate_reading/biotek_backend.py
+++ b/pylabrobot/plate_reading/biotek_backend.py
@@ -1,4 +1,5 @@
 import asyncio
+import enum
 import logging
 import math
 import re
@@ -75,11 +76,11 @@ async def _golden_ratio_search(
 
   return (b + a) / 2
 
-CytationModel = Literal["Cytation1", "Cytation5"]
+CytationModel = Literal["cytation1", "cytation5"]
 
 @dataclass
 class CytationImagingConfig:
-  model: CytationModel = "Cytation5"
+  model: CytationModel = "cytation5"
   camera_serial_number: Optional[str] = None
   max_image_read_attempts: int = 8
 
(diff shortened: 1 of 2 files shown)


pylabrobot/plate_reading/biotek_backend.py
```diff
@@ -87,6 +91,19 @@ class Cytation5ImagingConfig:
   filters: Optional[List[Optional[ImagingMode]]] = None
 
 
+_FOV: dict[str, dict[float, tuple[float, float]]] = {
+  "cytation1": {
+    4: (1288 / 596, 964 / 596),
+    20: (1288 / 3000, 964 / 3000),
```
rickwierenga, line 97: what is going on with these denominators?

for the cytation5, it's to convert um to mm

ben-ray: The cytation5 sensor is 2448x2048 while the cytation1 is 1288x964. The scaling factor from pixel to micron will be identical for both axis, but why does the image_size conversion not take the sensor size into account? I would expect the X and Y size to be different but the scaling factor is identical in X and Y in the original Cytation5 implementation. 

These numbers are my attempt to account for the significantly different x and y size of my sensor (my pixels are still perfectly square)

rickwierenga: this is not a scaling factor

ben-ray: Then how do I determine these and why are they not different for X and Y on the cytation5?

rickwierenga: i took the number from gen5.exe, and then did um -> mm conversion which seems to work

rickwierenga: at least it corresponds with their firmware commands.

you can use a firmware command to check

PR #607, opened 2025-07-19
Add new MFX DWP module (cat.-no. 188042) with pedestal support

har1eyk:
This commit introduces a new function, `MFX_DWP_module_188042`, for handling the Hamilton MFX DWP Module with the catalog number 188042. This module includes a pedestal that elevates the plate by 6mm, and the function accurately reflects this by setting the `pedestal_size_z` to 6.45mm. The function also specifies the module's dimensions (size_x and size_y) and positions (child_x and child_y) to account for the metal clamps at the corners. Additionally, the `size_z` calculation is updated to include the pedestal height, ensuring the correct overall height when the module is placed on the MFX carrier. The function returns a `PlateHolder` object with all these parameters initialized appropriately.

### DIFF
--- a/pylabrobot/resources/hamilton/mfx_modules.py
+++ b/pylabrobot/resources/hamilton/mfx_modules.py
@@ -33,12 +33,45 @@ def MFX_DWP_rackbased_module(name: str) -> PlateHolder:
     name=name,
     size_x=135.0,
     size_y=94.0,
-    size_z=178.0 - 18.195 - 100,
+    size_z=178.0 - 18.195 - 100, #59.81mm
     # probe height - carrier_height - deck_height
     child_location=Coordinate(4.0, 3.5, 178.0 - 18.195 - 100),
     model=MFX_DWP_rackbased_module.__name__,
     pedestal_size_z=0,
   )
+
+def MFX_DWP_module_188042(name: str) -> PlateHolder:
+    """Hamilton MFX DWP Module (cat.-no. 188042 / 188042-00).
+    This module contains a pedestal that elevates the plate by 6mm. The plate rests on the pedestal
+    It also contains metal clamps at the corners. 
+    https://www.hamiltoncompany.com/other-robotics/188042
+    """
+    # ── module footprint ──────────────
+    # size_x = 132.50          # mm
+    size_x = 135
+    # size_y = 92.05           # mm
+    size_y = 94 
+
+    # ── vertical geometry ──────────────────────────────
+    pedestal_size_z = 6.45   # The height of the pedestal
+    # size_z = 178.3 - 18.195 - 100   # 60.10mm is module height above MFX carrier. This height does not including pedestal
+    size_z = 184.745 - 18.195 - 100   # 60.10mm is module height above MFX carrier. This height includes pedestal.
+    # module.get_absolute_size_y() = 184.745 puts channel right on the pedestal
+
+    # ── where the plate’s lower-left corner sits ───────
+    child_x = 4.0           # mm  
+    child_y = 3.5           # mm
+    child_location = Coordinate(child_x, child_y, size_z)
+
+    return PlateHolder(
+        name=name,
+        size_x=size_x,
+        size_y=size_y,
+        size_z=size_z,
+        child_location=child_location,
+        model=MFX_DWP_module_188042.__name__,
+        pedestal_size_z=pedestal_size_z,
+    )
 
 
 def MFX_DWP_module_flat(name: str) -> PlateHolder:


pylabrobot/resources/hamilton/mfx_modules.py
```diff
@@ -33,14 +33,50 @@ def MFX_DWP_rackbased_module(name: str) -> PlateHolder:
...
+  """Hamilton MFX DWP Module (cat.-no. 188042 / 188042-00).
+  This module contains a pedestal that elevates the plate by 6mm. The plate rests on the pedestal
+  It also contains metal clamps at the corners.
+  https://www.hamiltoncompany.com/other-robotics/188042
+  """
+  # ── module footprint ──────────────
+  # size_x = 132.50          # mm
+  size_x = 135
+  # size_y = 92.05           # mm
+  size_y = 94
```
rickwierenga: what are these commented values?

pylabrobot/resources/hamilton/mfx_modules.py
```diff
@@ -33,14 +33,50 @@ def MFX_DWP_rackbased_module(name: str) -> PlateHolder:
...
+  # size_z = 178.3 - 18.195 - 100   # 60.10mm is module height above MFX carrier. This height does not including pedestal
+  size_z = (
+    184.745 - 18.195 - 100
+  )  # 60.10mm is module height above MFX carrier. This height includes pedestal.
+  # module.get_absolute_size_y() = 184.745 puts channel right on the pedestal
+
+  # ── where the plate’s lower-left corner sits ───────
+  child_x = 4.0  # mm
+  child_y = 3.5  # mm
+  child_location = Coordinate(child_x, child_y, size_z)
```
rickwierenga: please add the source of these numbers

either `# measured` or `# from spec`

if there is a spec, please link it

rickwierenga:
thanks!!

added two small comments

pylabrobot/resources/hamilton/mfx_modules.py
har1eyk: carryover from testing. Removed.

pylabrobot/resources/hamilton/mfx_modules.py
har1eyk: Measured values. Added comment to say they were measured.

rickwierenga:
thanks!

and just to confirm: you have tested this definition in the real world right?

BioCam:
Hi @har1eyk,

Thank you for making this PR!

I've actually made a definition for this part a couple of months ago and simply haven't gotten around to integrating it into PLR yet.

Some of my values are a bit different and it might be good to figure out why:

My definition:
```python
def Hamilton_MFX_plateholder_188042(name: str) -> PlateHolder:
  """Hamilton cat. no.: 188042
  Plate holder module w/ pedestal
  - tapped/clamps: Yes
  """

  return PlateHolder(
    name=name,
    size_x=135.0,
    size_y=94.0,
    size_z=183.95 - 18.195 - 100,
    # probe height - carrier_height - deck_height
    child_location=Coordinate(4.0, 4.0, 183.95 - 18.195 - 100),
    model=Hamilton_MFX_plateholder_188042.__name__,
    pedestal_size_z=-4.74,
  )
  ```

Your definition:
```python
def MFX_DWP_module_188042(name: str) -> PlateHolder:
  """Hamilton MFX DWP Module (cat.-no. 188042 / 188042-00).
  This module contains a pedestal that elevates the plate by 6mm. The plate rests on the pedestal
  It also contains metal clamps at the corners.
  https://www.hamiltoncompany.com/other-robotics/188042
  """
  # ── module footprint ──────────────
  size_x = 135 # measured
  size_y = 94 # measured

  # ── vertical geometry ──────────────────────────────
  pedestal_size_z = 6.45  # measured The height of the pedestal
  size_z = (
    184.745 - 18.195 - 100
  )  # 60.10mm is module height above MFX carrier. This height includes pedestal.
  # module.get_absolute_size_y() = 184.745 puts channel right on the pedestal with 50ul fil tip

  # ── where the plate’s lower-left corner sits ───────
  child_x = 4.0  # mm measured
  child_y = 3.5  # mm measured
  child_location = Coordinate(child_x, child_y, size_z)

  return PlateHolder(
    name=name,
    size_x=size_x,
    size_y=size_y,
    size_z=size_z,
    child_location=child_location,
    model=MFX_DWP_module_188042.__name__,
    pedestal_size_z=pedestal_size_z,
  )
```

rickwierenga:
thanks @BioCam 

also @har1eyk, this was partly a mistake on my side, the `pedestal_size_z ` must be a negative number. this was documented in the PlateCarrier docs, which I moved to the PlateHolder docs in https://github.com/PyLabRobot/pylabrobot/commit/266ac2174b57a505c3e62f0e636402a8847a8a50. Docs will be live on the website in approximately 40 minutes. You can view the file here: https://github.com/PyLabRobot/pylabrobot/blob/main/docs/resources/resource-holder/plate-holder.md and in the future docs will be here: https://docs.pylabrobot.org/resources/resource-holder/plate-holder.html

and to repeat the information from the docs here: the pedestal height is measured to the second highest point, e.g. the little ridges, not the bottom. given the discrepancies between your and camillo's definition + the unclear docs before, that might explain the issue

rickwierenga:
the last thing, brought up by @BioCam is that we might want to reconsider the naming of this resource. So far we have tried to optimize for readable (but consistent and unique) names that do not include the part number.

There is not standard for naming mfx modules (yet?)

and since they are all so similar, we might have to use the part number. but if we can avoid that that would be best imo

BioCam:
**Difference 1: `pedestal_size_z`**
Have you measured the pedestal height from the top surface to the 1st or 2nd lower surface?
There are 2 thin elevations on the left and right side of this plate_holder underneath the pedestal top.
Any plate placed on top of this plate_holder that has a very high `dz` value will sit on top of these smaller side elevations and never reach the full bottom.
I have made a video on automated probing of the 3 levels of a plate_holder like this here: https://www.youtube.com/watch?v=_uPf9hyTBog
Screenshot 2025-07-19 at 22 21 37

Rick also shows how to measure this in the docs here: https://docs.pylabrobot.org/resources/carrier/plate-carrier/plate_carrier.html#pedestal-z-height

**Difference 2: small difference in child_location -> y=4 VS y=3.5**
What I have done? - I looked at the VENUS "ground truth" data represented in the definition of the "standard" plate carrier y=8.5mm,
and then compared it with the first site of the MFX base carrier y=5mm. 
reference plate_car to mfx_car plus plate_holder

When channels move between plate carriers and mfx carriers they don't seem to move in the y-direction (would be very visible with core-based plate transfers).
This also makes sense from a design perspective -> it makes plate_holders more consistent (no matter mfx or plate carrier), and transfers faster.

I therefore measured it to be `8.5 - 5 = 3.5` mm.

**Difference 3: topmost measured surface of pedestal**
> 184.745 puts channel right on the pedestal with 50ul fil tip

Did you measure this with clld_probing or ztouch_probing?

**Error in both definitions**
We both said that the `size_z` of the entire plate holder is the height of the pedestal.
That is incorrect.

The `size_z` marks the lowest to the tallest point of a resource.
That means the tallest point is actually the top of the clamp frames.

(Why is this important: a) PLR tries to be as close to physical reality as possible, b) we are building automated crash prevention features that take the `size_z` as the known tallest point of a resource. This crash prevention feature will think that channels can move into the clamps... )

har1eyk:
Thanks for your comments. 

@rickwierenga, yes, I've tested this module definition in the real world with a variety of plasticware: tuberacks, 96w plates, DW plates, etc. 

Awesome work, @BioCam . Thank you for providing your measurements. I appreciate knowing there are discrepancies. I used calipers to calculate my pedestal size. I'm lost with some of your calculations and also how it's represented as a negative number. The `child_x, child_y` distances weren't rigorously tested by me, so I'm inclined to accept your measurements.

My pedestal height was calculated by taking the total width:
image

and subtracting the ridge width. 
image

BioCam:
The issue regarding the `pedestal_size_z` stems from this way of taking a measurement:

> and subtracting the ridge width. image

-> it is missing that a plate will never reach this "bottom" position because any SLAS-format plate will always be "stuck" higher up due to these ridges, marked here in yellow:

2987f75803c46a81c3a3d0f5d3fa3773b928a935_2_1380x442

We found the most accurate measurements + most peer-reviewable measurements by other PLR contributors are taken by the STAR(let) itself, i.e. using [z-probing (as described in the docs) ](https://docs.pylabrobot.org/user_guide/00_liquid-handling/hamilton-star/z-probing.html).
This way we can compare measurements across of the same item - in this case the standard plate_holder top - when we don't have technical drawings provided by the OEM.
The video I shared above shows the video evidence of the measurements for this plate_holder's pedestal_size_z being taken by the STAR itself:

Screenshot 2025-07-21 at 23 06 49

har1eyk:
Thanks for the thorough explanation, @BioCam. I trust your measurements. This module is a tricky one. 

@rickwierenga , let's reject my PR and accept @BioCam 's request instead.

rickwierenga:
wait @har1eyk we can still use the docs and everything that you made!

if you trust + verified camillo's definition, I'm happy to merge that in.

the power of open source definitions! iterate & get more accurate results!

thoughts on naming?

har1eyk:
Thanks, @rickwierenga . Yes, I trust Camillo's definitions. Let's use them. 

I prefer Camillo's name: "Hamilton_MFX_plateholder_188042"; its more descriptive. Maybe drop "Hamilton" so its shorter to "MFX_plateholder_188042". I like that the serial number is in the plate name as this will avoid confusion.

rickwierenga:
@BioCam you should be able to push a commit to Harley's branch

regarding the naming, i was actually thinking we keep hamilton and no part number to be consistent with other PLR resources. It's just the MFX module suffix thing that is not entirely clear.

we already have
- `MFX_DWP_rackbased_module`
- `MFX_DWP_module_flat`
not sure how to name this

rickwierenga:
thanks @har1eyk for first contribution!

thanks @BioCam for sharing this definition!

PR #610, opened 2025-07-20
Add thermocyler suppport

hazlamshamin:
## Description

This PR introduces a new high-level `run_pcr_profile` function within the `Thermocycler` class, simplifying the execution of standard PCR protocols. This function allows users to specify common PCR parameters (denaturation, annealing, extension temperatures and times, number of cycles, etc.) and automatically constructs the underlying temperature profile.

**Summary of Thermocycling Module Additions:**
- An abstract `ThermocyclerBackend` to standardize control interfaces.
- A `ThermocyclerChatterboxBackend` for simulation and testing.
- `OpentronsThermocyclerModuleV1` and `V2` implementations to drive Opentrons thermocyclers over HTTP.
- A high-level `Thermocycler` class for simplified device interaction.
- Updates to `pylabrobot/serializer.py` to enable proper serialization of thermocycler objects.
- Tests for thermocycler, chatterbox and opentrons backend.

The design choice for `run_pcr_profile` is that it generates a *linearized* sequence of steps for the entire PCR run, including all cycles. The main reason is runProfile in Opentrons HTTP API seemingly just do not allow combination of cyclic steps with non-cycic steps. While PCR protocols are conceptually cyclic, representing them as a linear sequence of steps means that backends could report progress through a single, continuous sequence of steps. This might make it easier to track the overall progress of a run and estimate completion times without needing to interpret nested cycle structures.

For users who require more specialized or non-mainstream PCR protocols (e.g., step-down PCR, touchdown PCR, or other complex temperature profiles), the existing `run_profile` function remains fully available and is the recommended method. `run_profile` provides maximum flexibility, allowing users to define any arbitrary sequence of temperature and hold time steps, thereby supporting a wide range of custom thermocycling experiments beyond standard PCR.

**Important Note on Opentrons Backend:**
This feature, particularly the `run_profile` and `run_pcr_profile` functions when used with the `OpentronsThermocyclerBackend`, now relies on the newly introduced `thermocycler_run_profile_no_wait` function in the `opentrons-python-api`. The original `thermocycler_run_profile` in `opentrons-python-api` was a blocking call (due to its `@command` decorator) that would wait for the entire PCR profile to complete. Given that PCR runs can be very long (hours), this blocking behavior often led to client-side timeout errors. The new `thermocycler_run_profile_no_wait` function (decorated with `@request_with_run_id`) addresses this by allowing a "fire-and-forget" execution: the command is sent to the Opentrons API, and the function returns immediately without waiting for the profile to finish. This prevents timeouts and allows `PyLabRobot` to manage the profile monitoring independently.

**Notes (Not Tested):**
- Wetlab of PCR product results.
- Liquid handling operations when labwares (i.e., PCR plate) are directly loaded onto the thermocycler.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Add documentation
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change
- [ ] Refactoring

## Checklist

- [x] I have read the [Contributor Guide](https://docs.pylabrobot.org/contributor_guide/contributing.html) document.
- [x] My code follows the code style of this project.
- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover the new added features.
- [x] All new and existing tests passed.


pylabrobot/thermocycling/thermocycler.py
```diff
@@ -0,0 +1,253 @@
...
+        """Get whether the lid is “open” or “closed”."""
+        return cast(str, await self.backend.get_lid_status())
+
+    async def get_hold_time(self) -> float:
+        """Get remaining hold time (s) for the current step."""
+        return await self.backend.get_hold_time()
+
+    async def get_current_cycle_index(self) -> int:
+        """Get the zero-based index of the current cycle."""
+        return (await self.backend.get_current_cycle_index()) - 1
```
rickwierenga: why not do this conversion in the backend and make the convention that backends return 0-indexed cycles?

pylabrobot/thermocycling/backend.py
```diff
@@ -0,0 +1,78 @@
...
+    """Execute thermocycler profile run."""
+
+  @abstractmethod
+  async def get_block_current_temperature(self) -> float:
+    """Get the current block temperature in °C."""
+
+  @abstractmethod
+  async def get_block_target_temperature(self) -> Optional[float]:
+    """Get the block target temperature in °C."""
+
```
rickwierenga: could the target temperature be something that we handle on the front end? is the reason for doing it on the backend that it will change throughout a profile run?

pylabrobot/thermocycling/opentrons.py
```diff
@@ -0,0 +1,98 @@
...
+    opentrons_id:   OT-API module “id” for your thermocycler.
+    child_location: Position where a plate sits on the block.
+    child:          Optional plate/rack already loaded on the module.
+    """
+    backend = backend or OpentronsThermocyclerBackend(opentrons_id=opentrons_id)
+    super().__init__(
+      name=name,
+      size_x=244.95,  # mm - front/back footprint
+      size_y=172.0,  # mm - left/right footprint
+      size_z=170.35,  # mm - closed-lid height
```
rickwierenga: the convention in PLR is to use size_x for left/right, y for front/back size

is there a reason to switch it here? it is kind of arbitrary in which orientation you measure these values, so we typically choose whatever is the conventional "norm" for that item. for example, landscape for plates. obviously, when choosing an orientation for a parent that implies the size orientation for children (e.g. wells in a plate) (to be technically complete: children might be rotated and have a different size definition)

pylabrobot/thermocycling/opentrons_backend.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class OpentronsThermocyclerBackend(ThermocyclerBackend):
+  """HTTP-API backend for the Opentrons GEN-1/GEN-2 Thermocycler.
+
+  All core functions are supported. run_profile() is fire-and-forget,
+  since PCR runs can outlive the decorator’s default timeout.
+
+  **Note**: Gen-2 via HTTP-API does not expose a separate lid target field,
+  so get_lid_target_temperature() will always return None.
```
rickwierenga: that sounds like an error case to me. should we raise an exception instead?

this api is confusing to me, and many people won't read the api documentation

rickwierenga:
this is an amazing PR!

added a couple questions/comments/discussion points

out of curiosity: did you use ai at all for writing this PR? It is so complete (before review!) that I think it would take many people a lot of time to figure out. I'm curious what tool you used, if any! either way, amazing work!

pylabrobot/thermocycling/thermocycler.py
hazlamshamin: Thanks to this comment, i just noticed a big mistake where the previous PR I created was not exactly same as my local code (i think i forgot to push it). In the updated ones, I actually make one-based indexing as the convention, both in frontend and backend. Opentrons raw dump return the one-based indexing, and it feels to me that it is more readable to use one-based indexing (eg when reading the current cycle 4/30 PCR cycles). Thus, making it consistent in the backend as well. Or do you think backend should be 0-based? Do you mind checking them :))

pylabrobot/thermocycling/backend.py
hazlamshamin: In the TemperatureController, I saw that this is the case. But I think thermocycler is dynamic and fetching the real temperature (both target and current) during a PCR run from the backend is safer? And yes, like you said, they keep changing during run_profile and run_pcr_profile.

pylabrobot/thermocycling/opentrons.py
hazlamshamin: My bad for assuming and not checking how PLR do it. Fixed this as PLR's convention. Also just found the dimension for Opentrons Thermocycler V1 is not the same as V2, in an old PDF, have also fixed this.

hazlamshamin:
Thank you :)). Sorry as well because the previous ones you have reviewed was an old one.... Do you mind reviewing them again? And yes i use gemini-cli (2.5-flash) :)) The current unit tests fail because the new module introduced in ot_api, and im not sure how to fix that dependencies during the tests in the workflow, kindly guidee

pylabrobot/thermocycling/opentrons_backend.py
hazlamshamin: Actually, wrong description. What actually happens is the HTTP-API *does* expose the lid target temperature. But, this is only exposed when the lid is active (eg heating or holding at target temperature, not idle). Which is also the case for the block target temperature, and what's expected as well. I have changed the docstring to describe this properly, but the code will still return None when the lid is not active and the target temperature not exposed. Do you think this is alright?

pylabrobot/thermocycling/backend.py
rickwierenga: Understood! Thanks for explanation!

pylabrobot/thermocycling/opentrons.py
rickwierenga: thanks!

pylabrobot/thermocycling/opentrons_backend.py
rickwierenga: Why do you choose that over raising an error?

rickwierenga:
> he current unit tests fail because the new module introduced in ot_api, and im not sure how to fix that dependencies during the tests in the workflow, kindly guidee

this was dependent on me to publish a new version of the `opentrons-http-api-client` which i just did. https://github.com/rickwierenga/opentrons-python-api/releases/tag/0.1.6

pylabrobot/thermocycling/thermocycler.py
rickwierenga: i think it is best to be consistent across PLR and be 0-indexed

when printing to the user, you can always add the 1 there

rickwierenga:
i made some changes:
- use a `Step` class instead of a dict for more type safety. dicts are easy to confuse
- use unittest.IsolatedAsyncioTestCase, redirect_stdout to reduce dependency on pytest
- fix the tests by upgrading the dependency and refactoring the code slightly (opentrons is very annoying because they only support 3.10)

what is left are 2 open comments:
- use 0 indexing
- decide if we raise an error for lid/block target temperature when not in a cycle

i will also add a small example to the README, we have this for every machine module

rickwierenga:
@hazlamshamin mind if change the two things from above: raising an error when there is not target temp and use 0 indexing? very eager to merge this!

hazlamshamin:
yes please!!! would appreciate! im so sorry, this week and next packed with main works. Thanks :))  i can run my test script afterwards to final confirm still works, especially the run pcr profile since it depends on the indexing

rickwierenga:
thank **you**!

rickwierenga:
> i can run my test script afterwards to final confirm still works, especially the run pcr profile since it depends on the indexing

if you would do that that would be great! just lmk if it works and i can merge this branch :)

edit: since the tests pass, physical reality testing _shouldn't_ be necessary (otherwise tests are not sufficient), but a good idea to try anyway and validate the tests!

hazlamshamin:
Tested physically and checked that the PCR actually finish till last steps queued. I think good to go!

rickwierenga:
amazing! thank you so much for developing & testing all this! a great addition to PLR :)

rickwierenga:
i like to make twitter posts for publicity around big updates, do you have an account that i can tag?

hazlamshamin:
it's @hzlmshmn :D

hazlamshamin:
> amazing! thank you so much for developing & testing all this! a great addition to PLR :)

to more new backends whatever new lab instruments we have in this lab (waiting for our cuvette + 96-well HT electroporator) 🚀

PR #616, opened 2025-07-23
Support negative x direction for 96 head

rickwierenga:
## Summary
- handle negative X when using the 96 channel head
- allow negative X in `move_core_96_head_to_defined_position`

## Testing
- `make format`
- `make test`

------
https://chatgpt.com/codex/tasks/task_e_6881470f0e90832b899120d6a917f18e

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -2188,7 +2188,7 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
     position = tip_spot_a1.get_absolute_location() + tip_spot_a1.center() + pickup.offset
     z_deposit_position += round(pickup.offset.z * 10)
 
-    x_direction = 0 if position.x > 0 else 1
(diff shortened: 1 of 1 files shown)


rickwierenga:
@BioCam please confirm if this works for your use case

BioCam:
`move_core_96_head_to_defined_position` appears to not allow the movement into a negative x direction.
The atomic commands all seem to work fine (though I've done limited testing).

The firmware command does not seem to have an `x_direction` argument?

But we do know that the firmware _can_ (sometimes?) accept other commands arguments, so I'd like to test whether at the firmware level we can use `send_command` with a specified `x_direction` will work.
At the Python level it currently raises an error because the Python method does not contain this argument.

rickwierenga:
image
it does have that argument according to the documentation, and looking at the code it is actually implemented.

let me verify this on one of our machines.

rickwierenga:
this code works for me

```
await backend.move_core_96_head_to_defined_position(x=-10, y=200)
```

the default values for y and z were all illegal (0), that is why you got the 32 parameter out of range error. when setting them correctly (and i just updated the default z and removed xy) it works fine

rickwierenga:
minimum is 271

PR #617, opened 2025-07-23
add byonoy luminescence plate reader

rickwierenga:
absorbance will follow soon

todo:
- [ ] resource model
- [ ] perform the check whether this machine does lum or abs (same firmware api)
- [ ] status checks: read out orientation, temperature, humidity, device info, etc.
- [ ] tests

### DIFF
--- a/pylabrobot/plate_reading/__init__.py
+++ b/pylabrobot/plate_reading/__init__.py
@@ -1,4 +1,5 @@
 from .biotek_backend import Cytation5Backend, Cytation5ImagingConfig
+from .byonoy import Byonoy
 from .clario_star_backend import CLARIOStarBackend
 from .image_reader import ImageReader
 from .imager import Imager
--- a/pylabrobot/plate_reading/byonoy.py
+++ b/pylabrobot/plate_reading/byonoy.py
@@ -1,0 +1,229 @@
+import asyncio
+import struct
+import threading
+import time
+from typing import List, Optional
+
+from pylabrobot.io.hid import HID
+from pylabrobot.plate_reading.backend import PlateReaderBackend
+from pylabrobot.resources.plate import Plate
+
+
+class Byonoy(PlateReaderBackend):
+  """An abstract class for a plate reader. Plate readers are devices that can read luminescence,
+  absorbance, or fluorescence from a plate."""
+
+  def __init__(self) -> None:
+    self.io = HID(vid=0x16D0, pid=0x119B)
+    self._background_thread: Optional[threading.Thread] = None
+    self._stop_background = threading.Event()
+    self._ping_interval = 1.0  # Send ping every second
+    self._sending_pings = True  # Whether to actively send pings
+
+  async def setup(self) -> None:
+    """Set up the plate reader. This should be called before any other methods."""
+
+    await self.io.setup()
+
+    # Start background keep alive messages
+    self._stop_background.clear()
+    self._background_thread = threading.Thread(target=self._background_ping_worker, daemon=True)
+    self._background_thread.start()
+
+  async def stop(self) -> None:
+    """Close all connections to the plate reader and make sure setup() can be called again."""
+
+    # Stop background keep alive messages
+    self._stop_background.set()
+    if self._background_thread and self._background_thread.is_alive():
+      self._background_thread.join(timeout=2.0)
+
+    await self.io.stop()
+
+  def _background_ping_worker(self) -> None:
+    """Background worker that sends periodic ping commands."""
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
+
+    try:
+      loop.run_until_complete(self._ping_loop())
+    finally:
+      loop.close()
+
+  async def _ping_loop(self) -> None:
+    """Main ping loop that runs in the background thread."""
+    while not self._stop_background.is_set():
+      try:
+        # Only send ping if pings are enabled
+        if self._sending_pings:
+          # Send ping command
+          cmd = "40000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008040"
+          print("> " + cmd + " (background ping)")
+          await self.io.write(bytes.fromhex(cmd))
+      except Exception as e:
+        print(f"Error in background ping: {e}")
+
+      # Wait for the ping interval or until stop is requested
+      self._stop_background.wait(self._ping_interval)
+
+  def _start_background_pings(self) -> None:
+    self._sending_pings = True
+
+  def _stop_background_pings(self) -> None:
+    self._sending_pings = False
+
+  async def _read_until_empty(self, timeout=30):
+    data = b""
+    while True:
+      chunk = await self.io.read(64, timeout=timeout)
+      if not chunk:
+        break
+      else:
+        print("< ", chunk.hex())
+      data += chunk
+
+      if chunk.startswith(b"\x70"):
+        await self.io.write(
+          bytes.fromhex(
+            "20007000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
+          )
+        )
+    return data
+
+  async def _wait_for_response(self, timeout=30):
+    time.sleep(1)
+    data = b""
+    t0 = time.time()
+    while True:
+      data += await self._read_until_empty(timeout=timeout - (time.time() - t0))
+      if len(data) > 64:
+        break
+      if time.time() - t0 > timeout:
+        print("Timeout waiting for response")
+        return data
+      time.sleep(0.1)
+    return data
+
+  async def open(self) -> None:
+    raise NotImplementedError(
+      "byonoy cannot open by itself. you need to move the top module using a robot arm."
+    )
+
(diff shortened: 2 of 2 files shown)


pylabrobot/plate_reading/byonoy_backend.py
```diff
@@ -0,0 +1,423 @@
...
+    finally:
+      loop.close()
+
+  async def _ping_loop(self) -> None:
+    """Main ping loop that runs in the background thread."""
+    while not self._stop_background.is_set():
+      if self._sending_pings:
+        # fluor uses this. however, the absorbance one also works for fluor.
+        # cmd = "40000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008040"
+        cmd = "0040000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000040"
```
byonoy-dne: This is the correct heartbeat message common to all devices. While this is technically optional, I do highly recommend keeping it in. You should however also send the command to disable the heartbeat when shutting down. It's the same message, just replace the 1 with a 0. Otherwise, the device hits a watchdog timer and reboots a couple of seconds after PLR drops the connection. Nothing critical, just nicer because otherwise users might get the Windows device disconnect/connect chimes.

pylabrobot/plate_reading/byonoy_backend.py
```diff
@@ -0,0 +1,423 @@
...
+      bytes.fromhex(
+        "0050000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000040"
+      )
+    )
+    await self.send_command(
+      bytes.fromhex(
+        "0000020700000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008040"
+      )
+    )
+    # above this it checks if absorbance is supported. which command?
```
byonoy-dne: None of them. The commands ask the device, what subset of the communication protocol it supports, what version of the protocol it understands, and finally what the content of data field 7 is. Which is QUEUE_DEPTH, the supported number of parallel requests the device can handle. For your purpose the USB VID/PID should be enough to assume the device supports the associated measurement method.

pylabrobot/plate_reading/byonoy_backend.py
```diff
@@ -0,0 +1,423 @@
...
+        "000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000040"
+      )
+    )
+    await self.send_command(
+      bytes.fromhex(
+        "0040000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008040"
+      ),
+      wait_for_response=False,
+    )
+    self._stop_background_pings()
```
byonoy-dne: This command sequence here is:
* Status request
* Absorbance measurement trigger
* Heartbeat
* Status request
* Status request
* Read data field 7
* Absorbance measurement trigger
* Heartbeat

Most of this should not be necessary. You ignore the status and data field replies, you have the heartbeat running in a background process anyway, and one measurement trigger should be enough.

pylabrobot/plate_reading/byonoy_backend.py
```diff
@@ -0,0 +1,423 @@
...
+    while True:
+      # read for 2 minutes max
+      if time.time() - t0 > 120:
+        break
+
+      chunk = await self.io.read(64, timeout=30)
+
+      if (
+        bytes.fromhex(
+          "30000000000034000000526573756c74732020546f7020526561646f75740a0a0000000000000000000000000000000000000000000000000000000000000000"
```
byonoy-dne: This is a diagnostic message: "Results  Top Readout"

Diagnostics are subject to change and this one in particular changes depending on the physical device orientation during readout. It would be better to listen for the `REP_LUM96_MEASUREMENT_IN` (`0x0600`) messages instead.

pylabrobot/plate_reading/byonoy_backend.py
rickwierenga: thanks for clarifying!

what's the difference with the last 2 bytes: 0x8040 vs 0x0040?

pylabrobot/plate_reading/byonoy_backend.py
rickwierenga: ah interesting, I had seen in your sdk there is a check for whether absorbance is supported, but this must be done by pidvid then. thank you for sharing what those methods mean.

It should be safe to delete them here right?

pylabrobot/plate_reading/byonoy_backend.py
rickwierenga: Interestingly this is exactly what I observe when packet sniffing your sdk (and I believe I only read data once there), but very much down to make the implementation more efficient :) thanks!

pylabrobot/plate_reading/byonoy_backend.py
rickwierenga: > REP_LUM96_MEASUREMENT_IN (0x0600)

where in the message would I expect to see 0x0600?

is there an overview of all messages that are supported?

pylabrobot/plate_reading/byonoy_backend.py
rickwierenga: answering my own question: you mentioned the last 2 bytes are routing information and it's best to set them to 0

rickwierenga:
@BioCam could you please test? I have verified both to work on our machines

BioCam:
Could you please confirm that this is the integration architecture?

Screenshot 2025-11-28 at 17 00 00

1. What is the `byonoy_absorbance_adapter`? Is it the same as what the OEM's manual calls "SBS Adapter"?
2. Is the term "base" here used to refer to two different things?:
(i) `_ByonoyBase` the superclass to the machine backends - application=machine communication -, and
(ii) `ByonoyBase` an aggregate `Resource` class that stores `child_location` for both the plate and the illumination_unit - application=resource modelling.
This is very confusing, why don't we adapt the terminology from the OEM for everything physical?:
2d32f4fafe7bfdc01af8564f9fcfd73ba8e78b37_2_590x499

pylabrobot/plate_reading/byonoy/byonoy.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+def byonoy_absorbance_adapter(name: str) -> ResourceHolder:
+  return ResourceHolder(
+    name=name,
+    size_x=127.76,  # measured
+    size_y=85.59,  # measured
+    size_z=14.07,  # measured
+    child_location=Coordinate(
+      x=-(138 - 127.76) / 2,  # measured
```
BioCam: Where does the 138 mm come from?

pylabrobot/plate_reading/byonoy/byonoy.py
```diff
@@ -0,0 +1,152 @@
...
+      size_y=95.7,
+      size_z=27.7,
+    )
+
+    self.plate_holder = _ByonoyAbsorbanceReaderPlateHolder(
+      name=self.name + "_plate_holder",
+      size_x=127.76,
+      size_y=85.59,
+      size_z=0,
+      child_location=Coordinate(x=(138 - 127.76) / 2, y=(95.7 - 85.59) / 2, z=27.7),
```
BioCam: Is 27.7 mm the distance from the bottom of the "SBS Adapter" to the surface that the plate sits on?

I am getting:
- SBS_adapter.size_z = 17 mm
- detection_unit.size_z = 20 mm
- detection_unit.child_location.z = 16 mm

BioCam: i.e. I am getting the exact same dimensions for the dimensions which are specified in the manual:

Screenshot 2025-11-28 at 17 11 07

BioCam:
Are measurements on this machine taken to 5 decimal places but the OEM's manual states the machine has a resolution to 3 decimal places?

BioCam:
explaination to [f9bca03](https://github.com/PyLabRobot/pylabrobot/pull/617/commits/f9bca03a3eae67a37772f28b6205c71e50435cea)

when going through the setup before, everything worked but when performing an absorbance measurement the firmware would speak to me: saying "Please initialise before measuring!", and fails to retrieve a measurement. 

PLR firmware/io log:
Screenshot 2025-12-01 at 16 26 33

... what a fun - and incredibly direct firmware - love it!

I noticed this is not happening IF I have used the app first -> I traced the issue to the fact that the app performs a specific reference measurement first, the machine appears to remember that measurement until reboot.

I added a PLR version of this measurement and refactored the measurement command to enable reference and sample measurements separately.

rickwierenga:
I deleted the new format for returning data which we should update in a future PR.

for now, you can use this function:

```python
data = reshape_2d(list(range(96)), (8, 12))  # example

well_id2value = {}
for col in range(len(data)):
  for row in range(len(data[0])):
    well = plate.get_item((col, row))
    well_id2value[well.get_identifier()] = data[col][row]
well_id2value
```

BioCam:
> I deleted the new format for returning data which we should update in a future PR.

Okay, did you update the tutorial to ensure it still works?

PR #619, opened 2025-07-24
PreciseFlex Arm Driver

miikee:
**Untested Driver**

This adds:
- ArmBackend base class
- PreciseFlexBackend
- TCP(IOBase) class
- Arm front end class

Docs haven't been updated yet.
There's a warning included in the PreciseFlexBackend docstring that it's untested.

### DIFF
--- a/docs/resources/library/opentrons.md
+++ b/docs/resources/library/opentrons.md
@@ -53,3 +53,8 @@ Unfortunately, most of the other labware (plates) is missing information that is
 - `opentrons_96_aluminumblock_generic_pcr_strip_200ul`
 - `opentrons_96_aluminumblock_nest_wellplate_100ul`
 - `opentrons_96_well_aluminum_block`
+
+
+## Plate Adapters
+
+- `opentrons_96_deep_well_temp_mod_adapter`
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -2469,6 +2469,7 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
     tip_pickup_method: Literal["from_rack", "from_waste", "full_blowout"] = "from_rack",
     minimum_height_command_end: Optional[float] = None,
     minimum_traverse_height_at_beginning_of_a_command: Optional[float] = None,
+    experimental_alignment_tipspot_identifier: str = "A1",
   ):
     """Pick up tips using the 96 head.
 
@@ -2490,6 +2491,7 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
       tip_pickup_method: The method to use for picking up tips. One of "from_rack", "from_waste", "full_blowout".
       minimum_height_command_end: The minimum height to move to at the end of the command.
       minimum_traverse_height_at_beginning_of_a_command: The minimum height to move to at the beginning of the command.
+      experimental_alignment_tipspot_identifier: The tipspot to use for alignment with head's A1 channel. Defaults to "tipspot A1".  allowed range is A1 to H12.
     """
 
     if isinstance(tip_pickup_method, int):
@@ -2504,8 +2506,6 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
 
     assert self.core96_head_installed, "96 head must be installed"
 
-    tip_spot_a1 = pickup.resource.get_item("A1")
-
     prototypical_tip = next((tip for tip in pickup.tips if tip is not None), None)
     if prototypical_tip is None:
       raise ValueError("No tips found in the tip rack.")
@@ -2518,19 +2518,21 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
     fitting_depth = prototypical_tip.fitting_depth
     tip_engage_height_from_tipspot = tip_length - fitting_depth
 
-    # Tip size–based z-adjustment
-    h_tip = self._get_hamilton_tip([tip_spot_a1])
-    if h_tip.tip_size == TipSize.LOW_VOLUME:
+    # Adjust tip engage height based on tip size
+    if prototypical_tip.tip_size == TipSize.LOW_VOLUME:
       tip_engage_height_from_tipspot += 2
-    elif h_tip.tip_size != TipSize.STANDARD_VOLUME:
+    elif prototypical_tip.tip_size != TipSize.STANDARD_VOLUME:
       tip_engage_height_from_tipspot -= 2
 
     # Compute pickup Z
-    tip_spot_z = tip_spot_a1.get_location_wrt(self.deck).z + pickup.offset.z
+    alignment_tipspot = pickup.resource.get_item(experimental_alignment_tipspot_identifier)
+    tip_spot_z = alignment_tipspot.get_location_wrt(self.deck).z + pickup.offset.z
     z_pickup_position = tip_spot_z + tip_engage_height_from_tipspot
 
     # Compute full position (used for x/y)
-    pickup_position = tip_spot_a1.get_location_wrt(self.deck) + tip_spot_a1.center() + pickup.offset
+    pickup_position = (
+      alignment_tipspot.get_location_wrt(self.deck) + alignment_tipspot.center() + pickup.offset
+    )
     pickup_position.z = round(z_pickup_position, 2)
 
     self._check_96_position_legal(pickup_position, skip_z=True)
@@ -2564,12 +2566,13 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
     drop: DropTipRack,
     minimum_height_command_end: Optional[float] = None,
     minimum_traverse_height_at_beginning_of_a_command: Optional[float] = None,
+    experimental_alignment_tipspot_identifier: str = "A1",
   ):
     """Drop tips from the 96 head."""
     assert self.core96_head_installed, "96 head must be installed"
 
     if isinstance(drop.resource, TipRack):
-      tip_spot_a1 = drop.resource.get_item("A1")
+      tip_spot_a1 = drop.resource.get_item(experimental_alignment_tipspot_identifier)
       position = tip_spot_a1.get_location_wrt(self.deck) + tip_spot_a1.center() + drop.offset
       tip_rack = tip_spot_a1.parent
       assert tip_rack is not None
--- a/pylabrobot/resources/opentrons/adapters.py
+++ b/pylabrobot/resources/opentrons/adapters.py
@@ -1,0 +1,9 @@
+from pylabrobot.resources.carrier import PlateHolder
+from pylabrobot.resources.opentrons.load import load_ot_plate_holder
+
+
+def opentrons_96_deep_well_temp_mod_adapter(name: str) -> PlateHolder:
+  z_offset = 5.1
+  return load_ot_plate_holder(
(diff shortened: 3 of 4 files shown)


miikee (commented):
Rebased to main.  
- Updated with PreciseFlexBackendAPI, PreciseFlexBackend, ArmBackend, and Arm.  Including integration tests with robot (marked with a pytest.mark.hardware).  
- Pytest.ini updated to ignore 'hardware' marked tests. 
- Updated TCP base class since original testing.  So this may need to be retested, but should still work.

pylabrobot/arms/coords.py
```diff
@@ -0,0 +1,29 @@
...
+from typing import Optional
+
+
+class ElbowOrientation(Enum):
+  RIGHT = "right"
+  LEFT = "left"
+
+
+@dataclass
+class JointCoords:
```
BioCam: What does this refer to, what are the units of each attribute?

Is this the motor position of each joint?
i.e. not a Coordinate in 3D space but a "1D" motor position values

rickwierenga:
@miikee could you please enable "allow edits by maintainers" so I can work on this PR?

miikee:
@rickwierenga just sent you an invite for the branch.  Let me know if it doesn't work.

rickwierenga:
thank you, this works.

do you know if there is a way to convert between JointCoords and CartesianCoords using their API? like do they expose a kinematics model?

miikee:
There's a method that converts Cartesian and Joint in the GPL code on the robot, but it's not exposed through the TCS server.  We'd need to write some custom GPL to expose it; and the user would also need to upload it.

rickwierenga:
noob question: how do I define a profile?

with

```python
await arm.backend.move_to(CartesianCoords(location=Coordinate(x=x, y=y, z=z), rotation=Rotation(x=roll, y=pitch, z=yaw)))
```

I get `PreciseFlexError: PreciseFlexError -2817: *Undefined profile*`

and with

```python
profile_id = 20
api = arm.backend.api
await api.set_motion_profile_values(
    profile_id,
    speed=25,
    speed2=0,
    acceleration=10,
    deceleration=10,
    acceleration_ramp=0.5,
    deceleration_ramp=0.5,
    in_range=10,
    straight=False,
)
```

I get `PreciseFlexError: PreciseFlexError -757: Object not instantiated. An object is being assigned to on the left-hand side of an equal sign or is being referenced in an expression and the object value block has not been allocated. To correct this problem, use the "New" qualifier in the DIM statement that declared the object to allocate its value block. - `

rickwierenga:
what fixed this: setting the port to `10100` instead of `10000`

(nmap sometimes shows 10000 as open, and has never showed 10100 as open, but specifying it in the api did work)

rickwierenga:
@miikee what are your thoughts on moving the api code into the backend? a lot of the code is just passing data. is there a reason we need that layer of abstraction?

miikee:
@rickwierenga Do you mean combining the class PreciseFlexBackendApi and PreciseFlexBackend?  My idea for keeping them separate was to make PreciseFlexBackend more human readable and to keep PreciseFlexBackendApi 1:1 with the TCS documentation.

rickwierenga:
firmware commands are a common pattern in PLR 😅  and I think it almost always makes sense to have one method per firmware command.

in some cases I think it also makes sense to have wrappers around those raw firmware-command-methods, to make them more user-friendly or have a version compatible with an abstract backend

but in cases like this where the firmware api is actually nice, it introduces more complexity than it reduces imo. for example, to home you have to read Arm.home -> PreciseFlex400Backend.home -> PreciseFlexBackendApi.home to see it's just a string `"home"` where it could be the more minimal Arm.home -> PreciseFlex400Backend.home (which is a common pattern in PLR). it would still be 1:1 with the TCS docs, but that then lives in the PreciseFlex400Backend. if needed we can add some more methods to PreciseFlex400Backend that then call those firmware-methods

miikee:
@rickwierenga combining the two would be fine if you think it'll lower complexity.

miikee:
@rickwierenga I also wanted to mention a complication with the pick/place/approach methods.  These work by setting a station location in the robot's memory.  Then the robot grabs from from that location.  But it grabs differently depending what access type is set to that station.  This is an internal flag it keeps with each station.  

- access type 1 = vertical (like a deck plate pad)
- access type 2 = horizontal (like a hotel nest pad)

So the movement changes for each:

- access type 1 place - the arm travels above the location and drops down
- access type 2 place - the arm travels in front of the location and makes a "box" movement - up, forward, down, open gripper

Depending on access type, the z-clearance changes meaning:

- access type 1 - z-clearance = distance above
- access type 2 - z-clearance = distance in front

So this can be a bit confusing... plus theres a few other access parameters that can be changed.  I've added a commit here to fix this with a AccessConfig dataclass to describe how to access the plate.  Let me know your thoughts.

rickwierenga:
awesome, thank you!

rickwierenga:
- is it correct that ElbowOrientation has a different integer encoding between pf400 and pf3400?
- can I specify for close and open gripper which position it goes to? ie plate width. in the code I see `gripper 1` and `gripper 2` for open and close. should I use `grasp_plate` instead of `close_gripper` or `pick_plate`?

miikee:
Im not too sure off the top of my head.  I was working with the pf3400 and setup the pf400 as a bit of place holder to work with in the future.  I can't remember if I looked up the info when I did that.  I'll need to review some things to get the answer to you.

With the PF3400, I don't think it has a grip width.  It's either open or close.  I'm not sure if this is true with the PF400 or not.  I'll look into it a bit more tonight or tomorrow once I'm home.

rickwierenga:
with the pf400 (which we have) I have been using `grasp_plate` and it works great. I am thinking about using that command for PLR. I don't think I would be able to use the other close gripper options because I can't specify the width for them.

miikee:
I looked through the docs and code some but can't find the definitive answer I'm looking for to answer your questions about the gripping and the ElbowOrientation.  My thoughts are that the ElbowOrientation was different due to a difference in the number of joints between PF3400 and PF400.  And that the use of the commands between grip width and binary open/close is more dependent on the gripper attached than the actual model of arm.  I'm going to be on-site with them Monday, so it might just be best to ask them.

miikee:
I see the ElbowOrientation swap you were talking about now.  That was wrong, both arms use 1 = LEFT and 2 = RIGHT. I just made a commit to fix that.

rickwierenga:
- removed `where` and `get_location` because `get_joint_position` and `get_cartesian_position` and `get_location_angles` and `get_location_xyz` can do all of that. I didnt see the point in having more complicated methods

rickwierenga:
also in PLR we now always have joints lists be in the format `[rail, base, shoulder, elbow, wrist, gripper]` so we can reason about lists irrespective of robot state. this makes the code much simpler. it also simplifies cases where users add a rails after already having code for a particular robot / share code between robots with/without rails. obviously, the firmware strings still follow the pf convention of ignoring the rails when not installed...

rickwierenga:
when gripping a plate using `pick_plate`, I get this:

```
  An axis has been commanded to accelerate too quickly or move too fast and does not have the power to perform the operation.  If necessary, this can be confirmed by datalogging the "Compensator output torque" (DataID 12304) and the "Position tracking error" (DataID 12320) for the axis in question.  If this is the source of the envelope error, the position error will increase significantly when the motor torque saturates at its maximum value for several tens of milliseconds.
  The axis has gone unstable due to a hardware failure or some other error.  This can be confirmed by listening for an audible noise or by datalogging the "Velocity tracking error" (DataID 12321) for the axis in question and looking for high frequency oscillations.
  An axis may have crashed into an obstacle and is unable to move to the specified position.
  The "Software envelope error limit" (DataID 10302) may be set to too small of a value. - 
```

the reason is probably that it tries to close too far, as the gripper closes all the way when I run it without a plate. strangely, `_set_grip_close_pos` does not make a difference (it does change the position for the `closer_gripper` command)

two questions:
- how to make it pick up a plate using `pick_plate`?
- how to recover from this error? `hp 1` and `attach 1` don't work.

rickwierenga:
re my second question:

I get the following error when attempting any command after the crash:

```
PreciseFlexError -1009: No robot attached. A function or method was executed that required that a robot be ATTACHED.  ...
```

when I attach, I get this:

```
PreciseFlexError: PreciseFlexError -1007: Robot not ready to be attached. An Auto Execution task or a GPL project has attempted to gain control of a robot by executing a Robot.Attach method or similar function, but the specified robot is not in a state where it can be attached. If this was generated by a GPL project, it might indicate: That the system is configured to execute DIOMotion blocks instead of a GPL motion program (DataID 200) or "Auto start auto execute mode" (DataID 202) is not set to TRUE. If so, please check the Setup>Startup Configuration web page to verify the current setup. - 
```

I have to power cycle by physically powering off and on the robot.

rickwierenga:
> how to make it pick up a plate using pick_plate?

turns out, it pulls information not from the `_set_grip_close_pos` but from `grasp_data` (`set_grasp_data` and `get_grasp_data`). pushing a nice api in a bit.

miikee:
I havent run into this error with the pick plate before.  But it maybe that the speed in your profile is too fast.  So that may need to be lowered in the speed profile because some joints may not be able to keep up.

The other part is I add in a setting to keep the wrist between +/- 180° to prevent some wild spinning of the wrist.  It's a bitwise setting called GPL_Single flag (0x1000) in commit a332ab511052ea532c9bf0c164eeb5f494ee66a7

I think this including this would be useful because the wrist can get a little wild sometimes... but it may be the cause of this problem. I'd check the profile setting first though.  I think the pick always just runs on profile 1.

rickwierenga:
I think the reason is our end effector has prongs, which makes the default width different than yours. It is my first time using `pick_plate`.

miikee:
If setup(), which has the hp 1 and attach 1 isn't working, it may have gotten hung up with the GPL code running on the system.

Go to the web application, I think its 192.168.0.1. Then select admin button and then click the top option in the side panel... I think it maybe be pendant or operator or similar.  And in there I think you need to stop the GPL program that's loaded then need to load it again.  I think there's a stop button and then you need to load it again. (You can also do this from Brooks' GDS app, if you have that)

I dont have my laptop or the bot right now... so this is from memory.  And a bad memory too.

rickwierenga:
we are mostly ready I think!

I just wrote a small tutorial showing how to do the basic things

one thing still to discuss:
- there are many duplicitous methods, such as pick_plate and its lower level function `grasp_plate`, and `place_plate` and `release_plate`. I think we should consider removing the lower level methods because I am not sure who would use those, and it makes the api ambiguous. do I use `place_plate` or `release_plate`? if there's a reason to keep them, maybe make them private? they are not used by PLR, and probably won't be by users.

I will do the resource modeling (ie picking up resources instead of "pick up from coordinate") in a future PR.

miikee:
The only reason I can think of for keeping those functions is to have all endpoints of the TCS API exposed in python in case there is a difference and a user is specifically looking for one of them.  But I think its probably just better to remove them in favor of a non-ambiguous API.  

Brooks' TCS GPL code is open source if a user wants to explore things more. There's a few things not exposed through the TCS interface.  An example is the Kinematic Solve function for cartesian <-> joint conversions.  It exists on the robot, but isn't exposed through the TCS program that ships with the robot.

This is where Brooks' has a lot of their software - Guidance Development Studio (GDS), TCP Command Server (TCS), etc: 
https://www.brooks.com/support/brooks-preciseflex-support/  - But they require you to reach out to get a password for access: support_preciseflex@brooks.com

rickwierenga:
thank you so much for this massive effort!

miikee:
Thank you! Looking forward to collaborating on more devices soon!

PR #622, opened 2025-07-25
Reset Cytomat error register before raising errors

rickwierenga:
## Summary
- reset the Cytomat error register whenever the device reports an error

## Testing
- `make format`
- `make test`

------
https://chatgpt.com/codex/tasks/task_e_6883cfb207bc832bb06ba97f97d61540


rickwierenga:
to prevent the same error from being raised multiple times

PR #628, opened 2025-07-28
Add block status and refine lid/block status handling

hazlamshamin:
- Added `get_block_status` to ThermocyclerBackend and its implementations.                                                                                                     
- Refined `get_lid_open` and `get_lid_temperature_status` methods.                                                                                                             
- Updated Opentrons backend to handle missing cycle/step data gracefully.

### DIFF
--- a/pylabrobot/thermocycling/backend.py
+++ b/pylabrobot/thermocycling/backend.py
@@ -57,6 +57,14 @@ class ThermocyclerBackend(MachineBackend, metaclass=ABCMeta):
     """Return ``True`` if the lid is open."""
 
   @abstractmethod
+  async def get_lid_temperature_status(self) -> str:
+    """Get the lid temperature status."""
+
+  @abstractmethod
+  async def get_block_status(self) -> str:
+    """Get the block status."""
+
+  @abstractmethod
   async def get_hold_time(self) -> float:
     """Get remaining hold time in seconds."""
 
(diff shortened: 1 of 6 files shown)


rickwierenga:
thank you for this PR, that was fast!

if i understand it correctly:
- we are renaming get_lid_status -> get_lid_open. this returns a bool because there are only two possible values
- we are adding a new method get_lid_temperature_status. This indicates the previous implementation had a small bug because it was checking the `lidStatus` key instead of `lidTemperatureStatus`? (not trying to play gotcha, just trying to fully understand and document this change :))

for block, i saw you added a method `get_block_status` that was not there before. how do you feel about renaming this to `get_block_temperature_status`? Or i think more ideally `get_lid_temperature_status`->`get_lid_status` (the old name but new data) so they are consistent? I think status is more general and nice and temperature_status personally.

hazlamshamin:
exactly, yeap yeap. for the naming, i think it depends on whether we wanna use get_lid_open or get_lid_status for the open/close. despite opentrons use lidStatus for the open/closed, i think using the get_lid_open is more intuitively understood. so then, we can either use pair 1: get_block_temperature_status, get_lid_temperature_status, or straight away pair 2: get_block_status, get_lid_status. idm any, that is more understandable for developers.

rickwierenga:
I think the second option (get_x_status) is objectively the best way to do it. It is slightly confusing in the context of an OT since they made the 'wrong'/different decision, but the goal in PLR is to support many thermocyclers. I don't think we should import bad decisions just to be consistent with one device. I will update your branch here, merge it into the other one, then merge that into main.

Thank you again for the quick response and work!

PR #639, opened 2025-07-31
Update a4S Docs

BioCam:


rickwierenga:
again thank you for the PR!

note: instead of note in a custom indention block with emoji, you can use admonitions to use the automatic rendering for warnings/notes/etc. See all of them documented here: https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/kitchen-sink/admonitions.html

image

PR #641, opened 2025-08-02
Update Temperature Controller Docs

BioCam:
Restructuring, and adding some WIP tags


docs/user_guide/01_material-handling/temperature-controllers/_temperature-controllers.rst
rickwierenga, line 1: i think we should avoid underscores in the filenames since these files will be on the docs website

docs/user_guide/01_material-handling/temperature-controllers/_temperature-controllers.rst
```diff
@@ -0,0 +1,127 @@
...
+=======================
+
+Temperature controllers are defined as **machines with one or both** of these **features**:
+
+- `heating` 
+- `actively cooling`
+
+a **material** or **enclosed volume** from a room-temperature baseline (≈20-25°C).
+
+Based on this definition machines that include temperature control but are not primarily temperature controllers, build on top of the temperature controller definition.
```
rickwierenga: i don't understand this sentence

docs/user_guide/01_material-handling/temperature-controllers/_temperature-controllers.rst
BioCam: Let's discuss; my reasoning in this and all other docs folders I created:

1) there can be many files in a folder with different starting letters, having the "overview" .rst always on the top is very helpful

2) I see the "overview" .rst file as a 'private' file that should never leave its folder. Thinking long-term, other files might come and go, but this one will always stay

If it is ugly in a URL we can remove it though but will have to update that scheme of not using a prefix "_" for all other overview rst files too :)

docs/user_guide/01_material-handling/temperature-controllers/_temperature-controllers.rst
BioCam: I guess this is how we know a human wrote it 😅 👀

Chat rewrote with some help to:
"Machines that include temperature control features but are not primarily temperature controllers themselves should extend or build upon the temperature controller definition."

docs/user_guide/01_material-handling/temperature-controllers/_temperature-controllers.rst
rickwierenga: in the folder structure yes, i get this argument

however, my primary concern here is the URL

rickwierenga: there is no "easy" way to do this as far as i can find

(you can "include" files in other files with desired name, but that's adding more complexity than it removes)

docs/user_guide/01_material-handling/temperature-controllers/_temperature-controllers.rst
rickwierenga: i like that

PR #652, opened 2025-08-13
Updated method reference in "Dropping tips" section of basic.ipynb

MichaelSkehan:
In the "Dropping Tips" section of the "Getting started with liquid handling on a Hamilton STAR" the text refers to the method `discard_tips()` when it should be `drop_tips()` for the subsequent example.


rickwierenga:
Thanks!!

PR #656, opened 2025-08-20
Fix broken doc links in README

yaoyuyang:
Fix the following links
* Installation
* Getting Started
* Contributing
* API Reference


rickwierenga:
thank you for find & quick fix!

PR #658, opened 2025-08-23
PlateReader.read_dye

rickwierenga:
concept: introducing convenience methods for reading specific dyes on plate readers


rickwierenga:
closing because this is not the approach we wanna take

see forum post: https://discuss.pylabrobot.org/t/rfc-introducing-convenience-methods-for-reading-specific-dyes-on-plate-readers/321

PR #659, opened 2025-08-23
Automate CLARIOstar & Hamilton HEPA fan VID:PID Search at Runtime

BioCam:
Hi everyone,

tiny PR to avoid head scratching about why PLR does not connect to machines:

Many machines' Vendor ID:Product ID (VID:PID) combos ([Understanding VID and PID in USB Devices](https://blog.stackademic.com/understanding-vid-and-pid-in-usb-devices-a183c6ec0d25)) are **not** automatically searched for by `pylibftdi`.

As a result, we have to manually tell Python to search for them.
PLR users might not be familiar with this requirement and might then incorrectly attribute failure to connect to their machines as a "bug of PyLabRobot" - we must fix that misconception before it arises.

In this PR, we propose adding the required information into the runtime's `USB_VID_LIST` and `USB_PID_LIST` in the specific backend's .py file.
This means the required information is only imported when needed and when needed import is automated without the user having to identify the IDs every time.

### DIFF
--- a/pylabrobot/only_fans/hamilton_hepa_fan_backend.py
+++ b/pylabrobot/only_fans/hamilton_hepa_fan_backend.py
@@ -1,14 +1,16 @@
 import asyncio
 
+# Make pylibftdi scan the Hamilton HEPA Fan VID:PID
+# appears as ID 0856:ac11 B&B Electronics Model USOPTL4
+from pylibftdi import driver
+
 from pylabrobot.io.ftdi import FTDI
 
 from .backend import FanBackend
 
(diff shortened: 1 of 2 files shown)


rickwierenga:
Thanks for PR!

This took me a long time to figure out when I integrated the plate reader. And it is very non-obvious and cumbersome for users to edit the installed library.

While it's normally not recommended to edit the internal of other libraries, in this case it's necessary since there is no good api for it. Clever

rickwierenga:
Could you please run make format to format

rickwierenga:
Thanks!

PR #660, opened 2025-08-24
Docs Update: add placeholder notebook for BMG Labtech CLARIOstar

BioCam:


BioCam:
The BMG Labtech CLARIOstar (Plus) PLR integration is currently being re-developed.

This PR sets the stage for this development:
Laying out a new docs structure to unambiguously declare the CLARIOstar's exposed machine features, their state, and the "how to use".

While different people are working on different machine feature integrations they can simultaneously document their integration in the dedicated section of the updated docs notebook.

pylabrobot/plate_reading/clario_star_backend.py
rickwierenga, line 1: let's keep a `CLARIOStarBackend` that will raise a depreciation error, similar to when we added `Backend`

rickwierenga:
nice changes, making it easy to document future features

for absorbance and luminescence, why not add the existing code as an example? it is already functional

BioCam:
You are right; absorbance and luminescence are "minimally functional" in PLR (i.e. basic reading is exposed but reading subsets of wells, orbital diameter, number of technical replicate measurements per well, ... are parameters not yet available).

I will add the example code of what is currently available back in and make a note of the feature parameters that we know of which are currently missing.

rickwierenga:
nice, thank you!

rickwierenga:
please format and this should be ready to merge

PR #669, opened 2025-09-11
Fix `request_iswap_rotation_drive_orientation`

BioCam:


rickwierenga:
Why not make PARKED_RIGHT part of the enum?

BioCam:
> Why not make PARKED_RIGHT part of the enum?

Can do but if it is part of the enum one would expect it to work the same way as the other enums... but that is not the case - when used with `rotate` for example it would break the method?

rickwierenga:
We can raise an error in that case

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -7567,14 +7569,18 @@ class RotationDriveOrientation(enum.Enum):
...
+    PARKED_RIGHT = None
 
   async def rotate_iswap_rotation_drive(self, orientation: RotationDriveOrientation):
-    return await self.send_command(
-      module="R0",
-      command="WP",
-      auto_id=False,
-      wp=orientation.value,
-    )
+    if orientation.value:
```
rickwierenga: Why base this on the enum value? It is very implicit. Also the enum values typically don't matter. It would better to check 

```
if orientation in {not, allow}:
  raise
...
```

BioCam: The enum values are the reason for the existence of the enums, I thought? The vallues are what the firmware requires as input as in the example you highlighted.

We can of course make it more explicit though

PR #670, opened 2025-09-11
add `BioER_96_wellplate_Vb_2200ul`

har1eyk:
BioER makes a deepwell plate similar to the NEST deepwell plate. It's 2.2mL with a v-bottom. I've attached a pdf. 

[BioER_96w_deepwell_plate.pdf](https://github.com/user-attachments/files/22283980/BioER_96w_deepwell_plate.pdf)

### DIFF
--- a/pylabrobot/thermocycling/opentrons_backend_usb.py
+++ b/pylabrobot/thermocycling/opentrons_backend_usb.py
@@ -1,0 +1,336 @@
+# For direct control of the Opentrons Thermocycler to any USB port.
+# Does not require an Opentrons liquid handler to use.
+
+import asyncio
+from typing import List, Optional
+
+from pylabrobot.thermocycling.backend import ThermocyclerBackend
+from pylabrobot.thermocycling.standard import (
+  BlockStatus,
+  LidStatus,
+  Protocol,
+)
+
+try:
+  import serial.tools.list_ports
+  from opentrons.drivers.thermocycler import ThermocyclerDriverFactory
+  from opentrons.drivers.thermocycler.abstract import AbstractThermocyclerDriver
+  from opentrons.drivers.types import ThermocyclerLidStatus
+
+  USE_OPENTRONS_DRIVER = True
+  _import_error = None
+except ImportError as e:
+  USE_OPENTRONS_DRIVER = False
+  _import_error = e
+
+
+async def set_temperature_no_pause(
+  driver,
+  temperature: float,
+  hold_time_seconds: Optional[float],
+  hold_time_minutes: Optional[float],
+  ramp_rate: Optional[float],
+  volume: Optional[float],
+) -> None:
+  """Set temperature without waiting for completion."""
+  seconds = hold_time_seconds if hold_time_seconds is not None else 0
+  minutes = hold_time_minutes if hold_time_minutes is not None else 0
+  total_seconds = seconds + (minutes * 60)
+  hold_time = total_seconds if total_seconds > 0 else 0
+
+  if ramp_rate is not None:
+    await driver.set_ramp_rate(ramp_rate=ramp_rate)
+
+  await driver.set_plate_temperature(temp=temperature, hold_time=hold_time, volume=volume)
+
+
+async def wait_for_block_target(driver) -> None:
+  """Wait for block temperature to reach target."""
+  max_attempts = 300  # 5 minutes max wait (300 * 1 second)
+  attempt = 0
+
+  while attempt < max_attempts:
+    try:
+      plate_temp = await driver.get_plate_temperature()
+      if plate_temp.target is not None and abs(plate_temp.current - plate_temp.target) < 1.0:
+        break
+    except Exception as e:
+      if "invalid thermistor" in str(e).lower() or "error" in str(e).lower():
+        raise RuntimeError(f"Thermocycler hardware error: {e}")
+      print(f"Temperature check failed (attempt {attempt + 1}), retrying: {e}")
+    attempt += 1
+    await asyncio.sleep(1.0)
+  else:
+    raise TimeoutError(f"Temperature did not reach target within {max_attempts} seconds")
+
+
+async def execute_cycle_step(
+  driver,
+  temperature: float,
+  hold_time_seconds: float,
+  ramp_rate: Optional[float] = None,
+  volume: Optional[float] = None,
+) -> None:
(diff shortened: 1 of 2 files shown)


rickwierenga:
apologies for late response, I was traveling

the linked pdf seems to be unrelated 😅

har1eyk:
> the linked pdf seems to be unrelated 😅

The related pdf has been uploaded: 
[BioER_deepwell_plate.pdf](https://github.com/user-attachments/files/22346277/BioER_deepwell_plate.pdf)

pylabrobot/resources/bioer/plates.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def BioER_96_wellplate_Vb_2200ul(name: str) -> Plate:
+  """BioER Cat. No. BSH06M1T-A (KingFisher-compatible)
+  Spec: https://en.bioer.com/uploadfiles/2024/05/20240513165756879.pdf
+  """
+  return Plate(
+    name=name,
+    size_x=127.30,  # from spec
+    size_y=85.20,  # from spec
```
rickwierenga: the spec says 127.1 and 85

happy to go with these numbers if you measured them to be more accurate, but int hat case we would need a `# measured` comment

pylabrobot/resources/bioer/plates.py
```diff
@@ -0,0 +1,38 @@
...
+    size_z=44.2,  # from spec
+    lid=None,
+    model=BioER_96_wellplate_Vb_2200ul.__name__,
+    ordered_items=create_ordered_items_2d(
+      Well,
+      size_x=8.2,  # from spec (inner well width)
+      size_y=8.2,  # from spec (inner well length)
```
rickwierenga: the spec says 8.25

pylabrobot/resources/bioer/plates.py
```diff
@@ -0,0 +1,38 @@
...
+      size_z=42.4,  # from spec (well depth)
```
rickwierenga: I can't find this in the spec

pylabrobot/resources/bioer/plates.py
```diff
@@ -0,0 +1,38 @@
...
+      dx=9.5,  # measured (column pitch)
+      dy=9.5,  # measured (row pitch)
+      dz=1.2,  # from spec (plate height - well depth)
```
rickwierenga: `dz` refers to the bottom of the well _including_ the material z thickness

so i think this should be `44.2 - 42.4 - 0.8 = 1`, assuming 42.4 is correct (see above).

How did you get 1.2 given these numbers?

pylabrobot/resources/bioer/plates.py
har1eyk: Nice catch. Values updated according to spec.

pylabrobot/resources/bioer/plates.py
har1eyk: Values updated according to spec.

pylabrobot/resources/bioer/plates.py
har1eyk: size_z confirmed. Comment changed to `# measured`.

pylabrobot/resources/bioer/plates.py
har1eyk: Value updated. Optimal well dz = 6mm. `# measured`

rickwierenga: `6 + 0.8 + 42.4 = 49.2` is much bigger than the plate height of 44.2. I generally like taking pragmatic values of the spec if calipered, but this is a big difference. why?

when you move a pipette to `well.get_absolute_location(x='c', y='c', z='cavity_bottom')`, is it actually at the bottom?

har1eyk: Thanks for the help, Rick. I agree that there's a big difference. We expect dz should be 1 but I measured 6mm. I can't explain it. When I enter dz=1 my tips crash into the bottom and bend. I'm performing my measurements on the recently defined `Hamilton_MFX_plateholder_DWP_metal_tapped`. It has a pedestal size of -4.74. I wonder if that could be an issue. 

At dz=6mm, tips directed to `cavity_bottom` or to `liquid_height=[0]` both go to the same z-coord. This was confirmed for both 50ul and 1000ul filtered tips. 
```

a1 = dp1["A1"][0]
coord = a1.get_absolute_location(x="c", y="c", z="cavity_bottom")
print(f"Absolute coordinates of A1 cavity bottom: {coord}")
CHANNEL = 4

# move in X, Y, then Z
await lh.move_channel_x(CHANNEL, coord.x)
await lh.move_channel_y(CHANNEL, coord.y)
await lh.move_channel_z(CHANNEL, coord.z)
time.sleep(2)
# await lh.move_channel_z(CHANNEL, 270) # for 50ul filtered
await lh.move_channel_z(CHANNEL, 240) # for 1000ul filtered
time.sleep(2)
await lh.aspirate(dp1["A1"], vols=[0], use_channels=[4], liquid_height=[0], settling_time=[2])
```

20250923_185950000_iOS_400w_15fps_64c

rickwierenga: is it sitting on the bottom of the wells (e.g. side of plate hovers) or on the plate edges touching the little ridges around the pedestal?

(the reason I am hammering on this is it's really important to have accurate plate definitions because they power everything and mean we don't have to use offsets)

har1eyk: I appreciate your follow up. Yes, the plate wells sit on top of the pedestal. The plate side hovers. 

image

har1eyk:
Hi, Rick, are there any other checks I can perform? What do you recommend is my next step to confirm the accuracy of the parameters?

rickwierenga:
thanks for the ping. merging this now because it works for you. if someone else in the future has a problem or different measurement we can 'refactor' it then. please let me know if you ever run into a problem with it

& thanks for contributing this definition!

rickwierenga:
renamed to capital `BioER_96_wellplate_Vb_2200uL` with capital `uL` in https://github.com/PyLabRobot/pylabrobot/commit/e677f197e8b30df4ddfa4e9777c4dccf86bffcdd

BioCam:
@har1eyk, @rickwierenga 

Just noticed, shouldn't the `_Vb` be after the volume in the naming?

PR #673, opened 2025-09-16
usb backend for opentrons thermocycler standalone operation

ben-ray:


rickwierenga:
this is super nice

rickwierenga:
to confirm, this only works on python 3.10 right? like all opentrons code due them not wanting to support newer versions https://github.com/Opentrons/opentrons/issues/12006

ben-ray:
> to confirm, this only works on python 3.10 right? like all opentrons code due them not wanting to support newer versions [Opentrons/opentrons#12006](https://github.com/Opentrons/opentrons/issues/12006)

unfortunately, yes, you need a python 3.10 environment to run this.

we can re-implement the (very simple) [thermocycler gcode driver](https://github.com/Opentrons/opentrons/blob/edge/api/src/opentrons/drivers/thermocycler/driver.py) that is currently being imported from the opentrons repo to eliminate the requirement

rickwierenga:
what error do you get?

importing runs fine for me on 3.13.7

```
(env) rickwierenga@ricks-MacBook-Pro-2 opentrons % python api/src/opentrons/drivers/thermocycler/driver.py
(env) rickwierenga@ricks-MacBook-Pro-2 opentrons % echo $?   
0
(env) rickwierenga@ricks-MacBook-Pro-2 opentrons % python --version
Python 3.13.7
```

hazlamshamin:
hi @ben-ray and @rickwierenga , i tested on our Gen2. it works on my macbook (with usb c to usb3 adapter)!!!

initially it could not detect the thermocycler (Failed to initialize thermocycler backend: No Opentrons Thermocycler found with VID:PID 04d8:ed8c). then when i ran serial.tools.list_ports.comports() locally, i got "/dev/cu.usbmodem5034501420361: VID=0x0483 PID=0xed8d desc=Thermocycler FS". therefore when specifying the VID:PID for the port (--port /dev/cu.usbmodem5034501420361), it works!

So I think the Thermocycler that I has, has a different or newer STMicroelectronics bridge, than the legacy Microchip bridge, hardcoded as the ID in the backend.

So, I think, maybe we can put this new ID to widen the detection list, as well Falls back to any serial port whose description contains “thermocycler”?

edit: created PR on this PR
https://github.com/ben-ray/pylabrobot/pull/1

ben-ray:
> initially it could not detect the thermocycler (Failed to initialize thermocycler backend: No Opentrons Thermocycler found with VID:PID 04d8:ed8c). then when i ran serial.tools.list_ports.comports() locally, i got "/dev/cu.usbmodem5034501420361: VID=0x0483 PID=0xed8d desc=Thermocycler FS". therefore when specifying the VID:PID for the port (--port /dev/cu.usbmodem5034501420361), it works!
> 
> 
> 
> So I think the Thermocycler that I has, has a different or newer STMicroelectronics bridge, than the legacy Microchip bridge, hardcoded as the ID in the backend.

Thanks for testing on gen2 - we only have gen1 here so adding the gen2 PID is highly appreciated!

rickwierenga:
I like adding the vidpid @hazlamshamin found.

> as well Falls back to any serial port whose description contains “thermocycler”?

This is slightly sketchy since we don't know that's the OTTC for sure, but I do like presenting it as an option: when vidpid fails we can say "do you want to use `/dev/X`?" Note that this will only work on macOS. On Linux most serial devices are `/dev/ttyUSB<num>` and on Windows `Com<num>`.

I'm fine merging that in because it's nice, but it's not super crucial. Two vidpid for both versions should be sufficient.

rickwierenga:
on second thought, why not just default to the serial port that says thermocyler when available. it's unlikely someone has two thermocyclers connected to a computer, and it will increase the "just works" factor of PLR. Let's keep it in.

pylabrobot/thermocycling/opentrons_backend_usb.py
```diff
@@ -123,12 +129,17 @@ def __init__(self, port: Optional[str] = None):
...
         )
+      elif len(opentrons_ports) > 1:
+        available_ports = [p.device for p in opentrons_ports]
+        raise RuntimeError(
+          f"Multiple Opentrons Thermocyclers found: {available_ports}. Please specify the port explicitly."
+        )
+      else:
+        self.port = opentrons_ports[0].device
     else:
       self.port = port
```
rickwierenga: finding the port based on pidvid belongs in `setup` since it communicates with io, not in init

the port also does not have to be stored on self in that case

ben-ray:
I think its best to only establish connections to thermocyclers that are supported. I made a minimal change that enables multiple thermocyclers to be connected to the same computer while explicitly filtering out any thermocyclers not supported. This enables a mixture of inheco and opentrons thermocyclers to be used on the same deck at the same time without unintended interactions with other drivers.

For users with multiple thermocyclers connected, they will need to specify the port specifically.

rickwierenga:
nice, I like that

pylabrobot/thermocycling/opentrons_backend_usb.py
ben-ray: experiencing llm-induced encephalopathy, thank you for fixing

pylabrobot/thermocycling/opentrons_backend_usb.py
```diff
@@ -0,0 +1,465 @@
...
+try:
+  import serial.tools.list_ports
+  from opentrons.drivers.thermocycler import ThermocyclerDriverFactory
+  from opentrons.drivers.thermocycler.abstract import AbstractThermocyclerDriver
+  from opentrons.drivers.types import PlateTemperature, Temperature, ThermocyclerLidStatus
+
+  USE_OPENTRONS_DRIVER = True
+except ImportError:
+  USE_OPENTRONS_DRIVER = False
+
```
rickwierenga: as is the new pattern in PLR, let's keep track of the ImoprtError so we can raise it later if needed. see examples elsewhere in the repo

ben-ray:
(has been tested on a new gen1 thermocycler without thermistor errors, and seems to work reliably!)

pylabrobot/thermocycling/opentrons_backend_usb.py
```diff
@@ -0,0 +1,465 @@
...
+from typing import List, Optional
+
+from pylabrobot.thermocycling.backend import ThermocyclerBackend
+from pylabrobot.thermocycling.standard import (
+  BlockStatus,
+  LidStatus,
+)
+from pylabrobot.thermocycling.standard import (
+  Protocol as ThermocyclingProtocol,
+)
```
rickwierenga: why not keep the standard PLR name?

pylabrobot/thermocycling/opentrons_backend_usb.py
```diff
@@ -0,0 +1,465 @@
...
+    assert self.driver is not None
+    await set_temperature_no_pause(
+      driver=self.driver,
+      temperature=temperature,
+      hold_time_seconds=hold_time_seconds,
+      hold_time_minutes=hold_time_minutes,
+      ramp_rate=ramp_rate,
+      volume=volume,
+    )
+    await wait_for_block_target_simple(self.driver)
```
rickwierenga: this is a private method that does not seem to be used. I suggest either making it public or deleting

pylabrobot/thermocycling/opentrons_backend_usb.py
```diff
@@ -0,0 +1,465 @@
...
+    plate_temp = await self.driver.get_plate_temperature()
+    if plate_temp.target is not None and abs(plate_temp.current - plate_temp.target) < 1.0:
+      return BlockStatus.HOLDING_AT_TARGET
+    return BlockStatus.IDLE
+
+  async def get_hold_time(self) -> float:
+    """Get remaining hold time in seconds."""
+    # USB driver doesn't provide hold time information
+    return 0.0
+
```
rickwierenga: let's raise NotImplementedError instead

pylabrobot/thermocycling/opentrons_backend_usb.py
```diff
@@ -0,0 +1,465 @@
...
+    - Proper temperature waiting
+    """
+    # Convert protocol to execution format
+    steps = self._convert_protocol_to_steps(protocol, block_max_volume)
+
+    # Initialize cycle tracking
+    self._total_cycle_count = 1  # Single execution of the entire protocol
+    self._total_step_count = len(steps)
+    self._current_cycle_index = 1
+    self._current_step_index = 0
```
rickwierenga: is this necessary? I think for most `self._execute_cycles` takes care of it

pylabrobot/thermocycling/opentrons_backend_usb.py
```diff
@@ -0,0 +1,465 @@
...
+        for step in stage.steps:
+          if len(set(step.temperature)) != 1:
+            raise ValueError(
+              f"Opentrons thermocycler only supports a single unique temperature per step, got {set(step.temperature)}"
+            )
+          temperature = step.temperature[0]
+          hold_time = step.hold_seconds
+          ramp_rate = step.rate if step.rate is not None else None
+          steps.append((temperature, hold_time, ramp_rate))
+    return steps
```
rickwierenga: what is the consideration of converting this to the intermediary `List[tuple]` format instead of looping

```
for stage in protocol.stages:
      for _ in range(stage.repeats):  # Repeat the entire stage
        for step in stage.steps:
```

directly in `_execute_cycles`?

potential to simplify

pylabrobot/thermocycling/opentrons_backend_usb.py
rickwierenga: there are many unused private methods for which I would say the same

pylabrobot/thermocycling/opentrons_backend_usb.py
```diff
@@ -0,0 +1,465 @@
...
+
+    # Cycle tracking variables (similar to thermocycler.py)
+    self._total_cycle_count: Optional[int] = None
+    self._current_cycle_index: Optional[int] = None
+    self._total_step_count: Optional[int] = None
+    self._current_step_index: Optional[int] = None
+
+  def _clear_cycle_counters(self) -> None:
+    """Clear the cycle counters."""
+    self._total_cycle_count = None
```
rickwierenga: unnecessary doc string (here and elsewhere). generally if the string is the same as the method name it's safe to just leave it out to reduce bloat

pylabrobot/thermocycling/opentrons_backend_usb.py
```diff
@@ -0,0 +1,465 @@
...
+    super().__init__()
+    if not USE_OPENTRONS_DRIVER:
+      raise RuntimeError(
+        "Opentrons drivers are not installed. Please install the opentrons package."
+      )
+
+    self.port = port
+
+    self.driver: Optional[AbstractThermocyclerDriver] = None
+    self._current_protocol: Optional[ThermocyclingProtocol] = None
```
rickwierenga: driver private?

pylabrobot/thermocycling/opentrons_backend_usb.py
```diff
@@ -0,0 +1,465 @@
...
+
+  async def _read_lid_temperature(self) -> Temperature:
+    """Read the current lid temperature."""
+    assert self.driver is not None
+    return await self.driver.get_lid_temperature()
+
+  async def _read_block_temperature(self) -> PlateTemperature:
+    """Read the current block temperature."""
+    assert self.driver is not None
+    return await self.driver.get_plate_temperature()
```
rickwierenga: there is this and there is the PLR `get_block_current_temperature`. why? can `_read_block_temperature` be deleted? same for lid

pylabrobot/thermocycling/opentrons_backend_usb.py
```diff
@@ -0,0 +1,465 @@
...
+
+  @property
+  def total_step_count(self) -> Optional[int]:
+    """Get the total step count."""
+    return self._total_step_count
+
+  @property
+  def current_step_index(self) -> Optional[int]:
+    """Get the current step index."""
+    return self._current_step_index
```
rickwierenga: why do these exist when we already have methods for them?

pylabrobot/thermocycling/opentrons_backend_usb.py
rickwierenga: eg https://github.com/PyLabRobot/pylabrobot/blob/20633c758671d452907b72a944018a5b98aa895e/pylabrobot/thermocycling/opentrons_backend.py#L26

pylabrobot/thermocycling/opentrons_backend_usb.py
```diff
@@ -290,6 +275,22 @@ async def setup(self):
...
+          f"No Opentrons Thermocycler found with supported VID:PID pairs: {self.SUPPORTED_USB_IDS}"
+        )
+      elif len(opentrons_ports) > 1:
+        available_ports = [p.device for p in opentrons_ports]
+        raise RuntimeError(
+          f"Multiple Opentrons Thermocyclers found: {available_ports}. Please specify the port explicitly."
+        )
+      else:
+        self.port = opentrons_ports[0].device
+
```
rickwierenga: `self.port` is still set here

I don't think setup should change that

why not have a local `port` var in just `setup`

pylabrobot/thermocycling/opentrons_backend_usb.py
rickwierenga: https://github.com/PyLabRobot/pylabrobot/pull/673/commits/f543d4dc6ffbc1a0d30ac5073c4db2386de42661#r2361363808

pylabrobot/thermocycling/opentrons_backend_usb.py
ben-ray: keeping these just so i can expose abstract thermocycler class getters for these values

pylabrobot/thermocycling/opentrons_backend_usb.py
```diff
@@ -0,0 +1,338 @@
...
+      else:
+        port = opentrons_ports[0].device
+
+    self._driver = await ThermocyclerDriverFactory.create(port, self._loop)
+    assert self._driver is not None
+
+  async def stop(self):
+    if self._driver is not None:
+      await self.deactivate_block()
+      await self.deactivate_lid()
```
rickwierenga, line 216: Should driver be set to none here?

pylabrobot/thermocycling/opentrons_backend_usb.py
```diff
@@ -0,0 +1,338 @@
...
+      elif len(opentrons_ports) > 1:
+        available_ports = [p.device for p in opentrons_ports]
+        raise RuntimeError(
+          f"Multiple Opentrons Thermocyclers found: {available_ports}. Please specify the port explicitly."
+        )
+      else:
+        port = opentrons_ports[0].device
+
+    self._driver = await ThermocyclerDriverFactory.create(port, self._loop)
+    assert self._driver is not None
```
rickwierenga: assert here is not necessary 🙃

pylabrobot/thermocycling/opentrons_backend_usb.py
```diff
@@ -0,0 +1,338 @@
...
+            if len(set(step.temperature)) != 1:
+              raise ValueError(
+                f"Opentrons thermocycler only supports a single unique temperature per step, got {set(step.temperature)}"
+              )
+            temperature = step.temperature[0]
+            hold_time = step.hold_seconds
+            ramp_rate = step.rate if step.rate is not None else None
+            step_index += 1
+            self._current_step_index = step_index
+            await execute_cycle_step(
```
rickwierenga: instead of two variables `step_index` and `self. _current_step_index`, let's just use `self._current_step_index`

same for `total_steps` just above

rickwierenga:
thank you so much for this!

i have been wanting to make this PR since i first saw the file 2 years ago in the media lab

PR #674, opened 2025-09-19
Update Hamilton TipRack Names

BioCam:
Hi everyone,

Long been discussed, here now a first implementation of the PLR TipRack naming, starting with the (currently PLR-integrated) TipRacks from Hamilton.

Summary:
Screenshot 2025-09-19 at 13 43 15

Docstrings still contain the OEM's name for information purposes and easing the transition process.
Screenshot 2025-09-19 at 13 40 45

Catalogue/part numbers are a requirement.
Where part numbers are not available TODOs hint at the insecurity of the definition.
I also started collecting information regarding CORE-I vs CORE-II cat number differences inside the docstrings (this _can_ in the future be extended for safety/compatibility checking features).

Please let me know whether you have suggestions on updating the new naming system or see etch cases.


rickwierenga:
now that we are in the process, let's also make them lower case because they are functions. not all resource definition functions follow this pattern yet (they should, and we should fix them later), but let's not introduce more things we are gonna have to rename later. I will do this

rickwierenga:
corresponding forum thread: https://discuss.pylabrobot.org/t/naming-tip-racks/150

BioCam:
Amazing updates, thank you @rickwierenga. I love all of them :)

Issue #679, opened 2025-09-24
TipTracker Errors on STAR 96-head Using Partially-Filled TipRack


BioCam:
This is already working when using STAR_backend directly on the machine but not with Visualizer+tip_tracking:

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[27], line 4
      1 # Remove supernatant
      2 for idx in range(no_plates):
----> 4     await lh.pick_up_tips96(lh.deck.get_resource("tip_1000ul_00"))

File ~/pylabrobot/pylabrobot/liquid_handling/liquid_handler.py:1396, in LiquidHandler.pick_up_tips96(self, tip_rack, offset, **backend_kwargs)
   1394       tip_spot.tracker.rollback()
   1395     self.head96[i].rollback()
-> 1396   raise error
   1397 else:
   1398   for i, tip_spot in enumerate(tip_rack.get_all_items()):

File ~/pylabrobot/pylabrobot/liquid_handling/liquid_handler.py:1390, in LiquidHandler.pick_up_tips96(self, tip_rack, offset, **backend_kwargs)
   1388 pickup_operation = PickupTipRack(resource=tip_rack, offset=offset)
   1389 try:
-> 1390   await self.backend.pick_up_tips96(pickup=pickup_operation, **backend_kwargs)
   1391 except Exception as error:
   1392   for i, tip_spot in enumerate(tip_rack.get_all_items()):

File ~/pylabrobot/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py:2193, in STARBackend.pick_up_tips96(self, pickup, tip_pickup_method, z_deposit_position, minimum_height_command_end, minimum_traverse_height_at_beginning_of_a_command)
   2191     break
   2192 if prototypical_tip is None:
-> 2193   raise ValueError("No tips found in the tip rack.")
   2194 assert isinstance(prototypical_tip, HamiltonTip), "Tip type must be HamiltonTip."
   2195 ttti = await self.get_or_assign_tip_type_index(prototypical_tip)

ValueError: No tips found in the tip rack.
```

<img width="147" height="541" alt="Image" src="https://github.com/user-attachments/assets/ea6e3ee4-8433-4df8-9aaa-a5acff11d491" />

rickwierenga:
is tip tracking enabled when you hit this error case?

BioCam:
Yes, and the error disappears when I 
```python
set_tip_tracking(False)
```

[...]

PR #680, opened 2025-09-25
fix STARBackend.pick_up_tips96 with tip tracking enabled

rickwierenga:
fixing https://github.com/PyLabRobot/pylabrobot/issues/679

it was actually broken for all pick_up_tips96 with tip tracking enabled regardless of the state of the tip rack because LH removes the tips before calling the backend command


rickwierenga:
also added a test for this behavior

PR #681, opened 2025-09-25
Enable Partial 96head Usage in `LiquidHandler`

BioCam:
Currently `LiquidHandler` assumes that 96head usage always has 96 tips attached to the 96head.

This is limiting.

Here I am updating the liquid_handler.py `aspirate96` & `dispense96` methods to enable partial 96head usage:

These method now sort for channels that have a tip attached to them, and then update their pipetting behaviour only for those tips and corresponding wells.


rickwierenga:
Lets pad the indices where there is no tip with None so we keep the channel index information.

rickwierenga:
see https://github.com/PyLabRobot/pylabrobot/pull/680 for an example

PR #685, opened 2025-09-26
Take z position for `STARBackend.drop_tips96()` from operation instead of separate parameter

BioCam:


pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -2213,7 +2213,7 @@ async def pick_up_tips96(
   async def drop_tips96(
     self,
     drop: DropTipRack,
-    z_deposit_position: float = 216.4,
+    z_deposit_position: Optional[float] = None,
     minimum_height_command_end: Optional[float] = None,
     minimum_traverse_height_at_beginning_of_a_command: Optional[float] = None,
```
rickwierenga, line 2217: why keep this at all? 😅

rickwierenga (commented):
thanks for fixing this!

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: in case someone does want to use the backend method for some reason

rickwierenga: this doesn't really make sense because they would have to use `DropTipRack` in any case to specify the rest of the operation. That is because this method is the implementation of the universal LHBackend.drop_tips96 command

having one parameter separate from the rest doesn't make sense

if someone wants to have the raw firmware command they should use the method for it instead of the universal PLR command. we provide one method per firmware command. in this case `STARBackend.discard_tips_core96`

BioCam: I see, you're saying the current implementation never made sense in the first place and we should therefore fix it in here rather than propagate it.

Okay, convinced - I will remove it

PR #687, opened 2025-09-30
`STARBackend.move_iswap_y_relative` channel 0 collision check

rickwierenga:


rickwierenga:
- introduces `STARBackend.iswap_request_module_y` method to request the current position of the iswap module (not gripper)
- checks whether bringing the iswap forward will crash into the first channel
- checks whether brining the first channel backward will crash into the iswap

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -7213,6 +7229,17 @@ async def request_iswap_position(self) -> Coordinate:
       z=(resp["zj"] / 10) * (1 if resp["zd"] == 0 else -1),
     )
 
+  @staticmethod
+  def _iswap_y_inc_to_mm(y_inc: int) -> float:
+    mm_per_increment = 0.046302083
+    return round(y_inc * mm_per_increment, 2)
+
+  async def iswap_request_module_y(self) -> float:
```
BioCam, line 7237: This is a duplication and already exists in PyLabRobot since #666

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -7213,6 +7229,17 @@ async def request_iswap_position(self) -> Coordinate:
       z=(resp["zj"] / 10) * (1 if resp["zd"] == 0 else -1),
     )
 
+  @staticmethod
+  def _iswap_y_inc_to_mm(y_inc: int) -> float:
```
BioCam, line 7233: This is a duplication as well, and has existed in PyLabRobot as `y_drive_increment_to_mm` for a while

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
rickwierenga: thanks, I forgot this exists

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
rickwierenga: I thought this is for pip channels, but the conversion is actually the same

PR #695, opened 2025-10-05
add `get_location_wrt`, support decks in a bigger space

rickwierenga:
previously, everything used `get_absolute_location` (location of a resource wrt its root)

which is good enough when the deck is the root, or the lh is the root and deck wrt lh is 0,0,0.

but when lh is in a bigger space, like Workcell, it falls apart because the absolute location is not relevant to a particular liquid handler machine

this PR addresses that by introducing `get_location_wrt` (get location of a resource with respect to another) and uses it with `lh.deck`.


rickwierenga:
rotations we still have to do

as well as resources wrt others that are not or are not in between it and the root

PR #697, opened 2025-10-07
Fix `STARBackend.pick_up_tips96()`

BioCam:
Hi everyone,

## Problem Statement

`STARBackend.pick_up_tips96()` is broken on multiple levels:

**1. The z-coordinate at which tip_pickup has so far been taking place was hardcoded to z=216.5 mm - this is (a) for some tips slightly incorrect, and (b) eliminates the usage of "non-standard" tip_racks such as Hamilton's nested tip racks (NTRs).**
old_way_mistake_no1

**2. The conversion from mm to dmm has so far been applied twice to the tip_pickup z-coordinate offset:**
old_way_mistake_no2

**3. The check for whether the 96-head can go to the tiprack has been applied based on the *bottom* of tip_spot "A1", which doesn't make sense because the 96-head never needs to go to that z-coordinate which is *inside* the tiprack or the tipcarrier.
This leads to incorrect triggering of not being able to go to tip_racks below a the standard tip_carrier height, e.g. NTRs.**

old_way_mistake_no3

## PR Content

1. I made `STARBackend.pick_up_tips96()` compute the location to which the 96-head actually has to go dynamically based on the tips it is picking up + the (potentially variable) location of the tiprack.
2. removed double dmm conversion.
3. Gave the 96-head check the actual z-coordinate used for pickup derived from fix no.1 above.

## Testing

This is a change to an atomic command, i.e. not something we should take lightly - and therefore should perform extensive testing before merging into main.

I tested tip pickup with everything I have:
Standard tip_carrier tiprack pickup of tips...
- 10 ul
- 50 ul
- 300 ul
- 1000 ul

NTR tiprack pickup of tips...
- 50 ul
- 300 ul

Screenshot 2025-10-07 at 18 44 22

Can I please ask others to also test this new `STARBackend.pick_up_tips96()` method for inter-lab cross-validation?

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -2201,43 +2201,76 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
     return ret
 
   async def pick_up_tips96(
-    self,
-    pickup: PickupTipRack,
-    tip_pickup_method: int = 0,
-    z_deposit_position: float = 216.4,
-    minimum_height_command_end: Optional[float] = None,
-    minimum_traverse_height_at_beginning_of_a_command: Optional[float] = None,
+      self,
+      pickup: PickupTipRack,
+      tip_pickup_method: int = 0,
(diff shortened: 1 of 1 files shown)


rickwierenga:
thank you @BioCam !

PR #698, opened 2025-10-08
add LH.transfer and LH.distribute

rickwierenga:
two new functions:

- `LH.transfer`: 1 to 1 relation between source and destination. There might be multiple sources and destinations, but the number of items in each list is equal and the items at index `i` match.

```py
await lh.transfer(
  plate_1.column(3),
  plate_3.column(3),
  vols=[50, 100, 150, 200, 250, 300, 350, 360],
)
```

- `LH.distribute`: 1 to many relation between source and destination. In a dictionary, because the number of items in the destinations might not be equal for every source.

```py
await lh.distribute(
  {
    plate_1.get_item("A1"): [(plate_2.get_item("B1"), 10), (plate_2.get_item("C1"), 10)],
    plate_1.get_item("A2"): [(plate_2.get_item("B2"), 10), (plate_2.get_item("C2"), 10), (plate_2.get_item("D2"), 10)],
    plate_1.get_item("A3"): [(plate_2.get_item("B3"), 10), (plate_2.get_item("C3"), 10)],
  },
  dead_volume=0
)
```

TODO:
- [ ] handle cases where volume exceeds tip volume (need to re-aspirate)
- [ ] take a tip generator as an argument so we can pick up new tips and handle more transfers/distributions than there are channels (this is big for hardware agnosticity, we can specify the same transfer on STAR and OT and it will use one channel for the OT and 8 for the STAR.)


Colinwehrhan:
Whats the dead volume variable?

Also for an aliquot, one aspiration to many destination, it is universal for all liquid handlers I have worked with to always waste the first dispense and the last dispense in to source as they will always be off from the rest. This could or could not be baked in.

rickwierenga:
thanks @Colinwehrhan 

> Whats the dead volume variable?

Volume that is aspirated on top of the sum of destination volumes for a certain channel, so the last dispense does not reach 0. In case of your suggestion to "waste" the first and last dispense this would not be necessary.

> Also for an aliquot, one aspiration to many destination, it is universal for all liquid handlers I have worked with to always waste the first dispense and the last dispense in to source as they will always be off from the rest. This could or could not be baked in.

cool, yes will definitely add as an option

Colinwehrhan:
Gotcha, also sometimes ppl waste it to the waste or another trough as to not to foam the sample source. As with all physical liquid handling there's always a variation you don't think of that is mission critical for an assay

rickwierenga:
> a variation you don't think of that is mission critical for an assay

so far it hasn't really been possible to build good abstractions for this kind of stuff. with PLR for the first time we have at least the opportunity to encode some of this knowledge in our functions&parameters.

cwehrhan:
looping back to this after a long time, haven't had a chance to test this but have time since its a bit slow spinning things back up,  but before I do I think most people would call distribute "aliquot" 

I'll try these out this week.

PR #702, opened 2025-10-10
added atc support for proflex backend

xbtu2:


pylabrobot/thermocycling/proflex.py
rickwierenga, line 1: do you think we should subclass the Proflex in a new `ATCBackend` that adds these methods? My understanding is the automated close/open lid are not supported on the Proflex itself

xbtu2: you are right. it's probably for the best to have a separate backend

xbtu2: do you think it's better to have something like `thermofisher_thermocycler` and have `proflex`, `ATC` and in the future `quantstudio` inherit from that?

rickwierenga: sounds great!

Proflex subclass of ThermoFisherThermocycler shouldn't add anything right?

also im thinking ThermoFisherThermocycler should be an abc?

xbtu2: yeah i think that's a good idea

rickwierenga:
what do you think @xbtu2

xbtu2:
looks good to me!

rickwierenga:
thanks for contributing this!

PR #710, opened 2025-10-14
Fix `STARBackend.get core()` & `STARBackend.put core()`

BioCam:
Hi everyone,

## The Problem

1. The x-coordinate of `STARBackend.get core()` & `STARBackend.put core()` was 1 mm too high on a STAR when compared to VENUS - I noticed because the CORE-gripper tool pickup was working inconsistently.
2. PLR empowers the usage of a `STARBackend.core_adjustment: Coordinate` attribute. However, the `Coordinate` class takes mm as offset values and `STARBackend.get core()` & `STARBackend.put core()` have so far been given the mm coordinate values straight into the firmware commands dmm values. This also fails when a float is used in the `Coordinate` class .

## PR Content

1. Fixed the `STARBackend.get core()` & `STARBackend.put core()` x-coordinate for usage on a STAR - validated on machine.
2. Added correct conversion from `Coordinate` mm to dmm.


BioCam:
Tangentially relevant note (not caused nor influenced by this PR's content but on topic):

PLR currently uses a fairly advanced, dynamic "tip definition inside firmware" approach using command `C0TT` to reassign tips when they are encountered during a PLR run.

To my understanding, this is not currently including non-`Tip` "ChannelHeadTools" like the Gripper!

The Gripper is firmware-encoded to `tip_type_index = 14`.

This means that, currently, I would expect the usage of CORE-grippers to fail **IF** one were to use 15 _different_ types of tips in the same automated Protocol... this is very unlikely to be the case 😅 
But while we're on topic, worth thinking about a permanent solution?

rickwierenga:
> The Gripper is firmware-encoded to tip_type_index = 14.

good catch, I had not realized this

we should definitely define this dynamically like we do for actual tips

BioCam:
> we should definitely define this dynamically like we do for actual tips

I agree - I see the bigger issues to be more that we haven't created the `ChannelHeadTool` class that should be the deciding factor for "tip type indexing" into the firmware 🤔

rickwierenga:
yes exactly `ChannelHeadTool` would make implementing this much neater

rickwierenga:
thanks!

PR #711, opened 2025-10-14
BioShake backend

Nat-is-coding:
A new backend for the BioShake, as well as adding the class to the __init__.py file. Let me know if anything else is needed or revised.


pylabrobot/heating_shaking/bioshake_backend.py
```diff
@@ -0,0 +1,133 @@
...
+        return decoded
+
+    except Exception as e:
+        raise RuntimeError(
+            f"Unexpected error while sending '{cmd}': {type(e).__name__}: {e}"
+        ) from e
+
+  async def setup(self):
+    await super().setup()
+    await self.io.setup()
```
rickwierenga: should the device `home` or `reset` on setup?

Nat-is-coding: Good question. Initially, I had `home` in the setup process, but removed it shortly after when running the script again and again, where I was like, "this does not need to `home` the device every time the connection is established." 

But if someone were to integrate this backend into their PLR method with a liquid handler, where the BioShake is being called once to run their singular protocol, then yes, it should be homed within setup. 

But if someone was calling this script outside of that (on its own, like how I was using it), and calling said script multiple times but for different commands each time, then no, home should not be within the setup, else it would disrupt the process.

Does this make sense, and what do you think? Again, I'm speaking from someone who hasn't/yet used PLR to control a liquid handler, so I'm open to suggestions

rickwierenga: I can't quickly find a command in the firmware doc that would enable us to query whether the machine is homed.

when the user calls `setup` we should assume the machine was just power cycled, or in the worst possible state, and then make it into a state the user can use. So I would say homing is important in setup.

It would be nice for `setup` to take a parameter `skip_home` (as I do it for the opentrons) that would allow the user to specify whether they want to skip homing. Useful when developing, but you'd turn it off in production runs (as you mentioned). This is not a standard parameter of `HeaterShaker.setup`, so it would be passed as "backend_kwarg" (backend keyword argument), which is the pattern in PLR for adding parameters to otherwise universal commands.

The following should work:

```suggestion
  async def setup(self, skip_home: bool = False):
    await super().setup()
    await self.io.setup()
    if not skip_home:
      await self.home()
```

Nat-is-coding: No, sadly, there's no command to get the home or error state of the device, but good suggestion. In that case, I will also have the device reset before homing it, cause if the device is already in an error state, then homing it won't do any good. That happened to me today in the lab, where one of the bioshake was stuck in an error state and refused to be homed (aka still sending back an error response) until I sent the reset command to fix it. I will make the change accordingly and test tomorrow.

rickwierenga: sounds good, thanks for confirming :)

pylabrobot/heating_shaking/bioshake_backend.py
```diff
@@ -0,0 +1,133 @@
...
+
+  async def reset(self):
+    # Reset the BioShake if stuck in "e" state
+    await self._send_command(cmd="resetDevice", delay=30)
+
+  async def home(self):
+    # Initialize the BioShake into home position
+    await self._send_command(cmd="shakeGoHome", delay=5)
+
+  async def shake(self, speed: float, duration: Optional[float] = None):
```
BioCam: I think this should include a parameter for setting the acceleration speed as well

pylabrobot/heating_shaking/bioshake_backend.py
```diff
@@ -0,0 +1,133 @@
...
+    # Set the speed of the shaker
+    set_speed_cmd = f"setShakeTargetSpeed{speed}"
```
BioCam, line 149: Do we know what range of speed values the BioShake accepts?

If yes, I would recommend adding an assertion to raise an error in Python rather than at the firmware level

pylabrobot/heating_shaking/bioshake_backend.py
```diff
@@ -0,0 +1,133 @@
...
+
+try:
+  import serial
+
+  HAS_SERIAL = True
+except ImportError as e:
+  HAS_SERIAL = False
+  _SERIAL_IMPORT_ERROR = e
+
+class BioShake(HeaterShakerBackend):
```
BioCam, line 16: Do you know the different BioShake models this would likely work for?

Screenshot 2025-10-15 at 15 31 41

If we notice a difference in commands between different BioShakes I'd recommend specifying the class name further: e.g. `BioShakeQ1Backend` and adding the known catalgoue/order numbers of the machines that we know can be controlled with this backend into the class' docstring.

pylabrobot/heating_shaking/bioshake_backend.py
Nat-is-coding: I can definitely include that, though I would have to test how it works, as I never worked with controlling the acceleration before, or needed to as of yet.

pylabrobot/heating_shaking/bioshake_backend.py
Nat-is-coding: The integration manual I was referencing says that these commands are for: BioShake 3000, BioShake 3000 elm, BioShake 3000 elm DWP, BioShake D30 elm, BioShake 5000 elm, BioShake 3000-T, BioShake 3000-T elm, BioShake D30-T elm, Heatplate, and ColdPlate. 

I tested my script on the BioShake 3000 elm, BioShake 5000 elm, and BioShake D30-T elm, and can confirm that the firmware codes are the same.

As for the Q1 model, even though I do not have that, the [old integration manual](https://www.qinstruments.com/fileadmin/Article/MANUALS/Integration-manual-en.pdf) shares the same code as the [newer](https://www.qinstruments.com/fileadmin/Article/All/integration-manual-en-1-8-0.pdf) one I am referencing. Therefore, it should work the same. The only difference between them is the physical dimension of the device, obviously.

pylabrobot/heating_shaking/bioshake_backend.py
Nat-is-coding: The range of speed values depends on the model. For the BioShake 3000 series, the highest is 3000 rpm, and for the BioShake 5000 series, the highest is 5000 rpm, and so on. While the commands between each model are the same, I do have Python parse for errors when sending the commands. 

For example, when I send any temp-related command to a BioShake 3000 elm (a model that does not support heating), the command received by the device is like this: "u -> [the cmd you just sent]". I already have the script raise that as a runtime error if the "u ->" pops up, and I could imagine it is the same for applying a higher speed than the model does not support, but I will check to confirm.

If not, then I can further parse by first sending the `getDescription` command to get the name of the model, as the name itself will tell what features are supported: "T" means it has temperature control, "elm" means it has a plate locking mechanism, and so forth.

pylabrobot/heating_shaking/bioshake_backend.py
rickwierenga: I think this is good to keep for now (since it seems the firmware interface is for all BioShake models), maybe as camillo said put a comment describing which models you tested this on. In the future we can always refactor/extend the code to support more models. If we want to support the ColdPlate/HeatPlate separately they could share a base class. Future stuff

Nat-is-coding: That is true, even for the HeatPlate and ColdPlate, I don't know what else users want to control besides starting, getting, and stopping temp control with the current codes I have. Either way, I can prepare a table in the documentation page on the PLR Wiki for the BioShake, where I list the models and what features each supports (shaking, plate lock, heating, active cooling), and what models were already tested with the current script.

pylabrobot/heating_shaking/bioshake_backend.py
rickwierenga: it's not needed for this PR. feel free to add it if you want, but it can also be done in a future PR :)

pylabrobot/heating_shaking/bioshake_backend.py
rickwierenga: table sounds great!

pylabrobot/heating_shaking/bioshake_backend.py
BioCam: Depending on how much time you have to expand the integration I would recommend to go for your suggestion: creating a value checker:

- use `getDescription` during setup to set a class attribute `self.model= await self. _send_command("getDescription")`.
- Then use the known limits for different models in assertions for the command limits for the different machines.

This avoids confusion and raises explicit errors if someone thinks they can one-by-one replace one machine for another in an existing protocol but haven't checked the hardware difference in detail.

But if you want/need to get the backend integrated soon I would classify this as "nice-to-have", and can relatively easily be added later :)

This is awesome work @Nat-is-coding ! Thank you for doing this!

Nat-is-coding:
Also when I `make format` of my branch, it formatted the STAR_Backend.py, so I left it there in case it was needed to approve my branch...? *~~I tried~~*

rickwierenga:
> Also when I make format of my branch, it formatted the STAR_Backend.py, so I left it there in case it was needed to approve my branch...?

my bad... there was a bad version of main (because I didn't format before pushing there) that happens to be the version you made the branch from. it will be fixed when merging main back into this branch because I made the same fix on main. it will resolve automatically. sorry about that

pylabrobot/heating_shaking/bioshake_backend.py
```diff
@@ -0,0 +1,251 @@
...
+  async def set_temperature(self, temperature: float):
+    # Get the min and max set points of the device to assert temperature
+    min_temp = int(float(await self._send_command(cmd="getTempMin", delay=0.2)))
+    max_temp = int(float(await self._send_command(cmd="getTempMax", delay=0.2)))
+
+    assert (
+      min_temp <= temperature <= max_temp
+    ), f"Temperature {temperature} C is out of range. Allowed range is {min_temp}–{max_temp} C."
+
+    temperature = temperature * 10
```
BioCam, line 227: Should we wrap this into a `round(temperature * 10)` ?

rickwierenga: we are doing that here: https://github.com/PyLabRobot/pylabrobot/pull/711/files#diff-fb7fb73412de1d7d3cb3f2130f5a897e1ed89dd672ca8034ec24fb53c0ed1310R233 (albeit with an int conversion not round), after checking that it's a whole number. I suppose the error message could be improved, do you think it's necessary?

BioCam: Nice - yes, I think we should do it now while we still remember :)

PR #712, opened 2025-10-15
STARBackend.{aspirate,disense} probe_liquid_height parameter

rickwierenga:
Add a parameter `probe_liquid_height` to `STARBackend.aspirate` and `STARBackend.dispense`. If the liquid height is not set using the PLR universal `liquid_height` parameter (of `LH.aspirate` and `LH.dispense`), we will probe the liquid height using cLLD (specifically, the function `clld_probe_z_height_using_channel` @biocam introduced in https://github.com/PyLabRobot/pylabrobot/pull/69/). This is implemented in the new `STARBackend.probe_liquid_heights` function.

This will be useful when we do the automated `surface_following_distance` computation, so we don't rely on users to know the exact liquid height. While the STAR _can_ have the `surface_following_distance` parameter work with cLLD natively, the firmware requires us to specify the value for `surface_following_distance` at a time when the exact liquid height may not be known. Since we do need to know the current liquid_height to compute the accurate `surface_following_distance` (it is not linear in most wells), we need probing to determine it using probing before computing surface_following_distance. This PR adds probing, a future PR will add automated surface_following_distance.

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -1613,10 +1613,14 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
     tips: List[HamiltonTip],
     resource_offsets: Optional[List[Coordinate]] = None,
     move_to_z_safety_after: bool = True,
-  ):
+  ) -> List[float]:
     """Probe liquid heights for the specified channels.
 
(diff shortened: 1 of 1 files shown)


rickwierenga:
in the future i hope this functionality can even replace LLD or at least unify with it in the API.

to be complete we would need to implement firmware command `ZE` "Search for liquid surface (in two sections) via pressure LLD and possibly also via capacitive LLD"

PR #713, opened 2025-10-15
STARBackend.{aspirate,disense} auto_surface_following_distance parameter

rickwierenga:
Add a parameter to automatically compute the `surface_following_distance` parameter for STARBackend.{aspirate,disense} operations.

Nicely used with #712, but it can also be used with explicitly set `liquid_height`.


rickwierenga:
TODO: figure out how surface_following_distance works on dispense

rickwierenga:
it goes up instead of down, makes sense

PR #715, opened 2025-10-16
molecular devices plate readers

xbtu2:
molecular devices spectramax plate readers: m5 and 384plus

### DIFF
--- a/pylabrobot/plate_reading/plate_reader.py
+++ b/pylabrobot/plate_reading/plate_reader.py
@@ -74,10 +74,13 @@ class PlateReader(ResourceHolder, Machine):
     plate = self.get_plate() if len(self.children) > 0 else None
     await self.backend.close(plate=plate, **backend_kwargs)
 
-
   @need_setup_finished
   async def read_luminescence(
-    self, focal_height: float, wells: Optional[List[Well]] = None, use_new_return_type: bool = False, **backend_kwargs
+    self,
+    focal_height: float,
+    wells: Optional[List[Well]] = None,
+    use_new_return_type: bool = False,
+    **backend_kwargs,
   ) -> List[Dict]:
     """Read the luminescence from the plate reader.
 
@@ -104,12 +107,16 @@ class PlateReader(ResourceHolder, Machine):
         "The return type of read_luminescence will change in a future version. Please set "
         "use_new_return_type=True to use the new return type."
       )
-      return result[0]["data"]
+      return result[0]["data"]  # type: ignore[no-any-return]
     return result
 
   @need_setup_finished
   async def read_absorbance(
-    self, wavelength: int, wells: Optional[List[Well]] = None, use_new_return_type: bool = False, **backend_kwargs
+    self,
+    wavelength: int,
+    wells: Optional[List[Well]] = None,
+    use_new_return_type: bool = False,
+    **backend_kwargs,
   ) -> List[Dict]:
     """Read the absorbance from the plate reader.
 
@@ -137,7 +144,7 @@ class PlateReader(ResourceHolder, Machine):
         "The return type of read_absorbance will change in a future version. Please set "
         "use_new_return_type=True to use the new return type."
       )
-      return result[0]["data"]
+      return result[0]["data"]  # type: ignore[no-any-return]
     return result
 
   @need_setup_finished
@@ -186,5 +193,5 @@ class PlateReader(ResourceHolder, Machine):
         "The return type of read_fluorescence will change in a future version. Please set "
         "use_new_return_type=True to use the new return type."
       )
-      return result[0]["data"]
+      return result[0]["data"]  # type: ignore[no-any-return]
     return result


pylabrobot/plate_reading/molecular_devices_backend.py
```diff
@@ -0,0 +1,1154 @@
...
+
+
+@dataclass
+class MolecularDevicesDataCollection:
+  """A collection of MolecularDevicesData objects from multiple reads."""
+
+  container_type: str
+  reads: List["MolecularDevicesData"]
+  data_ordering: str
+
```
rickwierenga: this class does not add anything because the container type is specified by the user and the data ordering is universal in PLR

it would be better for the `read_*` functions to return the 2d list that other PLR plate readers do so protocols can be more hardware agnostic

(other PLR plate readers so far do not support / don' have multiple wavelengths implemented, so we might need to make an update to the PlateReader standard)

pylabrobot/plate_reading/molecular_devices_backend.py
```diff
@@ -0,0 +1,1154 @@
...
+
+class MolecularDevicesBackend(PlateReaderBackend, metaclass=ABCMeta):
+  """Backend for Molecular Devices plate readers."""
+
+  def __init__(self, port: str, res_term_char: bytes = b">") -> None:
+    self.port = port
+    self.io = Serial(self.port, baudrate=9600, timeout=0.2)
+    self.res_term_char = res_term_char
+
```
rickwierenga: is `res_term_char` changed between users?

pylabrobot/plate_reading/molecular_devices_backend.py
```diff
@@ -0,0 +1,1154 @@
...
+  async def _send_commands(self, commands: List[Union[Optional[str], Tuple[str, int]]]) -> None:
+    """Send a sequence of commands to the plate reader."""
+    for command_info in commands:
+      if not command_info:
+        continue
+      if isinstance(command_info, tuple):
+        command, num_res_fields = command_info
+        await self.send_command(command, num_res_fields=num_res_fields)
+      else:
+        await self.send_command(command_info)
```
rickwierenga: what do you think about calling send_command on every line as opposed to making a list then calling using a helper function

is there any logic somewhere that manipulates the list? I can't immediately find any instances

it would really simplify the code to send the commands directly

pylabrobot/plate_reading/molecular_devices_backend.py
```diff
@@ -0,0 +1,1154 @@
...
+      try:
+        error_code_str = parts[-1]
+        error_code = int(error_code_str.strip())
+        if error_code in ERROR_CODES:
+          message, err_class = ERROR_CODES[error_code]
+          raise err_class(f"Command '{command}' failed with error {error_code}: {message}")
+        else:
+          raise MolecularDevicesError(
+            f"Command '{command}' failed with unknown error code: {error_code}"
+          )
```
rickwierenga: else after raise

pylabrobot/plate_reading/molecular_devices_backend.py
```diff
@@ -0,0 +1,1154 @@
...
+      except (ValueError, IndexError):
+        raise MolecularDevicesError(
+          f"Command '{command}' failed with unparsable error: {response[0]}"
+        )
+
+    if "OK" not in response[0]:
+      raise MolecularDevicesError(f"Command '{command}' failed with response: {response}")
+    elif "warning" in response[0].lower():
+      logger.warning("Warning for command '%s': %s", command, response)
```
rickwierenga, line 319: else after raise

pylabrobot/plate_reading/molecular_devices_backend.py
```diff
@@ -0,0 +1,1154 @@
...
+        data_str = res[1]
+        parsed_data_collection = self._parse_data(data_str)
+
+        if collection is None:
+          collection = parsed_data_collection
+        all_reads.extend(parsed_data_collection.reads)
+
+      collection.reads = all_reads
+      return collection
+    else:
```
rickwierenga: else after return

pylabrobot/plate_reading/molecular_devices_backend.py
```diff
@@ -0,0 +1,1154 @@
...
+    carriage_speed: CarriageSpeed = CarriageSpeed.NORMAL,
+    read_from_bottom: bool = False,
+    pmt_gain: Union[PmtGain, int] = PmtGain.AUTO,
+    flashes_per_well: int = 10,
+    kinetic_settings: Optional[KineticSettings] = None,
+    spectrum_settings: Optional[SpectrumSettings] = None,
+    cuvette: bool = False,
+    settling_time: int = 0,
+    is_settling_time_on: bool = False,
+  ) -> MolecularDevicesDataCollection:
```
rickwierenga: I think `is_settling_time_on` can be inferred from `settling_time` not being 0? is there a reason to have both?

rickwierenga:
also, very important, please update the list of supported machines! https://github.com/PyLabRobot/pylabrobot/blob/main/docs/user_guide/machines.md

pylabrobot/plate_reading/molecular_devices_backend.py
xbtu2: no i'll fix it

pylabrobot/plate_reading/molecular_devices_backend_tests.py
```diff
@@ -0,0 +1,884 @@
...
+
+  def test_setup_stop(self):
+    # un-mock send_command for this test
+    self.backend.send_command = AsyncMock(wraps=self.backend.send_command)
+    asyncio.run(self.backend.setup())
+    self.mock_serial.setup.assert_called_once()
+    self.backend.send_command.assert_called_with("!")
+    asyncio.run(self.backend.stop())
+    self.mock_serial.stop.assert_called_once()
+
```
rickwierenga: tests can be `async` (inherit class from `unittest.IsolatedAsyncioTestCase`) so you don't need to use `asyncio.run`

pylabrobot/plate_reading/molecular_devices_backend.py
xbtu2: container type is something returned from the instrument. subclasses of this class also contains all wavelengths used for measurement. these classes are written to store all the parsed data but some of the fields may not be necessary for end users.

do you think we should update the plate reader frontend to have lists of wavelengths as parameters? should we also include kinetic and spectrum modes (common for most plate reads) in the frontend? alternatively, i can write wrappers in my backend that only do endpoint measurements and return 2d lists

rickwierenga: > container type is something returned from the instrument.

is it new information? i thought the user specifies it when reading

> subclasses of this class also contains all wavelengths used for measurement.

these are specified by the user when reading so don't need to be returned back

> do you think we should update the plate reader frontend to have lists of wavelengths as parameters?

yes. I can work on this in a small separate PR that I merge before this one because I will need to update the clariostar and cytation backends as well.

what do you think is a good data format? maybe dicts?

```py
>>> pr.read_absorbance(wavelenghts=[300, 400])
{
  300: [...],  # 2d optional float array
  400: [...],  # 2d optional float array
}
```

> should we also include kinetic and spectrum modes (common for most plate reads) in the frontend? 

this changes the output data format right? or can it still be a single 2d list per wavelength? in the case it changes maybe they should be separate functions on both the frontend and backend?

xbtu2: i agree with you. we can drop container type and all wavelengths.

what about the temperature at which the plate is read? Set point isn't always exactly the same as the actual temp.

what about returning a list of dicts?

`>>> pr.read_absorbance(wavelengths=[300,400], kinetic_interval=5, kinetic_num_reads=3)
[ 
  {'wavelength':300, 'time':'2015-08-23 18:12', 'temp':37,  'data': [...]},
  {'wavelength':400, 'time':'2015-08-23 18:12', 'temp':37, 'data': [...]},
  {'wavelength':300, 'time':'2015-08-23 18:13', 'temp':37, 'data': [...]},
  {'wavelength':400, 'time':'2015-08-23 18:13', 'temp':37,  'data': [...]},
...
]
`

xbtu2: what do you think? should i implement this?

rickwierenga: I like it

"temp"/"time" will be backend specific?

rickwierenga: but maybe list of list of dicts:

```
[intervals][wavelengths for this interval]["time","data","wavelength"]
```

or

list of dicts where the list is intervals and the data for each interval is keyed by the wavelength so it's easier to access?

```
[
  {
    600: {"data": [...], "temp": ..., "time": ...},
    700: {"data": [...], "temp": ..., "time": ...},
  }
]
```

xbtu2: temp and time shouldn't be backend specific. most plate readers allow temperature settings and kinetic reads should always have time associated to be useful.
the problem with using wavelength as key is that fluorescence measurments could have different combinations of ex/em wavelengths not just a single one.

rickwierenga: let's assume date+temp are universal, I would still prefer this format

every wavelength is read at every interval right?

> the problem with using wavelength as key is that fluorescence measurments could have different combinations of ex/em wavelengths not just a single one.

we could have a tuple as a key?

xbtu2: i think tuple as a key is fine. 
```
[
 { # timepoint0
  (ex0,em0):  {"data": [...], "temp": ..., "time": ...},
  (ex1,em1):  {"data": [...], "temp": ..., "time": ...},
 },
 { # timepoint1
  (ex0,em0):  {"data": [...], "temp": ..., "time": ...},
  (ex1,em1):  {"data": [...], "temp": ..., "time": ...},
 },
]
````
how about that?

Comment (rickwierenga) [reply pylabrobot/plate_reading/molecular_devices_backend.py]: Looks great!

File: pylabrobot/plate_reading/plate_reader.py [hunk 9]
```diff
@@ -77,11 +77,15 @@ async def close(self, **backend_kwargs) -> None:
...
-    """Read the luminescence from the plate.
+  ) -> List[Dict[Tuple[int, int], Dict]]:
+    """Read the luminescence from the plate reader.
 
     Args:
       focal_height: The focal height to read the luminescence at, in micrometers.
+
+    Returns:
+      A list of dictionaries, one for each timepoint. Each dictionary has a key (0, 0)
+      and a value containing the data, temperature, and time.
```
rickwierenga: why key it by `(0,0)`?

pylabrobot/plate_reading/plate_reader.py
```diff
@@ -94,11 +98,15 @@ async def read_luminescence(
...
-    """Read the absorbance from the plate in OD, unless otherwise specified by the backend.
+  ) -> List[Dict[Tuple[int, int], Dict]]:
+    """Read the absorbance from the plate reader.
 
     Args:
       wavelength: The wavelength to read the absorbance at, in nanometers.
+
+    Returns:
+      A list of dictionaries, one for each timepoint. Each dictionary has a key (wavelength, 0)
+      and a value containing the data, temperature, and time.
```
rickwierenga: why not use the wavelength as the key directly?

pylabrobot/plate_reading/plate_reader.py
rickwierenga, line 1: for existing machines I would like to have a warning before switching to this new data type

maybe we can have a flag on the front end `use_new_return_type = False`, which if it is false we give a warning. if true we return the new type. the backends always return the new type

pylabrobot/plate_reading/plate_reader.py
xbtu2: don't we want to have tuple as key for the sake of consistency among all read types?

rickwierenga: Indexing with `,0` is strange in the api. They are slightly different reads, I don't see why we would use the same format for all

pylabrobot/plate_reading/plate_reader.py
xbtu2: some readers allow wavelength filtering for luminescence and some don't. should we use 0 as key or something else?

xbtu2:
ready to merge?

rickwierenga:
we still have this problem: https://github.com/PyLabRobot/pylabrobot/pull/715#discussion_r2500447793

right now if I merged this it would break all protocol using a plate reader, there needs to be a migration path

xbtu2:
is the change in the last commit (“format”) not enough?

rickwierenga:
my bad 😬

PR #717, opened 2025-10-16
partial plate reading

rickwierenga:
enable reading parts (certain wells) of plates

all plate reader integrations in PLR so far require reading the entire plate

this PR enables reading parts of plates on the cytation, and updates the standard to support this by having the plate reading methods return `List[List[Optional[float]]]` instead of `List[List[float]]`


rickwierenga:
```python
await pr.read_luminescence(
  focal_height=4.5,
  integration_time=0.8,
  wells=plate["A1:B12"]
)
```

Issue #671, opened 2025-09-11
Should I put a modified Hamilton module in the Hamilton resource folder or in 'custom'


har1eyk:
I 3d printed adapters to fit on the `PLT_CAR_L5AC" carrier`. Seen in grey in the image below.
The adapters accept a `Hamilton_MFX_plateholder_DWP_metal_tapped` module.
I'd like to add a resource for these adapters. 

Should I make a new module like ` Hamilton_MFX_plateholder_DWP_metal_tapped_SHORTER` in `/pylabrobot/resources/hamilton/mfx_modules.py`?

Or, should  I make a folder like `custom` under folder `resources` to house the many 3d prints that will augment Hamilton module heights?

<img width="689" height="468" alt="Image" src="https://github.com/user-attachments/assets/773742c6-bb0a-4dd1-bfac-3c0e2536bfce" />

rickwierenga:
Will you share the stl file so it's easy for people to use it? If so, yes in custom but ideally some company/institution folder where you created this.

If it's entirely custom to you, it should go in your own codebase

rickwierenga:
Sick design btw

har1eyk:
Done. Link to .stl shared on our favorite forum. ;)
Adapters for Hamilton Carrier 188039 - 3D Printing - Lab Automation Forums
<https://labautomation.io/t/adapters-for-hamilton-carrier-188039/6561>

rickwierenga:
Awesome

it would be good to copy those instructions and files into the repo so we don't have to rely on an external server to host these

rickwierenga:
hey @har1eyk are you still working on this? anything I can do to help?

har1eyk:
Thanks for reaching out and offering to help, @rickwierenga . Will you tell me where to put the 3d files?

rickwierenga:
it would be great to create a folder for your company/institution in https://github.com/PyLabRobot/pylabrobot/tree/main/pylabrobot/resources and then have an stl with the same name as your resource definition.

it seems the bars are screwed into the `DWP_metal_tapped` part? if so, I would propose modeling them as one resource since we don't have a standalone `DWP_metal_tapped` definition afaik. So the total definition would be a `ResourceHolder` with the child location and size_z different from other mfx resource holders but otherwise equivalent (I think they are the same xy size right?)

har1eyk:
You can close this issue. Discussion continues in PR, https://github.com/PyLabRobot/pylabrobot/pull/720

PR #720, opened 2025-10-17
Add custom height hamilton module 3d print

har1eyk:
This PR is the result of a discussion about how to include custom 3d prints, https://github.com/PyLabRobot/pylabrobot/issues/671


rickwierenga:
very cool, thank you for the PR!

instead of a `custom` folder, let's find a different name because what's "custom"? I suggest your institution name or your own name.

har1eyk:
`custom` folder removed and institutional, `grindbio`, folder provided.

BioCam:
I really like this!

Small suggestion:
Having the contributing PLR-using company at the level of OEMs in the PLR Resource Library is confusing to newcomers who are searching for pre-integrated resources to purchase.

I recommend one subfolder in resources, maybe called `diy`, then move the PLR-using companies folders that contribute diy resources there:

```go
resources
- agenbio
- agilent
- alpaqua
- bioer
- boekel
- celltreat
- cellvis
- corning
- diy
  - grindbio
     - mfx_plateholders.py
-eppendorf 
...
```

I think this would make it much easier for people to distinguish what resources they can create themselves (based on the files and information shared by the contributing PLR-using company, as demonstrated in this PR), and what people can purchase directly from OEMs.

har1eyk:
Good suggestion, @BioCam . I like this idea. 

@rickwierenga , what do you think of this idea? Shall I nest `grindbio` under a `diy` folder? This organization would be clearer for new members and differentiate "purchasable" vs "DIY-able".

rickwierenga:
Convinced, go for it

har1eyk:
Thanks for your thoughts, @rickwierenga . Ready to merge!

rickwierenga:
thanks @har1eyk for the first DIY labware in plr:main! historic, and hopefully we will get many more

thanks @BioCam for the suggestion to put it under diy!

rickwierenga:
@har1eyk you should make a post on the announcements board! https://discuss.pylabrobot.org/c/announcements

PR #721, opened 2025-10-18
Correct CellTreat_12_troughplate_15000ul_Vb

fderop:
won't load otherwise since circular wells (default arg) don't accept diff x/y dim


rickwierenga:
thanks!

PR #724, opened 2025-10-22
hamilton star: support both core gripper types

rickwierenga:
support both core gripper types:

```
deck = HamiltonSTARDeck(
  core_grippers = "1000uL-5mL-on-waste" # "1000uL-at-waste"
)
```

only the `1000uL-5mL` ("on waste") variant was supported in PLR. Now we also support the `1000uL` "at waste" variant

image
image

### DIFF
--- a/pylabrobot/resources/hamilton/hamilton_decks.py
+++ b/pylabrobot/resources/hamilton/hamilton_decks.py
@@ -439,7 +439,9 @@ class HamiltonSTARDeck(HamiltonDeck):
     with_trash: bool = True,
     with_trash96: bool = True,
     with_teaching_rack: bool = True,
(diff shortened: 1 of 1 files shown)


rickwierenga:
in the unit test you can see we previously had an inconsistent location for the core grippers: pickup was correct but not drop

- ya: 124.0 -> 125.0 (as pickup already was)
- yb: 106.5 -> 107.0 (as drop already was)

rickwierenga:
relevant context: @BioCam fixed the x location in https://github.com/PyLabRobot/pylabrobot/pull/710/

rickwierenga:
also deleting `no_teaching_rack` and `no_trash` that we deprecated in https://github.com/PyLabRobot/pylabrobot/pull/347 last January

PR #730, opened 2025-10-24
Add check to core grippers

fderop:


pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -5193,6 +5193,10 @@ def _get_core_front_back(self):
       )
       * 10
     )
+    assert (
+      back_channel_y_center > front_channel_y_center
+    ), "back_channel_y_center must be greater than front_channel_y_center"
+    assert front_channel_y_center > 60, "front_channel_y_center must be less than 6mm (60 [0.1mm])"
```
BioCam, line 5199: @fderop, should this say `"front_channel_y_center must be bigger than 6mm (60 [0.1mm])"` ?

Also, should we add an "upper bound check" for the `back_channel_y_center` in addition to this "lower bound check" for the `front_channel_y_center`,  based on the never changing y-range of the channels?

Screenshot 2025-09-06 at 15 29 16 copy

fderop: i believe so - @rickwierenga hacked my account and pushed this commit from there

fderop: > Also, should we add an "upper bound check" for the `back_channel_y_center` in addition to this "lower bound check" for the `front_channel_y_center`, based on the never changing y-range of the channels?

oh yes sounds like a good idea

PR #732, opened 2025-10-25
fragment analyzer

xbtu2:
agilent fragment analyzer backend. this does interact with vendor software running on a windows computer that connects to the instrument. is this allowed?


rickwierenga:
Why not have a firmware interface?

xbtu2:
a firmware interface is possible but the ccd camera involved in capturing capillary image seems to be tricky to deal with. i thought this could be a solution for the time being. conceptually, there is not much difference between this and a machine with an embedded windows system exposing a tcp interface

rickwierenga:
the main problem is not windows itself (exactly as you mentioned), but it's highly correlated with the actual problem which is that these interfaces are not simple atomic apis. This particular interface seems mostly based on `run_method` and `set_ladder_file`. I am not super familiar with this class of machine, so correct me if I'm wrong, but this seems to offload protocol logic to a proprietary interface which makes making a universal API impossible. Is the plate geometry relevant at all? If so, how is it defined?

> ccd camera

does this run over the same interface? the cytation uses a flir camera, for which existing drivers exist

rickwierenga:
the proprietary software is a second problem. we can't host these files in this repo, so in case people have the machine but not the computer for some reason things get complicated. this might not be a problem for every machine necessarily, but it's not great

xbtu2:
ok i will submit another pr in the future for a firmware based interface

rickwierenga:
looking forward! closing this then.

PR #733, opened 2025-10-25
make volume tracker robust against floating point inaccuracies

fderop:


pylabrobot/resources/volume_tracker.py
```diff
@@ -150,7 +150,7 @@ def remove_liquid(self, volume: float) -> List[Tuple[Optional["Liquid"], float]]
   def add_liquid(self, liquid: Optional["Liquid"], volume: float) -> None:
     """Add liquid to the container."""
 
-    if volume > self.get_free_volume():
+    if volume > self.get_free_volume() and abs(volume - self.get_free_volume()) > 1e-6:
       raise TooLittleVolumeError(
```
rickwierenga: why do we need both? can't we just have the second condition?

fderop: true true, i'll change

pylabrobot/resources/volume_tracker.py
```diff
@@ -124,7 +124,7 @@ def remove_liquid(self, volume: float) -> List[Tuple[Optional["Liquid"], float]]
     """Remove liquid from the container. Top to bottom."""
 
     available_volume = self.get_used_volume()
-    if volume > available_volume and abs(volume - available_volume) > 1e-6:
+    if abs(volume - available_volume) > 1e-6:
```
fderop: I deleted it here then @rickwierenga

rickwierenga: also the `abs` is throwing it off (that would check for equality, we are checking for an inequality)

PR #735, opened 2025-10-28
Control Opentrons temperature module directly with USB

nedru004:
Instead of relying on using the opentrons API, control the opentrons temperature module directly with a serial usb connection. This allows me to use an opentrons module with a hamilton liquid handler (already had the temp module sitting around). 

I used this list of commands to communicate with device. 
https://github.com/Opentrons/opentrons-modules/tree/edge/arduino-modules/temp-deck#readme

This requires the user to define the com port. Default is to use the opentrons API if available

Also, if I understand correctly, this module has active cooling. I can cool to 4C

### DIFF
--- a/pylabrobot/temperature_controlling/opentrons.py
+++ b/pylabrobot/temperature_controlling/opentrons.py
@@ -2,6 +2,7 @@ from typing import Optional
 
 from pylabrobot.resources import Coordinate, ItemizedResource
 from pylabrobot.resources.opentrons.module import OTModule
+from pylabrobot.temperature_controlling.backend import TemperatureControllerBackend
 from pylabrobot.temperature_controlling.opentrons_backend import (
   OpentronsTemperatureModuleBackend,
 )
@@ -46,6 +47,7 @@ class OpentronsTemperatureModuleV2(TemperatureController, OTModule):
     if opentrons_id is not None and serial_port is not None:
       raise ValueError("Exactly one of `opentrons_id` or `serial_port` must be provided.")
 
+    backend: TemperatureControllerBackend
     if serial_port is not None:
       backend = OpentronsTemperatureModuleUSBBackend(port=serial_port)
     else:


rickwierenga:
thanks for the PR!

for the OT thermocycler we also have a usb and http backend, and there we implemented them as two separate backends (see https://github.com/PyLabRobot/pylabrobot/pull/673). I think this is cleaner than every single method having `if else`.

nedru004:
Ahh. That makes more sense. Thanks for the example. I will update to have a separate backend for usb control. I will also be 3d printing a mount for the MFX carrier.

rickwierenga:
sick

> 3d printing a mount for the MFX carrier.

follow up pr 👀

nedru004:
Updated the files. The original backend through OT is now restored to the original file and I added a new backend for usb control. This works on my end and matches the format of OT thermocycler. Do you want documentation for the user guide?

ben-ray:
thanks for implementing this btw, using it now

rickwierenga:
> Do you want documentation for the user guide?

yes please

and please let me know when it's ready for review

pylabrobot/temperature_controlling/opentrons.py
```diff
@@ -21,7 +21,7 @@ def __init__(
     self,
     name: str,
     opentrons_id: str,
-    child_location: Coordinate = Coordinate.zero(),
+    child_location: Coordinate = Coordinate(0, 0, 80.1), # dimensional drawing from OT (x and y are not changed wrt parent)
```
nedru004: This is confusing since the temperature module does change the height. It doesnt change x or y. Is this handled differently in opentrons? I need the change in height when using hamilton. 

Comes from this document:https://insights.opentrons.com/hubfs/Products/Modules/Temperature%20Module%20White%20Paper.pdf

image

nedru004:
I am ready for review.

I have one comment about the child location

rickwierenga:
image

great to see people are benefiting from not being constrained to python3.10 on opentrons like the vendor interface

rickwierenga:
thank you @nedru004!

rickwierenga:
I sometimes make twitter/linked in posts to announce new device integrations. people like seeing they can use their devices in new ways. which accounts should I tag to give credit?

nedru004:
Thanks!

x.com: @NedrudDavid
bsky.app: @david-nedrud.bsky.social

PR #738, opened 2025-11-03
Add agilent_96_wellplate_150uL_Vb

fderop:


pylabrobot/resources/agilent/plates.py
```diff
@@ -1,13 +1,37 @@
...
+    d=6.4, h_pyramid=12.5, liquid_volume=v
+  )
 
 
 def agilent_96_wellplate_150uL_Ub(name: str) -> Plate:
   """
   Part number: 5042-8502
-
-  https://www.agilent.com/cs/library/datasheets/public/ds-well-plate-specifications-5994-6035en-agilent.pdf
   """
```
rickwierenga, line 68: ?

fderop: I thought that led to the wrong file but it doesn't apparently

fderop: reverted

pylabrobot/resources/agilent/plates.py
```diff
@@ -41,3 +67,43 @@ def agilent_96_wellplate_150uL_Ub(name: str) -> Plate:
...
+def agilent_96_wellplate_150uL_Vb(name: str) -> Plate:
+  """
+  Part number: 5042-8502
+  """
+
+  diameter = 6.4  # from spec
+
+  well_kwargs = {
+    "size_x": diameter,  # from spec
+    "size_y": diameter,  # from spec
```
rickwierenga: which spec? please link it

pylabrobot/resources/height_volume_functions.py
```diff
@@ -434,6 +434,33 @@ def calculate_liquid_height_container_1segment_round_fbottom(
   return liquid_height
 
 
+def calculate_liquid_height_container_1segment_round_vbottom(
```
fderop, line 437: hard to believe that there was no existing v-bottom well calculator?

pylabrobot/resources/agilent/plates.py
fderop: I came to the conclusion that the old definition was simply wrong. There is no `Ub` of that type, only `Vb`, and the linked spec sheet, PLR dimensions, and part number match the `Vb`. don't think depreciation func is necessary for this correction.

PR #740, opened 2025-11-04
Nimbus Integration: TCP Backend and Interface Introspection 

cmoscy:
1. Added TCP I/O protocol with the Hamilton specific TCPBackend supporting Nimbus (and possibly Prep) specific communication and introspection through their IP->Harp2->Hoi2 protocol. 
2. Currently includes basics of instrument connection, introspection. 
3. Can use HamiltonCommand to construct message to run instrument commands while managing any responses received from the instrument.


cmoscy:
My fault with the typing/linting. Cleaning this up and will update shortly

rickwierenga:
it's fine to do those things last

for linting `make format` might do a lot easily

but otherwise don't bother making tests pass every revision, it's more about content than style at this point

rickwierenga:
do you think it would be possible to load some commands like aspirate and dispense and implement them in python in PLR?

cmoscy:
Got it! Was pretty minor to get it to pass. 

That's definitely possible! Primary bottleneck for me has been handling that Nimbus has several deck/waste block configurations. Pretty sure there's a way to detect a lot of this so there's no potentially dangerous assumptions made. 

Could add basic instrument control into a NimbusBackend pretty easily in the next 1-2 days (User would have to specify coordinates themselves until we get a deck layout defined). 

Tip pickup, drop, aspirate, dispense, initialization, and door control seem like a good minimal implementation? Anything else that would be worth prioritizing?

rickwierenga:
those "big four" + setup are the most important, everything else is nice to have. appreciate it!

for the stars the user has to specify some things about their deck layout, like where the core grippers are if they have them. after that it's just modeled through the universal resource model. we could have a similar thing for the nimbus

```python
deck = NimbusDeck(
  trash_pos = "option a"
)
```

cmoscy:
Sounds good, thanks for the example. Checkpointed the cleanup and improved the introspection to help me put all that together next!

cmoscy:
Almost there, work got busy but just some testing on aspirate and dispense left!

rickwierenga:
man this is so fk cool

cmoscy:
Aspirate and dispense implemented! cLLD and pLLD params can be passed to the interface, but so far I've only validated fixed height operations on hardware. Example notebook run in nimbus-dev attached. Logging is verbose so you can see wired commands and their params easily. 


https://github.com/user-attachments/assets/c8a23fef-4681-4408-9cac-add2141dea9f

rickwierenga:
legend

some high level comments:

- I made some improvements to the existing TCP io layer today (https://github.com/PyLabRobot/pylabrobot/pull/753). could you try using/extending that? would be ideal to have only one in all of PLR
- my nimbus is unfortunately quite broken so can't test this locally, but if it's working for you that's enough
- I want to use your implementation of the hamilton tcp protocol in the prep since yours is much more advanced. I think for that I will merge this branch and then rewrite the prep branch to use this code
- any other big things you want to do in this branch? if not I will review in more detail and maybe make some stylistic changes before merging

cmoscy:
Thanks for taking a look:

1. Definitely! Should be able to migrate over to that TCP layer this week. 
2. I might consider getting the CORE gripper working for plate transport either in this PR or an additional one. If you have a preference let me know. That's the last major feature I'd need in the near term on the liquid handler side. Next points of focus would likely be ODTC and/or Kingfisher Presto.

cmoscy:
Swapped to the refactored "Socket" interface. Passes tests on my hardware and runs that same example notebook. Might give adding the CORE plate motion commands a go this week/weekend. But everything else is good to go / should pass gh tests.

rickwierenga:
could you please "allow edit by maintainers" for this PR so I can make some simple edits like formatting?

cmoscy:
Done! Sorry that wasn't included earlier.

rickwierenga:
thanks

excited to finally start properly looking at this :)

pylabrobot/liquid_handling/backends/hamilton/tcp/messages.py
```diff
@@ -0,0 +1,1072 @@
...
+    }
+
+    # Check scalar types first
+    # Cast int to HamiltonDataType enum for dict lookup
+    try:
+      data_type = HamiltonDataType(type_id)
+      if data_type in scalar_parsers:
+        return scalar_parsers[data_type]()
+    except ValueError:
+      pass  # Not a valid enum value, continue to other checks
```
rickwierenga, line 367: since the data type is given by type_id, I don't think there will be another match in case the value conversion fails. it will just fail the other type_id checks and eventually reach "Unknown or unsupported type_id", even though we did recognize the type_id but parsing failed. why not fail early with the correct error? or am I missing something here?

rickwierenga: sorry never mind

rickwierenga:
is this correct?

- `z_start_positions`: begin of tip pickup/drop. (`begin_tip_pick_up_process`/`begin_tip_deposit_process` on STAR)
- `z_stop_positions`: end of tip pickup/drop
- `z_final_positions`: height of channels at the end of the commands (`z_position_at_end_of_a_command` in STAR)

if so, I am gonna change it to the star names so it's the same terminology across all Hamilton (and hopefully all liquid handler) backends

cmoscy:
Yes, that's right. Sounds great to make it more consistent across instruments

rickwierenga:
what is InitializeSmartRoll?

rickwierenga:
for traverse height, what is the highest possible default?

```
traverse_height: Uses deck z_max if not provided
```

This is very dangerous if the deck is not fully or not accurately defined in PLR.

in the star / vantage, we default to the maximum allowed value (or a user defined value)

cmoscy:
> what is InitializeSmartRoll?

Command used by the default initialization procedure in run control when channels are present as read in the log data from hamilton run control. "Roll" just is the roll off behavior where there is a small amount of x movement when tips are ejected by the channels, "rolling" them off into the waste bin since it generally touches the tips to the waste block surface before ejecting them. 



> for traverse height, what is the highest possible default?
> 
> ```
> traverse_height: Uses deck z_max if not provided
> ```
> 
> This is very dangerous if the deck is not fully or not accurately defined in PLR.
> 
> in the star / vantage, we default to the maximum allowed value (or a user defined value)

Totally valid and nice catch. If I remember correctly the max is 146mm (Or at least that's the default traverse height/highest I've tested). Happy to comb through and help eliminate any other instances of this behavior. Or if there's anything else I can help with let me know.

rickwierenga:
thanks

next question:

I saw some

```
TODO: Calculate from resources properly (resource Z + offset)
```

the value in these cases is actually correctly computed (taken from z_positions_mm), it already takes the offsets into account. is this an old todo? or what is there to add?

cmoscy:
> thanks
> 
> next question:
> 
> I saw some
> 
> ```
> TODO: Calculate from resources properly (resource Z + offset)
> ```
> 
> the value in these cases is actually correctly computed (taken from z_positions_mm), it already takes the offsets into account. is this an old todo? or what is there to add?

That's an old one in Nimbus Backend, must have missed removing it earlier. Thanks for asking.

Intended TODO's should only be one about seeing if I can get the z_max/traverse height through introspection. And another about _get_tip_type_from_tip as that method is pretty hacky / relies on knowledge about what tips hamilton offfers.

pylabrobot/liquid_handling/backends/hamilton/nimbus_backend.py
```diff
@@ -0,0 +1,2763 @@
...
+      # Convert to Hamilton coordinates (returns in mm)
+      hamilton_coord = self._deck.to_hamilton_coordinate(abs_location)
+
+      x_positions_mm.append(hamilton_coord.x)
+      y_positions_mm.append(hamilton_coord.y)
+      z_positions_mm.append(hamilton_coord.z)
+
+    # Convert positions to 0.01mm units (multiply by 100)
+    x_positions = [int(round(x * 100)) for x in x_positions_mm]
+    y_positions = [int(round(y * 100)) for y in y_positions_mm]
```
rickwierenga: this results in the same y position for all channels

on the STAR/Vantage, the y positions need to be specified for each channel even when discarding into trash. when they are not unique, it will discard the tips at this exact xyz one by one

is that also the case on the nimbus or will it automatically figure out how to discard all tips in one motion?

cmoscy: You're correct in that if we built a single xyz position, Nimbus will just use the exact spot one by one. 

Maybe this is a difference in how we're representing our waste block? I've set the nimbus deck to use tip spots within the waste block resource, up to 8 available. 

see  `_create_default_long_waste` line 107 at nimbus_decks.py

Your linked code up top uses `channel_idx` to access each defined waste position in the loop.

Each position supplied is a full x,y,z coordinate, y's should be unique.

rickwierenga: oops, missed that. I see how it works. thanks

pylabrobot/liquid_handling/backends/hamilton/nimbus_backend.py
```diff
@@ -0,0 +1,2453 @@
...
+          )
+        )
+        enabled = config["enabled"][0] if config["enabled"] else False
+        if channel_num not in self._channel_configurations:
+          self._channel_configurations[channel_num] = {}
+        self._channel_configurations[channel_num][2] = enabled
+        logger.debug(f"Channel {channel_num} configuration (index 2): enabled={enabled}")
+      except Exception as e:
+        logger.warning(f"Failed to get channel configuration for channel {channel_num}: {e}")
+
```
rickwierenga, line 1751: `self._channel_configurations` is written to, but I can't find where this information is used.

assuming it is used somewhere, would it be possible to query this configuration on setup instead of doing it jit in aspirate/dispense?

cmoscy: As I understand it, GetChannelConfiguration commands run jit prior to aspirate/dispense for some internal state in the Nimbus. If I remember correctly, if these are not called for "Active" channels just prior to an aspirate/dispense command, the command will fail. It's been awhile, so I can test to be sure. 

This pattern is present in aspirate/dispense commands from Hamilton's method editor/run control as well. This file shows the command sequence from com data when running protocols using their control software. 
[command_sequence_log.txt](https://github.com/user-attachments/files/24698879/command_sequence_log.txt)

Protocol was initialization, tip pickup, a few aspirate+dispense cycles, drop tips, park. with door control since I'm in an enclosed nimbus.

rickwierenga: thanks!

it's not super important, I was just curious about it as I work through the code

I will ask you at the end before merging to make sure everything is still working (since our nimbus unfortunately is still completely broken), so if you wouldn't mind testing with this part commented out at that time it'd be nice

pylabrobot/liquid_handling/backends/hamilton/nimbus_backend.py
rickwierenga: for other liquid handling robots, we actually compute offsets when discarding tips (=dropping to trash): https://github.com/PyLabRobot/pylabrobot/blob/48658109d551c1a5e05fdaf42d369aec48323b66/pylabrobot/liquid_handling/liquid_handler.py#L732 this essentially spaces tips equally in y-space

on setup, the y-coordinates are computed in the initialization function (so essentially hard-coded)

I will refactor the nimbus to use the same for consistency.

this could really simplify things as from the backend perspective it would just be a "tip drop with offsets". we would keep the "all_waste" check for roll vs "normal" drop (this distinction also exists on star / vantage as "place&shift" vs "drop")

I actually really like this approach of having distinct "tip trash locations" coded into the deck as you did, and I think we should consider making this the PLR-wide standard which `LH.discard_tips` will then use. But that would also require changing some other things, so I'm making it consistent first and then we can refactor everything at once later

pylabrobot/liquid_handling/backends/hamilton/nimbus_backend.py
cmoscy: Happy to test anytime! Including A/Bs if you’ve got a few variants/behaviors to test. 

Really appreciate that you're being thorough.  Hope it hasn’t been too painful getting everything into good shape/harmonized.

rickwierenga: thanks!

> Really appreciate that you're being thorough. Hope it hasn’t been too painful getting everything into good shape/harmonized.

it's fun! you made some slightly different choices in a few places for functionalities that also exist on other robots, which I like and want to integrate into other backends (eg https://github.com/PyLabRobot/pylabrobot/pull/740#discussion_r2702785728)

also I can't wait to rewrite my prep backend using your tcp layer, it's so much better than mine :)

rickwierenga:
I am surprised `PickupTips` does not have `z_final_positions`

pylabrobot/liquid_handling/backends/hamilton/nimbus_backend.py
cmoscy: 100% makes sense to get things consistent first

cmoscy:
> I am surprised `PickupTips` does not have `z_final_positions`

Right? PickupGripperTool also lacks a zFinal in the signature. Weird since most other operations have it. But double checked the extracted interface methods and their firmware documentation. They both don't include zFinal.

Pipettor Interface Methods from introspection layer for reference. 
[pipettor_method_ids.txt](https://github.com/user-attachments/files/24699780/pipettor_method_ids.txt)

pylabrobot/liquid_handling/backends/hamilton/nimbus_backend.py
```diff
@@ -0,0 +1,2453 @@
...
+    volumes = [op.volume for op in ops]  # in uL
+    # flow_rate should not be None - if it is, it's an error (no hardcoded fallback)
+    flow_rates: List[float] = []
+    for op in ops:
+      if op.flow_rate is None:
+        raise ValueError(f"flow_rate cannot be None for operation {op}")
+      flow_rates.append(op.flow_rate)
+    blowout_volumes = [
+      op.blow_out_air_volume if op.blow_out_air_volume is not None else 40.0 for op in ops
+    ]  # in uL, default 40
```
rickwierenga: this means the user has to specify flow rates, which isn't the case for other liquid handlers in PLR. they will often want to, but maybe not always (eg for simple prototyping)

what is the default value nimbus uses if anything?

rickwierenga:
what is the difference between HoiPacket and HoiResponse?

cmoscy:
> what is the difference between HoiPacket and HoiResponse?

HoiPacket is the lower level wire protocol, handles low level encoding/decoding. HoiResponse in the interpreter layer that translates enums, action codes, etc that map to hamilton responses/behaviors.

pylabrobot/liquid_handling/backends/hamilton/nimbus_backend.py
cmoscy: 1000 ul tip: 250 asp / 400 disp
300 and 50 ul tip: 100 asp / 180 disp 
 10 ul tip: 100 asp / 75 disp

rickwierenga:
> HoiPacket is the lower level wire protocol, handles low level encoding/decoding. HoiResponse in the interpreter layer that translates enums, action codes, etc that map to hamilton responses/behaviors.

It seems they contain the same information?

parse_message checks the action for error, but then in the success case copies information over into a `HoiResponse`:

```
 return SuccessResponse(
      action=action,
      interface_id=cmd_response.hoi.interface_id,
      action_id=cmd_response.hoi.action_id,
      raw_params=cmd_response.hoi.params,
      response_required=cmd_response.hoi.response_required,
    )
```

and it's pretty similar in the error case.

```
    if isinstance(hoi_response, ErrorResponse):
```

in `send_command` could just be `if action in (Hoi2Action.STATUS_EXCEPTION, Hoi2Action.COMMAND_EXCEPTION, Hoi2Action.INVALID_ACTION_RESPONSE)`

is there a reason I shouldn't merge them into one? (delete `HoiResponse` probably incl related refactoring)

cmoscy:
> > HoiPacket is the lower level wire protocol, handles low level encoding/decoding. HoiResponse in the interpreter layer that translates enums, action codes, etc that map to hamilton responses/behaviors.
> 
> It seems they contain the same information?
> 
> parse_message checks the action for error, but then in the success case copies information over into a `HoiResponse`:
> 
> ```
>  return SuccessResponse(
>       action=action,
>       interface_id=cmd_response.hoi.interface_id,
>       action_id=cmd_response.hoi.action_id,
>       raw_params=cmd_response.hoi.params,
>       response_required=cmd_response.hoi.response_required,
>     )
> ```
> 
> and it's pretty similar in the error case.
> 
> ```
>     if isinstance(hoi_response, ErrorResponse):
> ```
> 
> in `send_command` could just be `if action in (Hoi2Action.STATUS_EXCEPTION, Hoi2Action.COMMAND_EXCEPTION, Hoi2Action.INVALID_ACTION_RESPONSE)`
> 
> is there a reason I shouldn't merge them into one? (delete `HoiResponse` probably incl related refactoring)

Would be fine with me. I originally kept them separate to stay aligned with Hamilton’s .NET structure — it made it easier to test against the firmware DLLs.

Now that we've captured most of it, probably less important to be a 1:1 mirror.

rickwierenga:
> Yes, that's right. Sounds great to make it more consistent across instruments

Unfortunately the spacing between channels is a little odd here:

```
  Position 1 to 2: 19.863 - 1.880 = 17.983mm                                                                                                                                                                           
  Position 2 to 3: 1.880 - (-76.149) = 78.029mm                                                                                                                                                                        
  Position 3 to 4: -76.149 - (-94.132) = 17.983mm                                                                                                                                                                      
  Position 4 to 5: -94.132 - (-152.349) = 58.217mm                                                                                                                                                                     
  Position 5 to 6: -152.349 - (-170.332) = 17.983mm                                                                                                                                                                    
  Position 6 to 7: -170.332 - (-219.549) = 49.217mm                                                                                                                                                                    
  Position 7 to 8: -219.549 - (-237.532) = 17.983mm    
```

This means we can't use `get_tight_single_resource_liquid_op_offsets` which is used by `lh.discard_tips`.

Since I have already decided I want to move having decks define a tip position for every tip spot and having discard use these, it is fine to keep it in this pattern for the nimbus (the new pattern that other backends will also use soon) rather than figuring these things out right now.

rickwierenga:
@cmoscy could you please make sure the code still works as expected? then it will be ready to merge.

rickwierenga:
I am merging the binary parser (https://github.com/PyLabRobot/pylabrobot/pull/740/changes#diff-8e70176e729f055fd4aae4c482c93135752ce382647b1d60f2084a3194c6c902) separately right now so I can use it the tecan infinite PR while finish this up

cmoscy:
> @cmoscy could you please make sure the code still works as expected? then it will be ready to merge.

Will take a look tonight

cmoscy:
> > Yes, that's right. Sounds great to make it more consistent across instruments
> 
> Unfortunately the spacing between channels is a little odd here:
> 
> ```
>   Position 1 to 2: 19.863 - 1.880 = 17.983mm                                                                                                                                                                           
>   Position 2 to 3: 1.880 - (-76.149) = 78.029mm                                                                                                                                                                        
>   Position 3 to 4: -76.149 - (-94.132) = 17.983mm                                                                                                                                                                      
>   Position 4 to 5: -94.132 - (-152.349) = 58.217mm                                                                                                                                                                     
>   Position 5 to 6: -152.349 - (-170.332) = 17.983mm                                                                                                                                                                    
>   Position 6 to 7: -170.332 - (-219.549) = 49.217mm                                                                                                                                                                    
>   Position 7 to 8: -219.549 - (-237.532) = 17.983mm    
> ```
> 
> This means we can't use `get_tight_single_resource_liquid_op_offsets` which is used by `lh.discard_tips`.
> 
> Since I have already decided I want to move having decks define a tip position for every tip spot and having discard use these, it is fine to keep it in this pattern for the nimbus (the new pattern that other backends will also use soon) rather than figuring these things out right now.


Right on, you reminded me of that little detail on the Nimbus. Funny enough those are the default tip waste positions from their deck definition. Think it avoids the teaching positions in case you don't place the little metal piece that is suposed to cover them when not being actively used on their default long waste.

rickwierenga:
That makes sense.

Also, I made a forum post to discuss making this behavior the default: https://discuss.pylabrobot.org/t/per-channel-tip-discard-positions-on-deck-instead-of-computed-offsets/420 because it is more flexible and generally nicer

cmoscy:
Couple regressions, but nothing too bad. I like the consolidation you implemented, will do a little smoothing tomorrow evening. Shouldn't take long.

cmoscy:
Alright! Passes all my example protocols. Last thing was just remembering mixing flow rate(s) can't be zero, otherwise the command gets rejected as invalid. Tested out the flow rate defaults, and other arg combinations where defaults were made more consistent with other backends too. Looking good!

rickwierenga:
amazing! thank you so much for testing! (and of course writing all of it in the first place :))

were you able to test whether loading the configuration was actually needed?

I will fix the unit tests

cmoscy:
> amazing! thank you so much for testing! (and of course writing all of it in the first place :))
> 
> were you able to test whether loading the configuration was actually needed?
> 
> I will fix the unit tests

Of course! Happy that it can hopefully help some others out.

Doh! I will do that and stress test a little more. Worth checking but am pretty sure they're needed. 

I'll do a final set of tests tomorrow evening. If there's anything else you want me to verify let me know.

rickwierenga:
> Happy that it can hopefully help some others out.

already heard one person asking if PLR is compatible with the nimbus recently!

if I am not mistaken, this will the first time people can use protocols from a STAR on a Nimbus and vice versa

rickwierenga:
thank you so much for contributing this @cmoscy !!

PR #742, opened 2025-11-09
STAR surface following tutorial update

rickwierenga:


rickwierenga:
- clld_probe_z_height_using_channel calls self.request_pip_height_last_lld which is C0 meaning it can't be parallelized
- move_z_drive_to_liquid_surface_using_clld does not query, it only sends a command to PX so it can be parallelized

in the code above you see i call move_z_drive_to_liquid_surface_using_clld in parallel and then self.request_pip_height_last_lld once

PR #746, opened 2025-11-10
Add core grippers offset

fderop:


pylabrobot/resources/agilent/plates.py
rickwierenga, line 1: if needed this should be a different PR

PR #748, opened 2025-11-14
Add CoRe gripper ZB barcode read command (ZB)

Dalconzo:
This adds a core_read_barcode helper on the Hamilton STAR backend that wraps the ZB firmware command (“Read barcode using CoRe gripper”).
The command generated matches a known-good sequence from VENUS / simulator:
C0ZBid0215cp45zb2200th2750zy1287bd1ma0250 2100 0860 0200mr0mo000 000 000 000 000 000 000
The method parses the response (er, vl, bb/nn) and returns Barcode | None depending on whether a barcode was read. No changes are made to existing autoload barcode behavior.

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -5558,7 +5558,7 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
   async def core_read_barcode_of_picked_up_resource(
     self,
     rails: int,
-    reading_direction: Literal["vertical", "horizontal", "free"] = "vertical",
+    reading_direction: Literal["vertical", "horizontal", "free"] = "horizontal",
     minimal_z_position: float = 220.0,
     traverse_height_at_beginning_of_a_command: float = 275.0,
     z_speed: float = 128.7,
@@ -5569,7 +5569,7 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
 
     Args:
       rails: Rail/slot number where the barcode to be read is located (1-54).
-      reading_direction: Direction of barcode reading: 'vertical', 'horizontal', or 'free'. Default is 'vertical'.
+      reading_direction: Direction of barcode reading: 'vertical', 'horizontal', or 'free'. Default is 'horizontal'.
       minimal_z_position: Minimal Z position [mm] during barcode reading (220.0-360.0). Default is 220.0.
       traverse_height_at_beginning_of_a_command: Traverse height at beginning of command [mm] (0.0-360.0). Default is 275.0.
       z_speed: Z speed [mm/s] during barcode reading (0.0-128.7). Default is 128.7.
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_tests.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_tests.py
@@ -269,13 +269,13 @@ class TestSTARLiquidHandlerCommands(unittest.IsolatedAsyncioTestCase):
     set_tip_tracking(enabled=False)
 
   async def test_core_read_barcode_success(self):
-    """_star_core_read_barcode should send ZB and return a Barcode."""
+    """core_read_barcode_of_picked_up_resource should send ZB and return a Barcode."""
 
     self.STAR._write_and_read_command.return_value = (  # type: ignore
       "C0ZBid0001er00/00bb/08ABCDEFGH"
     )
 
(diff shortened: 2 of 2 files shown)


pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -5554,7 +5555,82 @@ async def core_move_plate_to_position(
...
+    if er_index == -1:
+      return None
+
+    error_code = resp[er_index + 2 : er_index + 7]
+    if error_code != "00/00":
+      return None
+
+    vl_index = resp.find("vl", er_index + 7)
+    bb_index = resp.find("bb/", er_index + 7)
+
```
rickwierenga: why do you need to pass `er_index + 7`?

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -5554,7 +5555,82 @@ async def core_move_plate_to_position(
...
+      mo="000 000 000 000 000 000 000",
+    )
+
+    if command_output is None:
+      return None
+
+    resp = command_output.strip()
+    er_index = resp.find("er")
+    if er_index == -1:
+      return None
```
rickwierenga: it would be good to raise the error to the user

you can use `self.check_fw_string_error(resp)` for this

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -5554,7 +5555,82 @@ async def core_move_plate_to_position(
...
+      vl_len = int(vl_str)
+    except ValueError:
+      vl_len = 0
+
+    bb_len_str = resp[bb_index + 3 : bb_index + 5]
+    try:
+      bb_len = int(bb_len_str)
+    except ValueError:
+      bb_len = 0
+
```
rickwierenga: I prefer to err on the side of letting users handle errors instead of unexpectedly returning None in cases where the data is of an unexpected format since the user will probably want to know this happened. Now it is difficult to differentiate between "no barcode" and "an error occurred"

rickwierenga (commented):
thanks and thank you for adding a test too!

some small comments below

what happens in the case when a barcode is not found? what does the machine return?

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
Dalconzo: we're skipping past the fixed length error field before searching for the barcode fields, error chunk is always erxx/xx for a total of 7, er_index points to the e so er_index + 7 makes sure we don't find anything in the error field matching our searches

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
Dalconzo: updated in new commit to raise error

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
Dalconzo: should be addressed in new commit

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -5554,7 +5555,86 @@ async def core_move_plate_to_position(
...
+    er_index = resp.find("er")
+    if er_index == -1:
+      # Unexpected format: no error section present.
+      raise ValueError(f"Unexpected CoRe barcode response (no error section): {resp}")
+
+    error_code = resp[er_index + 2 : er_index + 7]
+    if error_code != "00/00":
+      # Non-zero error code: raise a firmware error so the user sees it.
+      self.check_fw_string_error(resp)
+      return None
```
rickwierenga:
```suggestion
    self.check_fw_string_error(resp)
```

will check errors more generally (sometimes firmware responses are odd and this method is made to handle all of them)

rickwierenga:
```python
    # If the firmware returns more characters than declared, truncate to the declared length.
    if len(barcode_str) > bb_len:
      barcode_str = barcode_str[:bb_len]
```

why?

Dalconzo:
> ```python
>     # If the firmware returns more characters than declared, truncate to the declared length.
>     if len(barcode_str) > bb_len:
>       barcode_str = barcode_str[:bb_len]
> ```
> 
> why?

not getting the ability to comment on this directly for some reason. This is just in case the declared barcode length is different than the one actually supplied to us, deals with extra whitespace or padding after the end of the barcode without user intervention (probably just chat being extra cautious, we could probably get rid of it)

rickwierenga:
thank you! great first PR!

PR #752, opened 2025-11-17
BioShake Documentation

Nat-is-coding:
Here's the wiki documentation for using the BioShake backend. First time doing this, so please let me know what needs to be changed.

### DIFF
--- a/docs/user_guide/01_material-handling/heating_shaking/qinstruments.ipynb
+++ b/docs/user_guide/01_material-handling/heating_shaking/qinstruments.ipynb
@@ -61,14 +61,16 @@
    "source": [
     "from pylabrobot.heating_shaking import HeaterShaker\n",
     "from pylabrobot.heating_shaking import BioShake\n",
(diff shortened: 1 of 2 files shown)


rickwierenga:
looks good!

two small comments:

1. is there a reason to not use the `HeaterShaker` front end? You are using the backend directly
2. this

```python
for bs in [bs1, bs2, bs3, bs4]:
  await bs.setup(skip_home=False)
  await bs.shake(speed=500)
```

could be

```python
async def setup_and_shake(bs):
  await bs.setup(skip_home=False)
  await bs.shake(speed=500)

await asyncio.gather(
  setup_and_shake(bs1),
  setup_and_shake(bs2),
  setup_and_shake(bs3),
  setup_and_shake(bs4),
)
```

if you wanted to run them in parallel

Nat-is-coding:
1. Because I am still new to this and cannot tell the difference if I did or not, but it should be fixed now...?
2. Oh, nice!

rickwierenga:
thanks! will be live in 40 min or so

PR #753, opened 2025-11-18
`io.Socket` refactor

rickwierenga:
- don't have `read_once` in read, instead split into `read` and `read_until_eof` (not all higher level protocols require/support this)
- add `read_until` and `readline`
- write and read work with bytes, not strings
- save validation data in hex format
- update Proflex backend to new api


rickwierenga:
Thanks Bo

PR #758, opened 2025-11-21
add IMCS tips

rickwierenga:


pylabrobot/resources/imcs/tip_racks.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def imcs_tip() -> HamiltonTip:
+  """IMCS tip. Same as "Hamilton Standard volume (300 µL) tip without a filter", but tips are 6mm shorter."""
+  return HamiltonTip(
+    has_filter=True,
+    total_tip_length=59.9 - 6,  # - 6
+    maximal_volume=360,
+    tip_size=TipSize.STANDARD_VOLUME,
+    # requires "full_blowout" pickup method from STARBackend.pick_up_tips96. Not available on pip channels.
```
BioCam, line 13: Does this mean using these tips is not possible with the pip channels?

rickwierenga: Not right now. We can enable this in PLR by calling pickup on PX instead of C0, but for now the 96 head works fine.

It's really interesting this particular plunger option is not available on the channels though...

PR #759, opened 2025-11-21
added optional per-tip names, serialized them and kept tip type equality

### LINKED ISSUE #653: `Tip`s should have names
one step on the way to making them `Resource`s

Dalconzo:
Add optional name field to Tip and include it in serialize().
Have TipSpot generate unique per-tip names using a per-spot counter and, when available, the parent rack name.
Keep tip-type equality independent of name so Hamilton _get_hamilton_tip still groups tips by type.
Update Hamilton/Tecan tip constructors and tip tests to accept and round-trip name via serialization/deserialization.
closes #653

### DIFF
--- a/docs/resources/index.md
+++ b/docs/resources/index.md
@@ -179,6 +179,7 @@ library/corning
 library/eppendorf
 library/falcon
 library/hamilton
+library/imcs
 library/nest
 library/opentrons
 library/perkin_elmer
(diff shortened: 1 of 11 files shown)


rickwierenga:
is there a reason to have this optional? why not make it required?

Dalconzo:
making it required would break any current call sites I missed, old serialized tips, anything where someone creates their own tip instead of using TipSpot. All edge cases to be fair but I wanted to be cautious

rickwierenga:
ok nice yes. make add a deprecation warning when initialized without a name?

Dalconzo:
makes sense yeah give everyone a grace period before it's required

rickwierenga:
- added to repr
- add a `name` parameter to tip creators, this will be the best way in the future when tip is required at instant ion
  - check in TipSpot whether the function has one, if not warn
- serialization

PR #761, opened 2025-11-22
Add support for SSL in io.socket; fix ATC

xbtu2:
Add SSL connection mode for atc.
New read_until_eof does not work for atc. fixed


ben-ray:
I can't believe you scooped me by 4 minutes on this PR :p

rickwierenga:
merged into https://github.com/PyLabRobot/pylabrobot/tree/tftc-ssl so I can merge #762 into that as well (combing changes)

PR #762, opened 2025-11-22
ssl backend for thermo / applied biosystems ATC automated thermal cycler

ben-ray:
My 384 ATC never was able to connect by IP alone in the .exe software, but could connect by SSL. This adds SSL connection as an option to PLR by inputting a hostname into thermal cycler setup, like so:

```
backend = ATCBackend(
    ip=my_instrument_ip,
    port=7443, # yours might be different
    hostname="my_instrument_serial_number",
)

atc = Thermocycler(
    name="atc384",
    backend=backend,
    size_x=0, size_y=0, size_z=0,
    child_location=Coordinate(0,0,0)
)

await atc.setup()
```

### DIFF
--- a/pylabrobot/thermocycling/thermo_fisher/thermo_fisher_thermocycler.py
+++ b/pylabrobot/thermocycling/thermo_fisher/thermo_fisher_thermocycler.py
@@ -211,32 +211,39 @@ class ThermoFisherThermocyclerBackend(ThermocyclerBackend, metaclass=ABCMeta):
     self,
     ip: str,
     port: int = 7000,
-    hostname: Optional[str] = None,
+    shared_secret: Optional[bytes] = None,
+    serial_number: Optional[str] = None,
   ):
     self.ip = ip
     self.port = port
-    self.hostname = hostname
 
-    if hostname is not None:  # ssl connect
-      self.is_ssl = True
-      self.device_shared_secret = b"53rv1c3" + hostname.encode("ascii")
+    if port == 7443:
+      self.use_ssl = True
+      if shared_secret is None:
+        if serial_number is None:
+          raise ValueError("Serial number is required for SSL connection (port 7443)")
+        self.device_shared_secret = f"53rv1c3{serial_number}".encode("utf-8")
+      else:
+        self.device_shared_secret = shared_secret
 
-      ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
-      ctx.check_hostname = False
-      ctx.verify_mode = ssl.CERT_NONE
-      ctx.minimum_version = ssl.TLSVersion.TLSv1
-      ctx.maximum_version = ssl.TLSVersion.TLSv1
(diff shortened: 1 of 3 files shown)


pylabrobot/thermocycling/thermo_fisher/atc.py
```diff
@@ -7,13 +7,13 @@ class ATCBackend(ThermoFisherThermocyclerBackend):
   async def close_lid(self):
     if self.bid != "31":
       raise NotImplementedError("Lid control is only available for BID 31 (ATC)")
-    res = await self.send_command({"cmd": "lidclose"}, response_timeout=20, read_once=False)
+    res = await self.send_command({"cmd": "lidclose"}, response_timeout=20, read_once=True)
```
ben-ray, line 10: At least when using SSL, this command will timeout if I use read_once = False, but with read_once = True, the command completes when the instrument is physically finished with the movement

pylabrobot/thermocycling/thermo_fisher/atc.py
```diff
@@ -7,13 +7,13 @@ class ATCBackend(ThermoFisherThermocyclerBackend):
...
     if self._parse_scpi_response(res)["status"] != "OK":
       raise ValueError("Failed to close lid")
 
   async def open_lid(self):
     if self.bid != "31":
       raise NotImplementedError("Lid control is only available for BID 31 (ATC)")
-    res = await self.send_command({"cmd": "lidopen"}, response_timeout=20, read_once=False)
+    res = await self.send_command({"cmd": "lidopen"}, response_timeout=25, read_once=True)
```
ben-ray, line 17: it takes more than 20 seconds to open lid (~22 for me), hence increasing timeout to 25

xbtu2:
my pr for the same function was submitted minutes ago lol

pylabrobot/thermocycling/thermo_fisher/thermo_fisher_thermocycler.py
```diff
@@ -206,11 +207,36 @@ def stage_to_scpi(stage: Stage, stage_index: int, stage_name_prefix: str) -> dic
...
+  ):
     self.ip = ip
     self.port = port
-    self.device_shared_secret = shared_secret
-    self.io = Socket(host=ip, port=port)
+    self.hostname = hostname
+
+    if hostname is not None:  # ssl connect
+      self.is_ssl = True
+      self.device_shared_secret = b"53rv1c3" + hostname.encode("ascii")
```
ben-ray: the instrument uses a different shared secret for ssl compared to non-ssl, so I remove the shared-secret option from the backend so the connection always uses the correct shared secret for ssl vs non-ssl

pylabrobot/thermocycling/thermo_fisher/atc.py
rickwierenga: I think this is fixed in #761, but read_once should be fine here. what do you think @xbtu2 ?

PR #764, opened 2025-11-22
Fix atc available blocks detection

xbtu2:
Fixed logic for detecting available blocks on thermofisher thermocyclers. Add error for attempting to setup without available blocks


rickwierenga:
I can fix linting when I merge #763 into main and then main into this branch

rickwierenga:
thanks @xbtu2 !

PR #765, opened 2025-11-24
Integrate Inheco Incubator Shaker

BioCam:
Hi everyone,

New PyLabRobot integration: the Inheco Incubator Shaker

inheco_incubator_shaker_physical_setup_overview

I've added a thorough docs tutorial to showcase how powerful this small machine is and how to use it programmatically.

Simple overview of the implementation:

Screenshot 2025-11-24 at 17 14 01

in code (see docs notebook for more details):

```python
from pylabrobot.storage.inheco import IncubatorShakerStack, InhecoIncubatorShakerStackBackend

iis_stack_backend = InhecoIncubatorShakerStackBackend(dip_switch_id = 2)
iis_stack = IncubatorShakerStack(backend=iis_stack_backend)

await iis_stack.setup(verbose=True)

# Connected to Inheco Incubator Shaker Stack on /dev/cu.usbserial-130
# DIP switch ID of bottom unit: 2
# Number of connected units: 2
# Unit composition: {0: 'incubator_shaker_mp', 1: 'incubator_shaker_dwp'}
```

---

TODO: test the resource definition I created for it (hence not yet in the docs tutorial, might require a recipe since it is getting a bit beyond basic control).

### DIFF
--- a/pylabrobot/__init__.py
+++ b/pylabrobot/__init__.py
@@ -1,5 +1,6 @@
 import datetime
 import logging
+import sys
 from pathlib import Path
 from typing import Optional, Union
 
@@ -64,3 +65,21 @@ def configure(cfg: Config):
 
 
 configure(CONFIG)
+
+
+def verbose(make_verbose: bool, level: int = logging.INFO) -> None:
+  """Add a StreamHandler to the pylabrobot logger to make logging output visible to the console.
+  If set to False, remove the console StreamHandler. This only removes StreamHandlers that output
+  to sys.stdout or sys.stderr, and will not affect FileHandlers or other subclasses.
+  """
+  logger = logging.getLogger("pylabrobot")
+  for handler in logger.handlers[:]:  # iterate over a copy to allow safe removal
+    if isinstance(handler, logging.StreamHandler) and not isinstance(handler, logging.FileHandler):
+      if handler.stream in (sys.stdout, sys.stderr):
+        logger.removeHandler(handler)
+  if make_verbose:
+    logger.setLevel(level)
+    handler = logging.StreamHandler()
+    handler.setLevel(level)
+    handler.setFormatter(logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s"))
+    logger.addHandler(handler)
--- a/pylabrobot/plate_reading/__init__.py
+++ b/pylabrobot/plate_reading/__init__.py
@@ -1,4 +1,11 @@
-from .biotek_backend import Cytation5Backend, Cytation5ImagingConfig
+from .agilent_biotek_backend import BioTekPlateReaderBackend
+from .agilent_biotek_cytation_backend import (
+  Cytation5Backend,
+  Cytation5ImagingConfig,
+  CytationBackend,
+  CytationImagingConfig,
+)
+from .agilent_biotek_synergyh1_backend import SynergyH1Backend
 from .chatterbox import PlateReaderChatterboxBackend
 from .clario_star_backend import CLARIOstarBackend
 from .image_reader import ImageReader
--- a/pylabrobot/plate_reading/agilent_biotek_backend.py
+++ b/pylabrobot/plate_reading/agilent_biotek_backend.py
@@ -1,0 +1,595 @@
+import asyncio
+import enum
+import logging
+import time
+from typing import Dict, Iterable, List, Optional, Tuple
+
+from pylabrobot.io.ftdi import FTDI
+from pylabrobot.plate_reading.backend import PlateReaderBackend
+from pylabrobot.resources import Plate, Well
+
+logger = logging.getLogger(__name__)
+
+
+class BioTekPlateReaderBackend(PlateReaderBackend):
+  """Backend for Agilent BioTek plate readers."""
+
+  def __init__(
+    self,
+    timeout: float = 20,
+    device_id: Optional[str] = None,
+  ) -> None:
+    super().__init__()
+    self.timeout = timeout
+
+    self.io = FTDI(device_id=device_id)
+
+    self._version: Optional[str] = None
+
+    self._plate: Optional[Plate] = None
+    self._shaking = False
+    self._slow_mode: Optional[bool] = None
+
+  def _non_overlapping_rectangles(
+    self,
+    points: Iterable[Tuple[int, int]],
+  ) -> List[Tuple[int, int, int, int]]:
+    """Find non-overlapping rectangles that cover all given points.
+
+    Example:
+      >>> points = [
+      >>>   (1, 1),
+      >>>   (2, 2), (2, 3), (2, 4),
+      >>>   (3, 2), (3, 3), (3, 4),
+      >>>   (4, 2), (4, 3), (4, 4), (4, 5),
+      >>>   (5, 2), (5, 3), (5, 4), (5, 5),
+      >>>   (6, 2), (6, 3), (6, 4), (6, 5),
+      >>>   (7, 2), (7, 3), (7, 4),
+      >>> ]
+      >>> non_overlapping_rectangles(points)
+      [
+        (1, 1, 1, 1),
+        (2, 2, 7, 4),
+        (4, 5, 6, 5),
+      ]
+    """
+
+    pts = set(points)
+    rects = []
+
+    while pts:
+      # start a rectangle from one arbitrary point
+      r0, c0 = min(pts)
+      # expand right
+      c1 = c0
+      while (r0, c1 + 1) in pts:
+        c1 += 1
+      # expand downward as long as entire row segment is filled
+      r1 = r0
+      while all((r1 + 1, c) in pts for c in range(c0, c1 + 1)):
+        r1 += 1
+
+      rects.append((r0, c0, r1, c1))
+      # remove covered points
+      for r in range(r0, r1 + 1):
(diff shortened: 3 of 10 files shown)


pylabrobot/storage/inheco/incubator_shaker.py
```diff
@@ -0,0 +1,220 @@
...
+    return self._size_x
+
+  @property
+  def size_y(self) -> float:
+    return self._size_y
+
+  @property
+  def size_z(self) -> float:
+    return self._size_z
+
```
rickwierenga, line 36: why does this need to be rewritten instead of using the code in Resource?

pylabrobot/storage/inheco/incubator_shaker.py
```diff
@@ -0,0 +1,220 @@
...
+    "incubator_mp": 18 - chamber_z_clearance,
+    "incubator_shaker_mp": 50 - chamber_z_clearance,
+    "incubator_dwp": 18 - chamber_z_clearance,
+    "incubator_shaker_dwp": 53 - chamber_z_clearance,
+  }
+
+  incubator_power_credits_per_type = {
+    "incubator_mp": 1.0,
+    "incubator_dwp": 1.25,
+    "incubator_shaker_mp": 1.6,
```
rickwierenga: make these private?

pylabrobot/storage/inheco/incubator_shaker_backend.py
```diff
@@ -0,0 +1,1875 @@
...
+    id_product: str = "6001",
+    write_timeout: float = 5.0,
+    read_timeout: float = 10.0,
+    logger: Optional[logging.Logger] = None,
+  ):
+    # Logger
+    self.logger = logger or logging.getLogger("pylabrobot")
+    self.logger.setLevel(logging.INFO)
+    logging.getLogger("pylabrobot.io.serial").disabled = True
+
```
rickwierenga: why? let's keep it enabled unless there is a super strong reason not to

pylabrobot/storage/inheco/incubator_shaker_backend.py
```diff
@@ -0,0 +1,1875 @@
...
+    # Core state
+    self.id_vendor = id_vendor
+    self.id_product = id_product
```
rickwierenga: vid and pid are stored in IO

pylabrobot/storage/inheco/incubator_shaker_backend.py
```diff
@@ -0,0 +1,1875 @@
...
+    self.dip_switch_id = dip_switch_id
+
+    self.port_hint = port
+    self.io: Optional[Serial] = None
+    self.port: Optional[str] = None
```
rickwierenga: should be in io for port and port hint

io can be initialized (but not set up) during init time of the backend

pylabrobot/storage/inheco/incubator_shaker_backend.py
```diff
@@ -0,0 +1,1875 @@
...
+    for unit_index in range(self.number_of_connected_units):
+      if self.temperature_control_status[unit_index]:
+        await self.stop_temperature_control(stack_index=unit_index)
+
+      if self.shaking_status[unit_index]:
+        await self.stop_shaking(stack_index=unit_index)
+
+    if self.io:
+      await self.io.stop()
+      self.io = None
```
rickwierenga: could you use asyncio.gather to parallelize stopping them? or does that not work

pylabrobot/storage/inheco/incubator_shaker_backend.py
```diff
@@ -0,0 +1,1875 @@
...
+      d = byte
+      for _ in range(8):
+        if (d ^ crc) & 1:
+          crc ^= 0x18
+          crc >>= 1
+          crc |= 0x80
+        else:
+          crc >>= 1
+        d >>= 1
+    return crc & 0xFF
```
rickwierenga, line 306: I don't think there's a good pattern in PLR right now but ideally we share this with other inheco implementations

pylabrobot/storage/inheco/incubator_shaker_backend.py
```diff
@@ -0,0 +1,1875 @@
...
+
+# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
+# IncubatorShakerUnit Class
+# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
+
+
+class InhecoIncubatorShakerUnit:
+  """High-level API for an individual Inheco Incubator/Shaker unit in a stacked system."""
+
+  def __init__(self, backend, index: int):
```
rickwierenga: why not make Unit the Machine (considering it has a backend) and the Stack just a higher level resource?

Right now the unit can't be used as an individual unit really, or at least it seems annoying to do so

it forces people to use a stack

I can imagine but maybe I am wrong that people use just one unit sometimes

rickwierenga (commented):
some initial comments

pylabrobot/storage/inheco/incubator_shaker_backend.py
BioCam: How similar are they?

pylabrobot/storage/inheco/incubator_shaker_backend.py
BioCam: Unfortunately no - I've created a note in the tutorial about this: 

image

pylabrobot/storage/inheco/incubator_shaker_backend.py
BioCam: unit does not have a backend - everything goes through the stack by firmware definition.
If we wanted to make every unit a machine then the stack would become the "interface", which wouldn't reduce the number of lines to code but would remove the master commands for when people use multiple units in a stack.

I agree that people might use just a single incubator (shaker) but by firmware one would have to always say index=0 (called "subdevice" in the documentation) in every firmware command.

pylabrobot/storage/inheco/incubator_shaker.py
BioCam: I am not sure tbh, I was quite surprised when I noticed these are private attributes?

rickwierenga: private because we don't want people to change them after initialization

for accessing them there is:
- `get_size_{x,y,z}`: get size in its own space
- `get_absolute_size_{x,y,z}`: size in absolute space

pylabrobot/storage/inheco/incubator_shaker_backend.py
rickwierenga: see example: https://github.com/PyLabRobot/pylabrobot/blob/main/pylabrobot/temperature_controlling/inheco/control_box.py#L7

pylabrobot/storage/inheco/incubator_shaker.py
BioCam: Thank you for the explanation - that makes a lot of sense; I've removed this

pylabrobot/storage/inheco/incubator_shaker.py
BioCam: done

pylabrobot/storage/inheco/incubator_shaker_backend.py
BioCam: e.g.:

Screenshot 2025-11-26 at 16 54 01

Screenshot 2025-11-26 at 16 55 24

pylabrobot/storage/inheco/incubator_shaker_backend.py
BioCam: fixed - merged into `pylabrobot.io.Serial`

pylabrobot/storage/inheco/incubator_shaker_backend.py
BioCam: TODO for future: find a way around gather lock

pylabrobot/storage/inheco/incubator_shaker_backend.py
BioCam: TODO: later refactor for unification

pylabrobot/storage/inheco/incubator_shaker_backend.py
```diff
@@ -0,0 +1,1778 @@
...
+      layer: Layer index (0-7). Default = 0.
+
+    Returns:
+      Number of connected devices (0-16).
+
+    Example:
+      Response "3" → 3 connected devices on that layer.
+    """
+    if not (0 <= layer <= 7):
+      raise ValueError(f"Layer must be 0-7, got {layer}")
```
rickwierenga, line 500: what is a "layer"?

BioCam: @rickwierenga, I must admit, I do not know - I tried to faithfully reproduce the firwmare command (openly shared by the OEM):

Screenshot 2025-12-01 at 16 47 58

I do not know how to interpret the meaning of "layer" in this description 😅

pylabrobot/storage/__init__.py
rickwierenga, line 1: ?

BioCam:
Thank you @rickwierenga for the thorough review and working with me on this PLR integration 😊

PR #766, opened 2025-11-25
fix atc num of temp zones

xbtu2:
ATC is supposed to have 3 temperature zones instead of 1. Not specifying all 3 temperatures causes early termination of protocols


BioCam:
@xbtu2, How often are people using the 3 temperature zones separately compared to using them all together?

Do you think it would be worth autofilling the temperature zones if only 1 is given?

e.g. if one uses a  96-wellplate and only declares a single temperature at each step (which seems to me the "status quo"), then it will fill the remaining 2 zones with the same temperature.

I think this would facilitate machine agnosticity for thermocycling protocols.

xbtu2:
@BioCam you are probably right. the 3 temperature zones are not even shown in the vendor demo software. i think i will add a param for multi-zone for `run_protocol` in the backend and just give a warning if temps are autofilled

PR #767, opened 2025-11-26
Add manual channel jog helpers for OT-2

hazlamshamin:


pylabrobot/liquid_handling/backends/opentrons_backend.py
```diff
@@ -579,6 +579,69 @@ async def list_connected_modules(self) -> List[dict]:
...
+    """Move a channel to an absolute y coordinate using savePosition to seed pose."""
+
+    pipette_id = self._pipette_id_for_channel(channel)
+    try:
+      res = ot_api.lh.save_position(pipette_id=pipette_id)
+      pos = res["data"]["result"]["position"]
+      current = Coordinate(pos["x"], pos["y"], pos["z"])
+    except Exception as exc:  # noqa: BLE001
+      raise NoChannelError("Failed to query current pipette position") from exc
+
```
rickwierenga: instead of `NoChannelError` should we use `RuntimeError` here?

pylabrobot/liquid_handling/backends/opentrons_backend.py
```diff
@@ -579,6 +579,69 @@ async def list_connected_modules(self) -> List[dict]:
...
+  async def move_channel_y(self, channel: int, y: float):
+    """Move a channel to an absolute y coordinate using savePosition to seed pose."""
+
+    pipette_id = self._pipette_id_for_channel(channel)
+    try:
+      res = ot_api.lh.save_position(pipette_id=pipette_id)
+      pos = res["data"]["result"]["position"]
+      current = Coordinate(pos["x"], pos["y"], pos["z"])
+    except Exception as exc:  # noqa: BLE001
+      raise NoChannelError("Failed to query current pipette position") from exc
```
rickwierenga: it would be nice to abstract this out into a `get_current_position(channel)` that is used by all 3 move commands. It would simplify the code and also be an individually nice function

rickwierenga:
great addition @hazlamshamin! just two small comments about the implementation

rickwierenga:
thank you!! super useful feature

PR #770, opened 2025-12-02
Hamilton 60ml trough carrier

nedru004:
Added hamilton 5x60ml trough carrier. Hamilton cat. no.: 53646-01. Measured locations of resources with caliper. Will carry hamilton_1_trough_60ml_Vb which is already in the pylabrobot resources.

### DIFF
--- a/pylabrobot/resources/hamilton/trough_carriers.py
+++ b/pylabrobot/resources/hamilton/trough_carriers.py
@@ -65,5 +65,5 @@ def Trough_CAR_5R60_A00(name: str) -> TroughCarrier:
       # pedestal_size_z=1.5, # measured
       name_prefix=name,
     ),
-    model="Trough_CAR_4R200_A00",
+    model=Trough_CAR_5R60_A00.__name__,
   )


pylabrobot/resources/hamilton/trough_carriers.py
```diff
@@ -39,3 +39,31 @@ def Trough_CAR_4R200_A00(name: str) -> TroughCarrier:
...
+        Coordinate(1.5, 199.0, 62.0 + 1.5),
+        Coordinate(1.5, 302.0, 62.0 + 1.5),
+      ],  # measured 62 to bottom of holder, but there is a 1.5mm pedestal
+      resource_size_x=19.0,
+      resource_size_y=90.0,
+      # pedestal_size_z=1.5, # measured
+      name_prefix=name,
+    ),
+    model="Trough_CAR_4R200_A00",
+  )
```
rickwierenga, line 69: I think the model should be this:

```suggestion
    ),
    model=Trough_CAR_5R60_A00.__name__,
  )
```

nedru004: Yes. That was a carry over from using 4R200 as a template.

rickwierenga:
Thanks!

PR #774, opened 2025-12-02
Resource Integration: `Greiner_384_wellplate_28ul_Fb`

BioCam:


docs/resources/library/greiner.md
```diff
@@ -0,0 +1,13 @@
+# Revvity
+
```
rickwierenga:
```suggestion
# Greiner
```

pylabrobot/resources/greiner/plates.py
```diff
@@ -0,0 +1,125 @@
...
+  # vol_ul = interpolate_1d(
+  #   h, _Greiner_384_wellplate_28ul_Fb_height_to_volume_measurements, bounds_handling="error"
+  # )
+  # return round(max(0.0, vol_ul), 3)
+  raise NotImplementedError(
+    "Computation of volume from height is not currently defined. "
+    "It is difficult (perhaps impossible?) to robustly perform liquid "
+    "level detection in such a small well AND such low volumes."
+  )
+
```
rickwierenga: is it defined or not?

BioCam: It is defined but the cLLD capabilities of the STAR keep failing - I take at least 3 cLLD measurements for each point at the calibration curve, occasionally our STAR manages to take one measurement down to 8uL, and fails in the repeats - interpretation: the well geometry and low volume together create a "charge dissipation" target that is not being recognised by the charge sensor.

Keeping this commented out code, would mean others can try as well and test whether their system is more sensitive, and compare.

This might be interesting to see with the Zeus X1 pipettes 👀

rickwierenga: I don't think we should be checking in uncommented code because it is very vague what the purpose of it is, and it's not discoverable. does that mean the measurements are also inaccurate? if so, we should not have this code. if not, why not have it as real code and then have a warning saying "clld on star does not work on this container"? that way people can use/debug as they like

pylabrobot/resources/greiner/plates.py
```diff
@@ -0,0 +1,125 @@
...
+  # See https://github.com/PyLabRobot/pylabrobot/pull/161.
+  # return Lid(
+  #   name=name,
+  #   size_x=127.76,
+  #   size_y=85.48,
+  #   size_z=None,           # measure the total z height
+  #   nesting_z_height=None, # measure overlap between lid and plate
+  #   model="Greiner_384_wellplate_28ul_Fb_Lid",
+  # )
+
```
rickwierenga, line 38: why have this commented out def?

pylabrobot/resources/greiner/plates.py
BioCam: Yeah fair enough, I wanted to highlight somehow that someone _has_ previously tried cLLD on this well instance and found the STAR is not capable of detecting it - fighting people's tendency to leave out very useful negative data.

But if it's not clear then, I agree, there's no point 😅

BioCam: Yeah fair enough, I wanted to highlight somehow that someone _has_ previously tried cLLD on this well instance and found the STAR is not capable of detecting it - fighting people's tendency to leave out very useful negative data.

But if it's not clear then, I agree, there's no point 😅

pylabrobot/resources/greiner/plates.py
BioCam: that is actually just the copy from **all** definitions - I am not sure why I found them like this originally a long time ago, but I assumed it's for information purposes?

rickwierenga: not all, but many are commented out because we know the part that exists is correct but they were missing the `nesting_z_height` parameter which you added in #161 . are the size x and y correct here? if so we can keep it in anticipation of someone adding the missing parameters

pylabrobot/resources/greiner/plates.py
rickwierenga: remove it?

pylabrobot/resources/greiner/plates.py
BioCam: Ohhh thank you for the reminder! I forgot about the `nesting_z_height` we solved, been a long time 😅 

No, I am unaware of a Lid existing for this plate - I am removing it and if someone finds one they can add it easily :)

PR #775, opened 2025-12-03
Opentrons plate holder

nedru004:
I would like to add the opentrons deep well plate aluminum block. I found a bunch of opentrons blocks added but not this one. Plus it is confusing how it pulls the data from opentrons repository instead of hardcoding into pylabrobot. Is this ok to add to the opentrons module.py file instead of figuring out how it pulls data from a website?

### DIFF
--- a/docs/resources/library/opentrons.md
+++ b/docs/resources/library/opentrons.md
@@ -53,3 +53,8 @@ Unfortunately, most of the other labware (plates) is missing information that is
 - `opentrons_96_aluminumblock_generic_pcr_strip_200ul`
 - `opentrons_96_aluminumblock_nest_wellplate_100ul`
 - `opentrons_96_well_aluminum_block`
+
+
+## Plate Adapters
+
+- `opentrons_96_deep_well_temp_mod_adapter`


rickwierenga:
why not load the definition from the OT library? this way we are not responsible for maintaining these definitions. (for OT plates, unfortunately they are not all correct and/or miss some information, but for their own labware I think we should trust them...)

there is a `_download_ot_resource_file` helper function. you will need to write an OT-json -> PlateAdapter / PlateHolder convertor

rickwierenga:
@nedru004 just confirming this definition works as well as the hardcoded one?

nedru004:
Yes, but I still needed to hardcode the z coordinate for the plate. I couldn't find where that was located in the json file. Here is a screenshot. I used cornerOffsetFromSlot for the plate location. Not sure how opentrons calculates z position. Any ideas? or should I just hardcode it?
image
I also tested this on a windows machine and the json downloads to /tmp folder, but that doesnt exist on Windows. I added a change for that.

Finally, when loading the opentrons temp module plus the adapter, it is too high and the tips will run into the deep well plate. It outputs a warning when loading the plate. I assume there is a travel height adjustment?

rickwierenga:
> I still needed to hardcode the z coordinate for the plate

unfortunately the OT definitions don't have enough information for complete automation / or to work on different robotic systems

> I used cornerOffsetFromSlot for the plate location.

If it works, it works. the cornerOffsetFromSlot should be the location between the slot and the resource, but I guess it's repurposed for the plate location in this case. (they do that kind of stuff, so if that's how it's done then it's good)

> I also tested this on a windows machine and the json downloads to /tmp folder, but that doesnt exist on Windows. I added a change for that.

Thanks!!

> I assume there is a travel height adjustment?

Yes, `ot2_backend.traversal_height = 120`

https://docs.pylabrobot.org/user_guide/00_liquid-handling/opentrons-ot2/hello-world.html

nedru004:
> Finally, when loading the opentrons temp module plus the adapter, it is too high and the tips will run into the deep well plate. It outputs a warning when loading the plate. I assume there is a travel height adjustment?

I meant to say too high on the hamilton. I assume that it is fine for the opentrons.
Is this the backend command? STAR.set_minimum_channel_traversal_height()

rickwierenga:
> STAR.set_minimum_channel_traversal_height()

yes you got it

set_minimum_channel_traversal_height and set_minimum_iswap_traversal_height

rickwierenga:
apologies for the delay - thanks for adding this!

PR #781, opened 2025-12-05
Add warning to hamilton troughs

nedru004:
The hamilton trough has a center support that will collide with pipette tips if using 1 or odd number of channels. I added a warning and a solution.

Address this conversation:
https://discuss.pylabrobot.org/t/hamilton-trough-center-support-interference/399/2


rickwierenga:
thanks!

PR #787, opened 2025-12-11
Upgrade `.request_pip_height_last_lld()`

BioCam:
Hi everyone,

`STAR.request_pip_height_last_lld()` currently returns the dictionary of the parsed firmware response.

I make it return the desired list of latest-detected liquid levels in mm (with regards to the deck) instead.

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -5939,9 +5939,7 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
     resp = await self.send_command(module="C0", command="RL", fmt="lh#### (n)")
 
     liquid_levels = resp.get("lh")
-    assert isinstance(
-      liquid_levels, list
-    ), f"Expected liquid_levels to be a list, got {type(liquid_levels).__name__} instead"
+
     assert (
       len(liquid_levels) == self.num_channels
     ), f"Expected {self.num_channels} liquid level values, got {len(liquid_levels)} instead"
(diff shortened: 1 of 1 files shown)


pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -8936,10 +8964,14 @@ async def clld_probe_z_height_using_channel(
...
       await self.move_all_channels_in_z_safety()
 
-    get_llds = await self.request_pip_height_last_lld()
-    result_in_mm = float(get_llds["lh"][channel_idx] / 10)
+    current_absolute_liquid_heights = await self.request_pip_height_last_lld()
+    result_probed_z_height = current_absolute_liquid_heights[channel_idx]
+
+    assert isinstance(result_probed_z_height, float), (
+      "Expected probed z-height to be a float, got " f"{type(result_probed_z_height)} instead."
+    )
```
rickwierenga: why do we need this

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -5913,14 +5914,41 @@ async def request_tip_presence(self) -> List[int]:
...
+        AssertionError: If the instrument response does not contain a valid
+        ``"lh"`` list.
     """
+    resp = await self.send_command(module="C0", command="RL", fmt="lh#### (n)")
+
+    liquid_levels = resp.get("lh")
+    assert isinstance(
+      liquid_levels, list
+    ), f"Expected liquid_levels to be a list, got {type(liquid_levels).__name__} instead"
+    assert (
```
rickwierenga: why do we need this? if the response does not match the format, we will raise an error when parsing the response

BioCam: removed

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: removed

PR #788, opened 2025-12-11
Hamilton verify deck is loaded

nedru004:
This is a function to call after assigning all carriers to a hamilton deck. This will check if all carriers are loaded and in the correct position. If not it will blink the LEDs until those carriers are loaded. In relation to this conversion: https://discuss.pylabrobot.org/t/hamilton-deck-led/402/6


rickwierenga:
hi David, apologies for the late response. I was super busy this last month preparing for slas, but finally finished that (for now).

I just tested this function and it works perfectly!

curious to get your opinion: should we print out when unexpected carriers are detected even when all expected carriers are loaded?

rickwierenga:
nice feature, thanks!

PR #791, opened 2025-12-12
Synergy H1 documentation

Nat-is-coding:
Added documentation for the Synergy H1 reader, as well as modified the Cytation documentation accordingly. Also, a minor edit on the Synergy H1 backend.


rickwierenga:
thank you @Nat-is-coding!

PR #682, opened 2025-09-25
Probe y-position of resource using channel with cLLD

BioCam:
This is take 2; previous PR: https://github.com/PyLabRobot/pylabrobot/pull/353

Updated the tip_bottom_diameter -> simply declare it and default to the 1.2 mm diameter for the teaching_needles since these are the most likely tool used for this function

PR #792, opened 2025-12-15
Create `STAR.clld_probe_x_position_using_channel()`

BioCam:
Hi everyone,

In previous PRs I've added different object probing methods to the `STARBackend` class:

**Z-Probing**
- [PR#69: Probe z height using channel (cLLD)](https://github.com/PyLabRobot/pylabrobot/pull/69)
- [PR260: Build ztouch Probing Function (force sensing)](https://github.com/PyLabRobot/pylabrobot/pull/260)

**Y-Probing**
- [PR#682: Probe y-position of resource using channel with cLLD](https://github.com/PyLabRobot/pylabrobot/pull/682)

## This PR

Here I am completing the series by adding **X-Probing** using cLLD:

```python
  async def clld_probe_x_position_using_channel(
    self,
    channel_idx: int,  # 0-based indexing of channels!
    probing_direction: Literal["right", "left"],
    start_pos_search: Optional[float] = None,  # mm
    end_pos_search: Optional[float] = None,  # mm
    move_to_safe_z_before_move_to_start_pos: bool = True,
    post_detection_dist: float = 2.0,  # mm,
    tip_bottom_diameter: float = 1.2,  # mm
    read_timeout=240.0,  # seconds
  ) -> float
```

Warnings this `STARBackend` method...
- can ONLY be used to probe _conductive_ materials (!),
- assumes you are using a Hamilton 1_000uL teaching needle for probing - if you use a different conductive tip, you have to modify the `tip_bottom_diameter` argument,
- should be used with caution - improper usage could break the channel.

Though, I've measured the default speed of this method to be 0.45 mm / second - so you have plenty of time to hit the emergency stop button if you e.g. accidentally made it probe a non-conductive material (e.g. a plate, tube, trough, ...).

**Power**
Together with z-probing and y-probing, x-probing now enables PyLabRobot programmers to make the STAR measure its own environment... and correct itself if something differs to the workcell model 👀 

We are now deep in PLR's "autonomation" stage 🚀 

Please let us know your thoughts, and whether you encounter any issues.

Happy automation 🦾

### DIFF
--- a/docs/user_guide/machines.md
+++ b/docs/user_guide/machines.md
@@ -47,8 +47,9 @@ tr > td:nth-child(5) { width: 15%; }
 |--------------|---------|----------|-------------|--------|
 | Hamilton | STAR(let) | <span class="badge badge-transfer">arm</span> | Full | [PLR](https://docs.pylabrobot.org/user_guide/00_liquid-handling/hamilton-star/_hamilton-star.html) / [OEM](https://www.hamiltoncompany.com/microlab-star) |
 | Hamilton | Vantage | <span class="badge badge-transfer">arm</span> | Mostly | [PLR](https://docs.pylabrobot.org/user_guide/00_liquid-handling/hamilton-vantage/_hamilton-vantage.html) / [OEM](https://www.hamiltoncompany.com/microlab-vantage) |
-| Hamilton | Prep | <span class="badge badge-transfer">arm</span> | WIP | [OEM](https://www.hamiltoncompany.com/microlab-prep) |
-| Tecan | EVO |  | Basic | [PLR](https://docs.pylabrobot.org/user_guide/00_liquid-handling/tecan-evo/_tecan-evo.html) / [OEM](https://lifesciences.tecan.com/freedom-evo-platform) |
+| Hamilton | Prep | | WIP | [OEM](https://www.hamiltoncompany.com/microlab-prep) |
+| Hamilton | Nimbus | <span class="badge badge-transfer">arm</span> | WIP | [OEM](https://www.hamiltoncompany.com/microlab-nimbus) |
+| Tecan | EVO | <span class="badge badge-transfer">arm</span> | Basic | [PLR](https://docs.pylabrobot.org/user_guide/00_liquid-handling/tecan-evo/_tecan-evo.html) / [OEM](https://lifesciences.tecan.com/freedom-evo-platform) |
 | Opentrons | OT-2 |  | Mostly | [PLR](https://docs.pylabrobot.org/user_guide/00_liquid-handling/opentrons-ot2/_opentrons-ot2.html) / [OEM](https://opentrons.com/products/ot-2-robot) |
 
 ### Pumps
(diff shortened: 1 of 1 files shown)


rickwierenga:
Completing the Trinity!!

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -8663,6 +8663,143 @@ def mm_to_z_drive_increment(value_mm: float) -> int:
...
+    num_rails = self.extended_conf["xt"]
+    track_width = 22.5  # mm
+    reachable_dist_to_last_rail = 125.0
+
+    max_safe_upper_x_pos = num_rails * track_width + reachable_dist_to_last_rail
+    max_safe_lower_x_pos = 95.0  # unit: mm
+
+    if start_pos_search is None:
+      start_pos_search = current_x_position
+    else:
```
rickwierenga: given how powerful the x-arm is, I would prefer having the user move to start_pos themselves and only support x-probing from the current position. that means removing the start_pos_search parameter and always starting at current_x_position

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -8663,6 +8663,143 @@ def mm_to_z_drive_increment(value_mm: float) -> int:
...
+      final_x_pos = sensor_triggered_x_pos - post_detection_dist
+
+      material_x_pos = sensor_triggered_x_pos + tip_bottom_diameter / 2
+
+    # final_x_pos is guaranteed to lie within safe bounds because:
+    # - start_pos_search and end_pos_search are bounded
+    # - XL motion is constrained to that interval
+    # - post_detection_dist always moves inward
+    await self.move_channel_x(x=final_x_pos, channel=channel_idx)
+
```
rickwierenga: why is this necessary?

BioCam: Do you mean the post detection distance or the comment?

BioCam: Do you mean the post detection distance or the comment?

rickwierenga: the movement? wouldn't it already be at the location at the end? why do you need to move it more?

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: Been giving this some thought and I agree with you, I have removed `star_pos_search` now, which also made this method a bit more concise :)

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: I've changed the comment for it to make this clearer:

```python
    # Move away from detected object to avoid mechanical interference
    # e.g. touch carrier, then carrier moves while tip still touching carrier -> friction on channel!
    await self.move_channel_x(x=final_x_pos, channel=channel_idx)
```

Also, we should have consistency across all probing commands as much as possible. The `post_detection_dist` argument is one example.

Also, even with the current slowness of the probing, there can be some slight pressure exerted on the object during the detection process (same as Y and Z probing). That pressure should be released as soon as possible, hence the need for a `post_detection_dist`.

PR #793, opened 2025-12-17
XPeel Support

rkaptur:
Python interface for controlling the Automated Plate Seal Remover (formerly XPeel). All commands defined in the manufacturer’s manual are implemented. The backend has been tested against firmware version 2.11.

### DIFF
--- a/pylabrobot/resources/utils.py
+++ b/pylabrobot/resources/utils.py
@@ -1,6 +1,7 @@
 import re
+from itertools import groupby
 from string import ascii_uppercase as LETTERS
-from typing import Dict, List, Optional, Type, TypeVar
+from typing import Any, Dict, List, Optional, Type, TypeVar
 
 from pylabrobot.resources.coordinate import Coordinate
 from pylabrobot.resources.resource import Resource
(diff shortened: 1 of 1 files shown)


pylabrobot/peeling/xpeel_backend.py
```diff
@@ -0,0 +1,309 @@
...
+        full_cmd.strip(),
+        self.response_timeout,
+      )
+
+    return responses
+
+  async def get_status(self):
+    """Request instrument status; returns *ready:XX,XX,XX."""
+    self.logger.debug("Requesting status...")
+    return await self._send_command("*stat")
```
rickwierenga: what is `XX,XX,XX`?

pylabrobot/peeling/xpeel_backend.py
```diff
@@ -0,0 +1,309 @@
...
+    return await self._send_command(
+      cmd,
+      expect_ack=True,
+      wait_for_ready=True,
+    )
+
+  async def seal_check(self):
+    """Check for seal presence; ready response encodes result."""
+    self.logger.debug("Checking for seal presence...")
+    return await self._send_command("*sealcheck", expect_ack=True, wait_for_ready=True)
```
rickwierenga: does this return a bool?

pylabrobot/peeling/xpeel_backend.py
```diff
@@ -0,0 +1,309 @@
...
+
+  async def get_tape_remaining(self):
+    """Query remaining tape."""
+    self.logger.debug("Querying remaining tape...")
+    return await self._send_command("*tapeleft", expect_ack=True, wait_for_ready=True)
```
rickwierenga: what does this return?

pylabrobot/peeling/xpeel_backend.py
```diff
@@ -0,0 +1,309 @@
...
+
+  async def get_seal_sensor_status(self):
+    """Get seal sensor threshold status."""
+    self.logger.debug("Getting seal sensor threshold status...")
+    return await self._send_command("*sealstat", expect_ack=True, wait_for_ready=True)
+
+  async def set_seal_threshold_upper(self, value: int):
+    """Set upper seal sensor threshold (0-999)."""
+    self.logger.debug(f"Setting upper seal sensor threshold to {value}...")
+    if not 0 <= value <= 999:
```
rickwierenga, line 290: what is the unit?

pylabrobot/peeling/xpeel_backend.py
```diff
@@ -0,0 +1,309 @@
...
+    flag = "y" if enabled else "n"
+    return await self._send_command(
+      f"*platecheck:{flag}",
+      expect_ack=True,
+      wait_for_ready=True,
+    )
+
+  async def get_seal_sensor_status(self):
+    """Get seal sensor threshold status."""
+    self.logger.debug("Getting seal sensor threshold status...")
```
rickwierenga, line 281: what are the possible status options?

rickwierenga:
made some changes:
- simulation is typically done by switching the backend rather than implementing this in an existing backend so I removed the simulation things. (in the future we should have a PeelerChatterboxBackend for this)
- moved print statements for every command to logger
- have explicit and human-readable parameters in the `peel` function so it's easier than setting the parameter directly. I chose some defaults that seem reasonable but please check those

pylabrobot/peeling/xpeel_backend.py
rkaptur: I think it’s an arbitrary unit returned by the seal sensor that defines the value **above** which the reading is considered “seal not detected.” With `set_seal_threshold_lower`, we can define the value **below** which the reading is considered “seal not detected.”

pylabrobot/peeling/xpeel_backend.py
rkaptur: Quoting the manual here: 

The ‘XX’ fields refer to the three possible error codes from the error table that occurred during the
previous Automated Plate Seal Remover motion. The error codes will remain in the ready response until
another motion command is made or until a restart command is sent. A single commanded action may
accumulate up to three errors. Since they are logged as they occur (left to right), any error code may
appear in any error field. Successful completion of any motion command will clear all three error codes
back to 00.

rickwierenga:
thanks for the information, I added response parsing based on your description.

ready to merge?

rkaptur:
thanks, looks good and ready to merge!

rickwierenga:
thank you! super helpful!

PR #797, opened 2025-12-20
Tecan Infinite 200 PRO plate reader backend (Infinite M Plex)

hazlamshamin:
- Add Tecan Infinite 200 PRO backend + tests.
- Supports absorbance and fluorescence, including masking to read specified list of wells, up to 384-well plate.
- Luminescence is experimental due to lack of glowing sample for testing.
- Make USB capture decoding resilient to malformed escape sequences.

EDIT: all three modes now reflect what OEM does with 100% accuracy, and luminescence is no longer experimental ([e92c6db])

Tests: make lint, make format-check, make typecheck, make test.

### DIFF
--- a/pylabrobot/plate_reading/tecan/infinite_backend_tests.py
+++ b/pylabrobot/plate_reading/tecan/infinite_backend_tests.py
@@ -498,7 +498,7 @@ def _egg_grid():
   ]
 
 
-class TestTecanInfiniteDecoders(unittest.TestCase):
+class TestTecanInfiniteDecoders(unittest.IsolatedAsyncioTestCase):
   def setUp(self):
     self.backend = ExperimentalTecanInfinite200ProBackend()
     self.plate = Plate_384_Well(name="plate")
@@ -590,7 +590,7 @@ class TestTecanInfiniteDecoders(unittest.TestCase):
     self._run_decoder_case(decoder, build_packet, extract_actual)
 
 
-class TestTecanInfiniteScanGeometry(unittest.TestCase):
+class TestTecanInfiniteScanGeometry(unittest.IsolatedAsyncioTestCase):
   def setUp(self):
     self.backend = ExperimentalTecanInfinite200ProBackend(counts_per_mm_x=1, counts_per_mm_y=1)
     self.plate = _make_test_plate()
@@ -621,7 +621,7 @@ class TestTecanInfiniteScanGeometry(unittest.TestCase):
     self.assertEqual((stage_x, stage_y), (3, 16))
 
 
-class TestTecanInfiniteAscii(unittest.TestCase):
+class TestTecanInfiniteAscii(unittest.IsolatedAsyncioTestCase):
   def test_frame_command(self):
     framed = ExperimentalTecanInfinite200ProBackend._frame_command("A")
     self.assertEqual(framed, b"\x02A\x03\x00\x00\x01\x40\x0d")


pylabrobot/plate_reading/tecan/infinite_backend.py
```diff
@@ -0,0 +1,1041 @@
...
+
+  async def open(self) -> None:
+    """Open the reader drawer."""
+
+    await self._send_ascii("ABSOLUTE MTP,OUT")
+    await self._send_ascii("BY#T5000")
+
+  async def close(self, plate: Optional[Plate]) -> None:  # noqa: ARG002
+    """Close the reader drawer."""
+
```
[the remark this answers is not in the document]
hazlamshamin: arguably, this is unneeded and intuitive. yes the parameter is indeed for interface compatibility but not needed to explain this in docstring.

pylabrobot/plate_reading/tecan_infinite_backend.py
```diff
@@ -0,0 +1,1041 @@
...
+def _consume_leading_ascii_frame(buffer: bytearray) -> Tuple[bool, Optional[str]]:
+  """Remove a leading STX...ETX ASCII frame if present."""
+
+  if not buffer or buffer[0] != 0x02:
+    return False, None
+  end = buffer.find(b"\x03", 1)
+  if end == -1:
+    return False, None
+  text = buffer[1:end].decode("ascii", "ignore")
+  del buffer[: end + 2]
```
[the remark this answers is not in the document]
hazlamshamin: i have pushed a new commit for extra guard for this after checking the hexdumps. additionally, there has already been guard for absorbance and fluorescence which checked that the consumed 

i pushed a new commit (ca73508f) for extra guard for this after checking the hexdumps. found cases where ETX arrives at end‑of‑packet (though this does not appear to change the measurement data). nonetheless the extra guard avoids consuming partial frames. also note in current code, we already validate measurement frames: ABS checks wavelength (words[1]), FI checks excitation/emission (words[1]/[2]). only LUM has no wavelength field to match.

pylabrobot/plate_reading/tecan/infinite_backend.py
```diff
@@ -0,0 +1,1407 @@
...
+      "MIRROR CLEAR",
+    ]
+    configure_cmds = [
+      f"EXCITATION 0,FI,{ex_decitenth},50,0",
+      f"EMISSION 0,FI,{em_decitenth},200,0",
+      "TIME 0,INTEGRATION=20",
+      "TIME 0,LAG=0",
+      "TIME 0,READDELAY=0",
+      "GAIN 0,VALUE=100",
+      "POSITION 0,Z=20000",
```
rickwierenga: >       "POSITION 0,Z=20000",

do you know what these units are? would be nice to parameterize them as focal_height

pylabrobot/plate_reading/tecan/infinite_backend.py
```diff
@@ -0,0 +1,1407 @@
...
+    }
+    commands = [
+      "READS CLEAR",
+      "EMISSION CLEAR",
+      "TIME CLEAR",
+      "GAIN CLEAR",
+      "POSITION CLEAR",
+      "MIRROR CLEAR",
+      "POSITION LUM,Z=14620",
+      "TIME 0,INTEGRATION=3000000",
```
rickwierenga: do you know what the integration time units are?

pylabrobot/plate_reading/tecan/infinite_backend.py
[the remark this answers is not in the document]
rickwierenga: agreed

rickwierenga:
could you please explain what `_MODE_CAPABILITY_COMMANDS` are?

are they just more parameters like this?

```
      f"BEAM DIAMETER={self._capability_numeric('ABS', '#BEAM DIAMETER', 700)}",
```

that we can add as commands? why not add them?

pylabrobot/plate_reading/tecan/infinite_backend.py
```diff
@@ -0,0 +1,1328 @@
...
+    try:
+      await self._send_ascii("QQ")
+    except TimeoutError:
+      logger.warning("QQ produced no response; continuing with initialization.")
+    await self._send_ascii("INIT FORCE")
+    self._device_initialized = True
+
+  async def _begin_run(self) -> None:
+    await self._initialize_device()
+    self._reset_stream_state()
```
rickwierenga, line 1020: is there a reason the device would un-initialize? every command calls `_begin_run`, which calls `_initialize_device`. but `_initialize_device` is already called on setup.

pylabrobot/plate_reading/tecan/infinite_backend.py
```diff
@@ -0,0 +1,1293 @@
...
+  """Scan configuration for Infinite plate readers."""
+
+  flashes: int = 25
+  counts_per_mm_x: float = 1_000
+  counts_per_mm_y: float = 1_000
+
+
+def _integration_value_to_seconds(value: int) -> float:
+  return value / 1_000_000.0 if value >= 1000 else value / 1000.0
+
```
rickwierenga: can you explain this?

pylabrobot/plate_reading/tecan/infinite_backend.py
hazlamshamin: i believe it should be the same as y and x, where counts_per_mm is 1_000

pylabrobot/plate_reading/tecan/infinite_backend.py
hazlamshamin: Currently, this is the fallback when there is any glitch with the usb. `_device_initialized=True` and no re-initialisation is being done if the flag remains `True`. However, I did notice, like sometimes, after leaving the machine for few hours, when I tried to read the plate, the computer fails to connect/control the plate reader (even after I have included this fallback). Still unsure what is the reason for this, that I had to physically power cycle and disconnect-reconnect the usb cable for it to work. Do you have any idea why this happen or what's the solution for this?

pylabrobot/plate_reading/tecan/infinite_backend.py
hazlamshamin: I overlooked this. Thanks to your review, I rechecked the DLL and have verified that firmware uses microseconds in all of its time unit. Was confused before during RE without the DLL, so this got mixed. Have pushed a new commit to fix this critical bug.

hazlamshamin:
> could you please explain what `_MODE_CAPABILITY_COMMANDS` are?
> 
> are they just more parameters like this?
> 
> ```
>       f"BEAM DIAMETER={self._capability_numeric('ABS', '#BEAM DIAMETER', 700)}",
> ```
> 
> that we can add as commands? why not add them?

I'm unable to directly comment to this so I'm quote replying.

`_MODE_CAPABILITY_COMMANDS` are meant to fetch the capability parameters supported by the machines (in case it is different for other models), as UI does this during initialisation. And yes, they are similar to the line that you quote. The idea was to query those parameters, then save them to be used for downstream commands. However, sometimes, I got errors during querying the full capability commands (the reason why I commented them now and leaving them there as reference). Therefore, I just did the very necessary one now, which is the beam diameter, with safe fallback of the default for the machine.

pylabrobot/plate_reading/tecan/infinite_backend.py
rickwierenga: thanks!

pylabrobot/plate_reading/tecan/infinite_backend.py
rickwierenga: chat suggests

  1. USB Selective Suspend - Windows/Linux power management suspends idle USB devices. Can be disabled in OS settings or  
  via libusb hints.                                                                                                       
  2. Device-side timeout - The plate reader itself may go into standby after idle time. The current _recover_transport    
  only triggers on read timeout, but the connection may appear "alive" while the device is unresponsive.

rickwierenga: so `_device_initialized = True` after hours of idle doesn't mean the device is actually responsive. I am removing the test because it would involve sending a command to the machine as a check, but we are already sending a command to the machine anyway

rickwierenga: https://github.com/PyLabRobot/pylabrobot/pull/797/commits/738ca07a8b964d536ddcfafba6e2c99fa3a538c7

rickwierenga:
thanks!

PR #799, opened 2025-12-22
Add `Resource.check_can_drop_resource_here()`

BioCam:


pylabrobot/resources/resource.py
```diff
@@ -837,3 +837,32 @@ def get_highest_known_point(self) -> float:
...
+    # Tree-wide invariants enforced at the root (e.g., global naming constraints).
+    self.get_root()._check_naming_conflicts(resource=resource)
+
+    # Extra drop-specific safety: prevent cycles
+    if self.is_in_subtree_of(resource):
+      raise ValueError(f"Cannot drop '{resource.name}' onto '{self.name}': would create a cycle.")
+
+    # Subclasses can add stricter “drop rules” here.
+    # Examples:
+    # - Only allow certain resource categories/types as children
```
rickwierenga: these should be done in `_check_assignment` (or even `assign_child_resource`, which is worse) not in `check_can_drop_resource_here` since this function is specific for dropping resources with arms.

pylabrobot/resources/resource.py
```diff
@@ -837,3 +837,32 @@ def get_highest_known_point(self) -> float:
...
+    # (delegated to `_check_assignment` to stay consistent as rules evolve).
+    self._check_assignment(resource=resource, reassign=reassign)
+
+    # Tree-wide invariants enforced at the root (e.g., global naming constraints).
+    self.get_root()._check_naming_conflicts(resource=resource)
+
+    # Extra drop-specific safety: prevent cycles
+    if self.is_in_subtree_of(resource):
+      raise ValueError(f"Cannot drop '{resource.name}' onto '{self.name}': would create a cycle.")
+
```
rickwierenga: this is a new check, so I would say it does not belong in this PR

also it should not be done in `check_can_drop_resource_here` but rather `_check_assignment`

pylabrobot/resources/resource.py
rickwierenga: so the point of this function is given the current state, is it possible to drop something here?

assignment is just a memory update, and is a superset of these constraints. drop constraints are that plus more

for example you can assign a plate while a door is closed but not drop something there

BioCam: nice - I like that; implemented

pylabrobot/resources/resource.py
BioCam: makes sense; implemented

rickwierenga:
nice, thanks for making a separate PR for this

rickwierenga:
when you get the chance, a new PR for the cyclic assignment would be sweet.  i hadn't previously considered this

PR #804, opened 2025-12-29
Tip-aware cLLD probing `start_pos_search` and safety validation

BioCam:


BioCam:
Thank you for catching this @rickwierenga !

I have excised the `C0` dependency out of `_move_z_drive_to_liquid_surface_using_clld` (and renamed it as per your suggestion), and moved the correction for the tip bottom (which was broken for both `start_search_pos` and `lowest_immers_pos`) into `clld_probe_z_height_using_channel` instead.

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -9046,31 +9083,81 @@ async def clld_probe_z_height_using_channel(
...
+    # Correct for tip length + fitting depth
+    tip_len = await self.request_tip_len_on_channel(channel_idx)
+
+    fitting_depth = 8  # mm, for 10, 50, 300, 1000 ul Hamilton tips
+    safe_head_top_z_pos = 334.7
+    safe_tip_bottom_z_pos = 99.98
+    safet_tip_top_z_pos = safe_head_top_z_pos - tip_len + fitting_depth
+
+    if start_pos_search is None:
+      start_pos_search = safe_head_top_z_pos - tip_len + fitting_depth
```
rickwierenga: why is this implicit behaviour good?

BioCam: Can you please elaborate what you mean?

BioCam:
Thank you @rickwierenga for the thorough review - much stronger now 🚀

PR #807, opened 2026-01-02
Error message when connecting to ambiguous FTDI device

BioCam:


rickwierenga:
tested and verified on vspin

pylabrobot/io/ftdi.py
```diff
@@ -28,61 +38,168 @@ def __init__(self, device_id: str, action: str, data: str):
...
+        f"VID:PID {vid_string}:{pid_string}, "
+        f"device_id {self._device_id}. "
+      )
+
+    if len(candidates) > 1:
+      raise RuntimeError(
+        f"Multiple FTDI devices found with specified criteria: "
+        f"VID:PID {vid_string}:{pid_string}, "
+        f"device_id {self._device_id}. "
+        f"Please specify the device_id parameter explicitly with the serial number of the desired device."
```
BioCam, line 157: This should show the candidates that were found to give actionable error information to the user

pylabrobot/io/ftdi.py
```diff
@@ -28,61 +38,168 @@ def __init__(self, device_id: str, action: str, data: str):
...
+    """Initialize the FTDI device connection with device resolution."""
+    if self._dev is not None and not self._dev.closed:
+      self._dev.close()
     try:
-      self.dev.open()
+      # Resolve which device to connect to
+      self._device_id = self._resolve_device_serial()
+
+      # Create and open device
+      self._dev = Device(lazy_open=True, device_id=self.device_id)
```
BioCam: The code section that adds the vid pid to search for has been completely removed in this version.
It could be added to the USB_PID_LIST/USB_VID_LIST directly, or given here as attributes to let pylabftdi handle the addition.
I would prefer explicit handling in pylabrobot for clarity

pylabrobot/io/ftdi.py
rickwierenga: all candidates will have the same information (pid, vid, serial)

pylabrobot/io/ftdi.py
rickwierenga: good catch, fixed in https://github.com/PyLabRobot/pylabrobot/pull/807/commits/30579635dcb5dda0bceeaa7449179cdab2de6f89

the list thing was always kind of a hack, you are not supposed to edit the internal constants of a library. I just didnt know better when adding that first in the clariostar backend. the Device parameter is the proper way to do it.

PR #812, opened 2026-01-07
add acceleration parameters `STAR` 96-head Y & Z movement; firmware-specific handling; park & move to safe; query 96 head config

BioCam:


rickwierenga:
very interesting stuff! I marked it as a draft, please mark as ready when ready

rickwierenga:
what do you think about naming all these methods with a `head96_` eg `park_96head` -> `head96_park`?

this way they will all show up nicely in autocomplete when typing `.head96_`

rickwierenga:
> what do you think about naming all these methods with a head96_ eg park_96head -> head96_park?

I am doing this

rickwierenga:
thanks @BioCam !

PR #815, opened 2026-01-09
Add NEST 2.2mL U-bottom deep well plate resource

jemancuso:


pylabrobot/resources/nest/plates.py
```diff
@@ -172,3 +173,34 @@ def nest_12_troughplate_15000uL_Vb(name: str) -> Plate:
...
+    size_z=41.85,  # from spec
+    lid=None,
+    model=NEST_96_wellplate_2200uL_Ub.__name__,
+    ordered_items=create_ordered_items_2d(
+      Well,
+      size_x=8.0,  # from spec (inner well width)
+      size_y=8.0,  # from spec (inner well length)
+      size_z=39.30,  # from spec
+      dx=9,  # measured
+      dy=7.5,  # measured
```
rickwierenga: it is actually possible to calculate dx from the drawing:

the space between the plate edge and center of wells in the first column is given as 14.05. `dx` means to left of the well, not center, so subtract half the plate width. 14.05 - 8 / 2 = 10.05.

If dx=9 works better for you, that's fair, sometimes the drawings are not accurate, but worth checking.

Same for dy.

pylabrobot/resources/nest/plates.py
```diff
@@ -172,3 +173,34 @@ def nest_12_troughplate_15000uL_Vb(name: str) -> Plate:
...
+      dz=2.3,  # measured
```
rickwierenga: this actually appears to be given as 2.55

rickwierenga:
thank you for the update @jemancuso !

PR #816, opened 2026-01-09
consistent liquid_surface_no_lld kwarg between aspirate and dispense on star

ben-ray:
liquid_surface_no_lld determines the starting position of an aspirate or dispense on a STAR when not using LLD

This PR will break implementations that fixed the string inconsistency using `liquid_surfaces_no_lld` as an aspirate kwarg and `liquid_surface_no_lld` for dispense on the STAR backend.

Ultimately we should decide on `liquid_surfaces_no_lld` or `liquid_surface_no_lld`, and use the same string for both aspiration and dispense. I prefer `liquid_surface_no_lld` because this is an exact string match to the firmware documentation.


rickwierenga:
thanks!

PR #841, opened 2026-01-26
add inheco odtc backend

ben-ray:
vibe-coded in 3 hours with gemini, tested against our 96w inheco ODTC and works

will need to be expanded to different model ODTCs (like 384)

The ODTC user flow is quite different from other thermal cyclers and as a result this backend is a little odd.

Before starting any PCR, you must first run a PreMethod for 7-10 minutes until the device is homogenously at temperature.
After PreMethod, you upload an xml protocol to the device (which is saved under a unique name until you delete it), which must have the same starting temp of your PreMethod.

If, by default, we decide to erase all methods in memory on the ODTC every time we write a new method, we could immediately upload and start a method containing a ramp from the current temp to the next temp you want to be, enabling set_block_temp and set_lid_temp to not be PreMethods and take <10 minutes

### DIFF
--- a/docs/user_guide/01_material-handling/thermocycling/inheco-odtc.ipynb
+++ b/docs/user_guide/01_material-handling/thermocycling/inheco-odtc.ipynb
@@ -39,7 +39,21 @@
    "execution_count": null,
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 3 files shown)


rickwierenga:
thanks for contributing this!

just confirmed basic methods work on our 384 well odtc

our 96 well odtc has some kind of problem, even the inheco software cannot find the ip. I will try again tomorrow

running a method gives

```
RuntimeError: command ExecuteMethod failed: 11 Invalid command parameter.
Method variant does not match the connected hardware variant: 3840000
Parameter name: methodName
```

not entirely unexpected

rickwierenga:
added an example notebook, start of docs

rickwierenga:
thanks!

rickwierenga:
merged as "experimental" given #894 might make breaking changes compared to your version. I am working on merging that right now

ben-ray:
#894 is much more complete, thx

PR #842, opened 2026-01-26
Expose Granular Channel Control: Dispensing Drive Request, Move, Empty

BioCam:
Screenshot 2026-01-26 at 15 37 22

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -1970,7 +1970,9 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
     if not (0 <= channel_idx < self.num_channels):
       raise ValueError(f"channel_idx must be between 0 and {self.num_channels-1}")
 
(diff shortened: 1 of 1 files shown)


rickwierenga:
what does "optionally returning to 0." mean?

rickwierenga:
answer for you: return to 0 is needed because empty is actually below 0

BioCam:
> what does "optionally returning to 0." mean?

Empty tips moves the piston to volume position -45 by default. It then stays there.

This optional parameter moves the plunger back to volume position 0 straight after. That seems like what people intuitively expect.

But there are situations when that is not desirable:
When there is still some liquid stuck to the inside of the tip, or if emtpy_tips happened while the tip bottom bottom has been immersed into a other liquid... In these scenarios any movement of the piston up would be undesirable.
Instead users might want to leave the piston at the lowest position, move out of the liquid, discard tips and let the piston automatically reset to 0 by itself - which happens during tip pickup I believe.

BioCam:
Volume position -45 was empirically determined to be the lowest consistent position achievable.

For reasons unknown, I could get down to -52 but it would sometimes require multiple commands to get there and would sometimes start erroring at -50.

rickwierenga:
works perfectly, thanks!

PR #843, opened 2026-01-27
STARBackend 96-Head Granular Control Update

BioCam:
- after initialisation 96-head dispensing drive is 215.92
- after tip pickup -> ALWAYS: 218.19
- after drop tip -> ALWAYS MOVES 215.9 UP !
- empty=True does move the 96-head piston all the way to vol=0 !
- IF slow 96-head dispensing drive flow_rate (~10 uL/sec) is chosen, vol=-25 can be reached

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -123,6 +123,23 @@ def need_iswap_parked(
   return wrapper
 
 
+def _requires_head96(
+  method: Callable[Concatenate["STARBackend", _P], Coroutine[Any, Any, _R]],
+) -> Callable[Concatenate["STARBackend", _P], Coroutine[Any, Any, _R]]:
+  """Ensure that a 96-head is installed before running the command."""
+
+  @functools.wraps(method)
+  async def wrapper(self: "STARBackend", *args, **kwargs):
+    if not self.core96_head_installed:
+      raise RuntimeError(
+        "This command requires a 96-head, but none is installed. "
+        "Check your instrument configuration."
+      )
+    return await method(self, *args, **kwargs)
+
+  return wrapper
+
+
 def parse_star_fw_string(resp: str, fmt: str = "") -> dict:
   """Parse a machine command or response string according to a format string.
 
@@ -2882,6 +2899,7 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
 
     return ret
 
+  @_requires_head96
   async def pick_up_tips96(
     self,
     pickup: PickupTipRack,
@@ -2922,8 +2940,6 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
 
     if tip_pickup_method not in {"from_rack", "from_waste", "full_blowout"}:
       raise ValueError(f"Invalid tip_pickup_method: '{tip_pickup_method}'.")
-
-    assert self.core96_head_installed, "96 head must be installed"
 
     prototypical_tip = next((tip for tip in pickup.tips if tip is not None), None)
     if prototypical_tip is None:
@@ -2987,6 +3003,7 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
         raise plr_e from e
       raise e
 
+  @_requires_head96
   async def drop_tips96(
     self,
     drop: DropTipRack,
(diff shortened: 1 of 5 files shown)


pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -7567,6 +7643,58 @@ async def head96_dispensing_drive_move_to_position(
...
 
+  async def head96_dispensing_drive_move_to_volume_position(
+    self,
+    vol: float,
+    flow_rate: float = 200.0,  # uL/sec
+    acceleration: float = 15_000.0,  # uL/sec**2,
+    current_limit: int = 7,
+  ):
+    """Move 96-head's dispensing drive to specified volume position
+
```
rickwierenga: how is this different from `head96_dispensing_drive_move_to_position`?

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -6910,6 +6957,32 @@ async def discard_tips_core96(
...
+        "Broken firmware command: raises STARFirmwareError: "
+        "{'CoRe 96 Head': UnknownHamiltonError('Position out of permitted area')}, H0DLid0080er54"
+    )
+    # resp = await self.send_command(
+    #   module="H0",
+    #   command="DL",
+    # )
+
+ 
+
```
rickwierenga, line 7059: instead of always raising, I think we might log the warning and then still try to send the command

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: The new method can move the dispense drive in the negative direction - it does make sense to deprecate the old one?

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: Are you sure?

If we know it is broken, is it not dangerous to have?

I wanted to have it here actually as a warning and to show that it has been considered, tested and found not to be safe, which avoids others from wasting time on it.

I have implemented it now but am not sure about it.

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
rickwierenga: why? both are in uL

DQ is absolute, DS is relative. for DS you query the current position to turn into an absolute movement. I don't see the point in that when we already have DQ

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
rickwierenga: yes, because it's not "dangerous broken". it just give an error before moving anything

there is a chance it might work for someone else

there is of course also a small chance that it will do something dangerous, but who knows? that's the case for every command

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
rickwierenga: i can confirm "head96_dispensing_drive_move_to_volume_position" (meaning relative firmware movement command, DS) moves down further than `head96_dispensing_drive_move_to_position` (meaning absolute firmware command, DQ)

however when you go down too far you get `STARFirmwareError: {'CoRe 96 Head': UnknownHamiltonError('Dispensing drive movement error')}, H0DSid0152er52`.

DQ will give `STARFirmwareError: {'CoRe 96 Head': UnknownHamiltonError('Position out of permitted area')}, H0DQid0146er54`

I think what is happening is that DS bypasses a position check and will just move the drive, possibly resulting in a drive position error

since i see no obvious reason for wanting to go outside the firmware allowed range, and it does carry risk since we bypass the restriction built into DQ.

for this reason i am removing this firmware command, DQ does what we need to do and is safer

if we ever needed to go outside the range (there is a small part that is not allowed by DQ but does not range a firmware movement error), we can have `head96_dispensing_drive_move_to_position_unsafe`

rickwierenga:
thanks!

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -6600,14 +6600,14 @@ async def request_core_96_head_initialization_status(self) -> bool:
     response = await self.send_command(module="H0", command="QW", fmt="qw#")
     return bool(response.get("qw", 0) == 1)  # type?
 
-  async def head96_dispensing_drive_and_squeezer_driver_initialise(
+  async def head96_dispensing_drive_and_squeezer_driver_initialize(
```
BioCam, line 6603: 😂 

makes sense - this is code, and code uses USian spelling :)

PR #846, opened 2026-01-28
Liconic backend (Liconic STX series automated incubator support)

sam-adaptyv:
Added a new backend for Liconic STX series of automated incubators. It is compatible with existing incubator backend commands also added a number of new incubator backend commands. Also added a new pice of hardware barcode_scanner which can be optionally initialized with a Liconic incubator to read barcodes with optional read_barcode variables added to current and added incubator commands. Right now the barcode is just printed to terminal as the incubator backed typically only returns the Plate object or nothing. If there was a barcode field added to the Plate resource that seems to be obvious place to return the scanned barcode but that could be done in the future.  For now the Keyence BL-1300 barcode scanner is the only model support but other models could be easily integrated. Documentation in user_guide > 01_material-handling > liconic.ipynb is also provided.

### DIFF
--- a/pylabrobot/storage/liconic/liconic_backend.py
+++ b/pylabrobot/storage/liconic/liconic_backend.py
@@ -35,7 +35,7 @@ LICONIC_SITE_HEIGHT_TO_STEPS = {
 }
 
 
-class LiconicBackend(IncubatorBackend):
+class ExperimentalLiconicBackend(IncubatorBackend):
   """Backend for Liconic incubators.
 
   Optionally accepts a BarcodeScanner instance for internal barcode reading.
--- a/docs/user_guide/01_material-handling/storage/liconic.ipynb
+++ b/docs/user_guide/01_material-handling/storage/liconic.ipynb
@@ -29,7 +29,7 @@
    "source": [
     "from pylabrobot.barcode_scanners import BarcodeScanner, KeyenceBarcodeScannerBackend\n",
     "from pylabrobot.resources.coordinate import Coordinate\n",
-    "from pylabrobot.storage import LiconicBackend\n",
+    "from pylabrobot.storage import ExperimentalLiconicBackend\n",
     "from pylabrobot.storage.incubator import Incubator\n",
     "from pylabrobot.storage.liconic.racks import liconic_rack_17mm_22, liconic_rack_44mm_10\n",
     "\n",
@@ -37,7 +37,7 @@
     "barcode_scanner_backend = KeyenceBarcodeScannerBackend(port=\"COM4\")\n",
     "barcode_scanner = BarcodeScanner(backend=barcode_scanner_backend)\n",
     "\n",
-    "liconic_backend = LiconicBackend(port=\"COM3\", model=\"STX220_HC\", barcode_scanner=barcode_scanner)\n",
+    "liconic_backend = ExperimentalLiconicBackend(port=\"COM3\", model=\"STX220_HC\", barcode_scanner=barcode_scanner)\n",
     "\n",
     "rack = [\n",
     "  liconic_rack_44mm_10(\"cassette_0\"),\n",
--- a/pylabrobot/storage/__init__.py
+++ b/pylabrobot/storage/__init__.py
@@ -3,4 +3,4 @@ from .chatterbox import IncubatorChatterboxBackend
 from .cytomat import CytomatBackend
 from .incubator import Incubator
 from .inheco.scila import SCILABackend
-from .liconic import LiconicBackend
+from .liconic import ExperimentalLiconicBackend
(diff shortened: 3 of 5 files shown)


pylabrobot/storage/liconic/racks.py
rickwierenga, line 1: to clarify, are these in mm or steps? Ideally all resources in PLR are in mm so they can be used in a hardware agnostic setting and make it easy for people to use/make their own labware.

pylabrobot/storage/liconic/liconic_backend.py
```diff
@@ -0,0 +1,559 @@
...
+    return barcode
+
+  def serialize(self) -> dict:
+    return {
+      **super().serialize(),
+      "port": self.io_plc.port,
+    }
+
+  @classmethod
+  def deserialize(cls, data: dict):
```
rickwierenga, line 578: `scan_barcode()` calls `ST 1910` to move the shovel to the barcode reading position but:

1. Doesn't call `_wait_ready()` before scanning — is the shovel guaranteed to have finished moving?
2. Doesn't call `RS 1910` afterward to return the shovel, unlike `read_barcode_inline()` which does both.

Is this intentional (e.g. caller is expected to handle reset), or should this match the `read_barcode_inline` pattern?

sam-adaptyv: 1. No the shovel isn't guaranteed to have finished moving. Often the barcodes can be read even while the shovel is moving so hasn't been found to be an issue. Only downside is it could slow down the speed of barcode scanning. 
2. Should call "RS 1910" and thus should match the read_barcode_inline pattern will ensure subsequent commands don't cause errors.

pylabrobot/storage/liconic/liconic_backend.py
```diff
@@ -0,0 +1,559 @@
...
+  17: 788,  # pitch=23, site_height=17
+  22: 959,  # pitch=28, site_height=22
+  23: 994,  # pitch=29, site_height=23
+  24: 1028,  # pitch=30, site_height=24
+  27: 1131,  # pitch=33, site_height=27
+  44: 1713,  # pitch=50, site_height=44
+  53: 2021,  # pitch=59, site_height=53
+  66: 2467,  # pitch=72, site_height=66
+  104: 3563,  # pitch=110, site_height=104
+}
```
rickwierenga, line 35: The 104mm rack (pitch 110mm) claims motor steps = 3563, but the formula `round(pitch * 1713 / 50)` that holds for all other sizes gives 3769. That is a 5.5% difference — too large for rounding.

All other sizes match the formula exactly:
```
pitch  claimed  computed
   11      377       377  OK
   17      582       582  OK
   ...
   72     2467      2467  OK
  110     3563      3769  FAIL
```

Is 3563 empirically determined from the hardware, or a data entry error? Added a test that documents this discrepancy.

pylabrobot/storage/liconic/liconic_backend.py
rickwierenga: thanks, I'll implement that!

pylabrobot/storage/liconic/racks.py
sam-adaptyv: So pitch and site_height are for sure in mm. The step variable is the number of steps the motors turns (i.e. steps of the motor) which moves the plate shuttle in z. So its specific to the motors used by Liconic. The conversion of pitch to steps is: step = pitch * ( 1713 / 50 ) rounded to the nearest integer. The steps value is what needs to sent directly to the memory in the liconic to move correctly ("WR DM23 x"). So we could calculate the steps value by the pitch value (which is mm) to meet this ideal.

pylabrobot/storage/liconic/liconic_backend.py
sam-adaptyv: It is a data entry error (aka typo) 104mm was used to compute 3563 but that is the plate height in mm when it should be calculated using the pitch value in mm which is 110 which would equals 3769 steps.

sam-adaptyv: Should be 110 for pitch and  3769 for steps

rickwierenga:
undid the changes to the front end and backend in anticipation for the coming refactor:

https://discuss.pylabrobot.org/t/updating-plr-api-for-machine-interfaces-discussion/445/

we'll make everything nice in that change

pylabrobot/storage/liconic/racks.py
rickwierenga: makes sense, and done on backend exactly where this stuff belongs. thanks!

rickwierenga:
labeled it `ExperimentalLiconicBackend` because we might make changes to the API in the upcoming refactor and this is to indicate the API might change. But we've actually been using the backend in some production protocols, works nicely!

rickwierenga:
thanks a lot for adding this!

sam-adaptyv:
of course Rick happy to help. Looking forward to contributing more!

PR #847, opened 2026-01-28
Enable `.probe_liquid_height()` to Detect Absence of Liquid

BioCam:
### Problem

Currently `STARBackend.probe_liquid_heights()` will error out if no liquid is detected instead of correctly detecting the absence of liquid (i.e. liquid_height=0.0 mm).
Consequently, `STARBackend.probe_liquid_volumes()` is also broken, and dispensing with cLLD into volume-non-deterministic destination containers is a hazard that can break aPs.

### Solution / This PR

`STARBackend.probe_liquid_heights()` already calculates the lowest immersion position to be the well cavity bottom for safety. If a container is empty this PR introduces the catching of the specific firmware error and converts it into liquid_height=0.0mm.
At the same time, a warning is logged for these containers since it is possible to have edge cases (e.g. if cLLD: low-volume liquids don't have enough dissipation volume to be detected; ultra-pure liquids are non-conductive)

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -1910,44 +1910,30 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
       await self.move_all_channels_in_z_safety()
       raise
 
-    # Compute average heights per channel and convert to relative to well bottom
-    absolute_liquid_heights = []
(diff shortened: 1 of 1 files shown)


docs/cookbook/star_semiautomated_plate_definition.ipynb
rickwierenga, line 1: not for this PR

rickwierenga:
works perfectly now, thanks for fixing! returning 0 is the correct behavior in this abstraction

PR #859, opened 2026-01-30
Make `STARBackend.aspirate96()`/`.dispense96()` from Quadrants Wellplate Type Adaptive

BioCam:
### The Problem

Aspirating and dispensing with the 96-head into different 384-wellplate quadrants on a STAR is currently broken:

The `STARBackend` hardcodes the "A1" well for non-rated plates to **all** wellplates (and "H12" to all 180degree rotated plates").

As a result, the backend generates the same firmware command for all 384-wellplate quadrants:

Simple distribute example with the 96-head:
Screenshot 2026-01-30 at 18 27 33

Note the identical `xs` and `yh` firmware parameter showing the coordinates are always the same.

### The Solution / This PR

Claude and I simply changed the reference point from hardcoding to the `aspiration.wells`/`dispense.wells` first instance for non-rotated plates (and last instance for 180degree rotated plates).

This will now be able to dynamically identify the correct coordinate no matter whether a 96-wellplate, 384-wellplate and (not-tested) 1536-wellplate.

Screenshot 2026-01-30 at 18 24 53

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_tests.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_tests.py
@@ -2,7 +2,7 @@
 
 import unittest
 import unittest.mock
-from typing import cast
+from typing import Literal, cast
 
 from pylabrobot.liquid_handling import LiquidHandler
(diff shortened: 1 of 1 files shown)


rickwierenga:
thanks!

Issue #822, opened 2026-01-18
Config for min channel spacing that's not 9 mm


HaamsRee:
I have a Hamilton STAR with a weird config. It has a single arm holding a 96-head and four 1 mL channels with 18 mm spacing. When running with probe_liquid_height=True in a 96-well format, this results in channel 3 probing in a position that's outside of the column of wells, channel 2 probing B1, channel 1 probing D1, etc. I found a temporary workaround by changing the probe to be sequential:

```python
    current_absolute_liquid_heights: List[float] = []
    for channel, container, tip, offset in zip(use_channels, containers, tips, resource_offsets):
      y_pos = container.get_location_wrt(self.deck, x="c", y="c", z="b").y + offset.y

      if sequential_probe:
        # # Give the active channel full freedom by parking others, then position only this channel.
        # await self.position_max_free_y_for_n(pipetting_channel_index=channel)
        # await self.move_channel_y(channel=channel, y=y_pos)
        await self.position_channels_in_y_direction({channel: y_pos})

      else:
        # Positions are sufficiently spaced; move the set in one go.
        await self.position_channels_in_y_direction(
          {channel: y for channel, y in zip(use_channels, target_y_positions)}
        )

      await self.move_z_drive_to_liquid_surface_using_clld(
        channel_idx=channel,
        lowest_immers_pos=container.get_absolute_location("c", "c", "cavity_bottom").z
        + tip.total_tip_length
        - tip.fitting_depth,
        start_pos_search=container.get_absolute_location("c", "c", "t").z
        + tip.total_tip_length
        - tip.fitting_depth
        + 5,
      )

      # TODO: Move z up after probing


      liquid_levels: List[int] = (await self.request_pip_height_last_lld())["lh"]  # type: ignore
      current_absolute_liquid_heights.append(float(liquid_levels[channel] / 10))
```

But I think there are other places where the 9 mm min spacing is hardcoded. It would be nice to have this as a config that the user can change, or read from the instrument.

rickwierenga:
thanks for bringing this up, I think you're the first person I hear from using 4mL channels

we certainly have 9mm hardcoded in some places ...

im certainly happy to work on this, just a heads up that there might be a few problems similar to this one

first of all: do you know if it's possible to mix&match 1mL and 4mL channels?

HaamsRee:
> thanks for bringing this up, I think you're the first person I hear from using 4mL channels
> 
> we certainly have 9mm hardcoded in some places ...
[...]

PR #862, opened 2026-01-31
Enable Variable Channel Spacing in `STARBackend`

BioCam:


pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -1470,6 +1471,9 @@ async def setup(
     async def set_up_pip():
       if (not initialized or any(tip_presences)) and not skip_pip:
         await self.initialize_pip()
+      self._channel_minimum_y_spacing = (
+        9.0  # TODO: identify from machine directly to override default
+      )
```
rickwierenga, line 1476: why not do that?

BioCam: Because the required method didn't exist until I created it today in https://github.com/PyLabRobot/pylabrobot/pull/870 (i.e. after this PR was created)

And because it would add multiple changes in the same PR (Enabling variable channel spacing | enabling detection of minimum possible y spacing - next one: merging the two into auto-detection of minimum y spacing during single-channel setup)

I've laid out the step by step plan here: https://github.com/PyLabRobot/pylabrobot/issues/822#issuecomment-3828390815

rickwierenga: sounds good

PR #866, opened 2026-02-02
add interface selection to FTDI

xbtu2:
- Interface Selection: Added an interface_select parameter to the FTDI constructor and passed it down to the underlying pylibftdi.Device during setup. This allows targeting specific interfaces on multi-port FTDI chips.
- Robust Device Logging: Improved the debug logging for connected devices. It now includes a try-except block when retrieving serial numbers to prevent ValueError crashes (common when a device is busy or permissions are restricted).
- Initialization: Updated the setup() method to utilize the new _interface_select member when instantiating the Device object.


rickwierenga:
thanks!

PR #871, opened 2026-02-03
Adding sequential pipetting

kelvin-ho-t-therapeutics:
Currently when aspirating from the same resource for all channels, the liquid handler will assume the user would like all pipettes to go into the resource well in one operation. However, when the resource well is too small, it raises a ValueError ("Resource is too small to space channels."). This causes a problem when doing assembly plates where one might aspirate from the same resource well (small 96-plate), but dispensing into different destination wells.

I used Chatgpt to write some code which detects whether it can go into the resource well in one operation and if not it goes into the sequential fallback which will process the list of aspirations as individual operations. I have tested this on a Hamilton STARlet with an internal protocol, but please test this using other wet lab protocols.


rickwierenga:
@BioCam is working on similar things :) @BioCam can you please handle this PR?

BioCam:
Sorry for the massive delay - and huge thank you to @kelvin-ho-t-therapeutics for this PR!

We have been working on a permanent solution to this and associated problems by creating a pipette positioning system in [PR#949 Encapsulate pipette batch scheduling into backend-agnostic module](https://github.com/PyLabRobot/pylabrobot/pull/949).

This has now been merged after a long development process.

We can now apply it to the problem of this PR and close this PR :)

PR #875, opened 2026-02-04
Resize `STAR(let)Deck.size_x`

BioCam:


pylabrobot/resources/deck.py
```diff
@@ -21,7 +21,7 @@ class Deck(Resource):
   def __init__(
     self,
     name: str = "deck",
-    size_x: float = 1360,
+    size_x: float = 1005,
```
rickwierenga: let's just remove the default values here altogether 😅

pylabrobot/resources/hamilton/hamilton_decks.py
```diff
@@ -501,15 +497,19 @@ def __init__(
...
+    # assign trash area, positioned 25mm to the right of the waste block
+    if with_trash:
+      if with_waste_block:
+        waste_block_x = self.get_resource("waste_block").get_location_wrt(self).x
+      else:
+        # Fallback: anchor to the rightmost rail when no waste block is present.
+        waste_block_x = self.rails_to_location(self.num_rails - 1).x
+      trash_x = waste_block_x + 25
+
+      self.assign_child_resource(
```
rickwierenga: if we have to put this stuff in this PR, then let's merge it with waste. so we have one flag. it is one object that contains both. we should not have two flags

pylabrobot/resources/deck.py
```diff
@@ -21,7 +21,7 @@ class Deck(Resource):
...
     size_y: float = 653.5,
     size_z: float = 900,
     origin: Coordinate = Coordinate(0, 0, 0),
```
rickwierenga, line 27:
```suggestion
    self,
    name: str = "deck",
    size_x: float,
    size_y: float,
    size_z: float,
    origin: Coordinate = Coordinate(0, 0, 0),
```

PR #881, opened 2026-02-07
correct core gripper resource position on hamilton star and starlet deck

rickwierenga:
@BioCam


rickwierenga:
mostly for visualizer, but also just for nice

rickwierenga:
image

rickwierenga:
not quite correct yet because there is some distance between the front gripper and the actual front of the resource

rickwierenga:
the way we made this work is by putting the known pickup locations in the center and moving the core gripper resource unit forward. information from ground truth venus

PR #886, opened 2026-02-12
Add request_tip_presence() for hardware-based error recovery

rickwierenga:
## Summary
- Adds `request_tip_presence() -> List[Optional[bool]]` to `LiquidHandlerBackend` for querying hardware tip sensors
- Implements it across all backends: STAR, Vantage, Nimbus (hardware query), chatterbox backends (tip tracker), and serializing backend
- Uses it in `LiquidHandler.pick_up_tips` and `drop_tips` error handling to determine per-channel success from hardware ground truth, falling back to `ChannelizedError` info when not available
- Catches `BaseException` (including `KeyboardInterrupt`) to ensure state tracking stays accurate

## Test plan
- [x] All 52 existing liquid handler tests pass
- [ ] Test with STAR hardware: verify `request_tip_presence` returns correct values after partial pick_up_tips failure
- [ ] Test with Vantage hardware: verify `query_tip_presence` wrapper works
- [ ] Test KeyboardInterrupt during pick_up_tips correctly queries tip presence and updates state

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -6398,24 +6398,14 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
     return float(z_pos_query["rd"] / 10)
 
   async def measure_tip_presence(self) -> List[Optional[bool]]:
-    """Request tip presence on each channel.
+    """Measure tip presence on all single channels using their sleeve sensors.
 
     Returns:
       A list of length `num_channels` where each element is `True` if a tip is mounted,
       `False` if not, or `None` if unknown.
     """
-    raw = await self.channels_sense_tip_presence()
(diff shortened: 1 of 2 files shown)


BioCam:
Is this a MEM-READ (`request_`) or a MEASUREMENT (`sense_`/`measure_`) command?

jrast:
I think it's on the border between MEM-READ and MEASUREMENT. The firmware monitors the presence sensor all the time, so the actual command does not reach out to the sensor to query in this exact instance, you will get the last state instead. I would go with a `request_` here. Just my 2¢

rickwierenga:
saw this a bit too late, renaming it back to request :)

BioCam:
Is this breaking all existing `STARBackend.channels_sense_tip_presence()`?

rickwierenga:
forgot to push the last commit, made a new PR for it: https://github.com/PyLabRobot/pylabrobot/pull/897/

PR #891, opened 2026-02-13
Fix focal_height docstring units to millimeters

hazlamshamin:


rickwierenga:
good catch thank you!

PR #892, opened 2026-02-13
Add bioER plate height volume calcs

har1eyk:
This PR adds volume<=>height calculations for the BioER deepwell plate. The relationship was fitted with a polynomial as it proved more accurate than a circular conical frustum. 

Volume (µL) | Height (mm)
-- | --
75 | 3.85
100 | 4.51
200 | 7.44
300 | 9.42
500 | 13.87
600 | 15.30
750 | 17.76
1000 | 22.08
1500 | 30.00
2000 | 38.07

BioER_plate_height_vs_volume

image


rickwierenga:
thanks!

PR #894, opened 2026-02-15
Inheco ODTC

cmoscy:
ODTC backend. Runs on-device protocols, and uses a scratch protocol name for uploading and executing individual commands like set mount temperature.


Stonelinks:
FWIW i used this branch successfully today on our ODTC

rickwierenga:
have you noticed the client side parallelism check being necessary? It adds some complexity to code. I am not opposed to including more client side checks when this is necessary for safe device operation, but this can also be a risk if we are more stringent than what the machines requires.

Stonelinks:
Honestly, we are using it in a pretty simplistic fashion right now (single thread wait for a protocol to finish executing), so no haven't noticed any issues with client parallelism. Also very new to pylabrobot in general.

cmoscy:
> have you noticed the client side parallelism check being necessary? It adds some complexity to code. I am not opposed to including more client side checks when this is necessary for safe device operation, but this can also be a risk if we are more stringent than what the machines requires.

1. Don't think not including the client side check would break anything, totally fine if we want ODTC to just let the ODTC handle it by returning the error code 4 instead. 

More as a source of documentation since the spec is disclosed here. But doesn't mean it has to be wired directly into execute_command


[TD_SILA_FWCommandSet (25210).pdf](https://github.com/user-attachments/files/26010799/TD_SILA_FWCommandSet.25210.pdf)

rickwierenga:
prefer to have them handle it if it works, less for us to maintain :) error messages should be equally descriptive

cmoscy:
> prefer to have them handle it if it works, less for us to maintain :) error messages should be equally descriptive

Sure! Feel free to pull it out, or would you like me to make the corresponding changes/test?

rickwierenga:
- also, removed the ODTCThermocycler class for now, in anticipation of the refactor we are working on (http://discuss.pylabrobot.org/t/updating-plr-api-for-machine-interfaces-discussion/445)
- moved your improvements of odtc sila into inheco sila for the shared stuff
- only use 96 and 384 for types rather than duplicating

rickwierenga:
> Sure! Feel free to pull it out, or would you like me to make the corresponding changes/test?

no worries I can take care of it :) we have both the 96 and 384 well odtc. thanks for offering :)

rickwierenga:
what is the difference between ODTCMethodSet vs ProtocolList?

cmoscy:
> ProtocolList

Level of effort. ProtocolList is just going to pull stored method names for reference (or running an existing protocol by name). Where ODTCMethodSet will parse them into their respective ODTCProtocol objects (for detailed information, using as a starting point to edit and send as a new protocol, etc).

rickwierenga:
should we trust ODTCConfig or ODTCProtocol defaults?

rickwierenga:
> Level of effort. ProtocolList is just going to pull stored method names for reference (or running an existing protocol by name)

ODTCMethodSet could easily provide the list of names, so I deleted ProtocolList

rickwierenga:
does `build_progress_from_data_event` typically get `"current_step_index"` etc. in the payload? what does this depend on? the conditional parsing smells suspicious

rickwierenga:
in general do you know how rigid the data stream is? many dicts say `.get` rather than just direct access. I think we should be able to trust the data / if there are differences explain them and handle explicitly

cmoscy:
> in general do you know how rigid the data stream is? many dicts say `.get` rather than just direct access. I think we should be able to trust the data / if there are differences explain them and handle explicitly

ODTC Config should be trusted defaults/where settings from a downloaded protocol are stored. 

> does build_progress_from_data_event typically get "current_step_index" etc. in the payload? what does this depend on? the conditional parsing smells suspicious

and 
> in general do you know how rigid the data stream is? many dicts say .get rather than just direct access. I think we should be able to trust the data / if there are differences explain them and handle explicitly

I agree, I overlooked this earlier but am not satisfied with how progress / parsing works while methods are running. I'm definitely missing something there. I'll take a look soon and see if I can come back with a satisfying answer, or a cleaner version/fix.

rickwierenga:
how about for the defaults we provide? the PLR interface will be `Protocol` through `run_protocol`, which does not have the same information as `ODTCProtocol`, which is why `ODTCConfig` exists. Right now both `ODTCConfig` and `ODTCProtocol` provide defaults . Like `post_heating` is True/False. What should we provide as the default? Defaults should only be set in one place, meaning `post_heating` will be a required parameter for either config or protocol. In terms of code I strongly prefer protocol. But not sure which _values_ you want those to be, since right now there are conflicts

Stonelinks:
Just want to say I'm following along here and awaiting this being merged into main. Thanks for your work!

PR #901, opened 2026-02-18
Replace PreciseFlexJointCoords with PFAxis IntEnum + dict

rickwierenga:
## Summary
- Replace `PreciseFlexJointCoords` dataclass with `PFAxis(IntEnum)` + `Dict[int, float]`, mirroring the KX2 pattern
- `move_to` now accepts partial dicts — unspecified axes are filled from current position (e.g. `await arm.move_to({PFAxis.RAIL: 300})`)
- Remove unused `JointCoords` type alias from `standard.py`

## Test plan
- [x] All 172 existing PreciseFlex backend tests pass


rickwierenga:
this is to unify it with the kx2 #880 

dicts are nicer because you can specify individual joints rather than having to specify them all at once.

PR #902, opened 2026-02-19
fix: add volume functions for Corning 96-well flat and V-bottom plates

margaretsantiago:
Summary
Adds proper geometry-based volume-from-height and height-from-volume functions for two Corning 96-well plates, using dimensions from Corning's MicroplateDimensions technical drawing.
Changes
height_volume_functions.py: New compute_volume_from_height_conical_frustum_in_well() and inverse. Interpolates radius at liquid height for tapered wells.
corning/plates.py: Added volume functions to Cor_96_wellplate_360ul_Fb using frustum geometry (top 6.86mm, bottom 6.35mm). New Cor_96_wellplate_320ul_Vb for standard V-bottom 96-well (cat. 3894/3896) with cone+cylinder geometry. Deprecation stub for old Cos_96_Vb name.
Verification
Flat-bottom max volume: 365.8 uL (Corning published: 360 uL, 1.6% from rounded nominal). V-bottom max volume: 319.97 uL (Corning published: 320 uL). All round-trips accurate to <0.01mm / <0.01uL.
Data sources: Corning MicroplateDimensions96-384-1536.pdf, Opentrons shared-data JSON, Corning e-catalog.


pylabrobot/resources/corning/plates.py
```diff
@@ -82,6 +102,82 @@ def Cos_96_EZWash(name: str, with_lid: bool = False) -> Plate:
...
+def Cor_96_wellplate_320ul_Vb_Lid(name: str) -> Lid:
+  raise NotImplementedError("This lid is not currently defined.")
+
+
+# Previous names in PLR:
+def Cos_96_Vb(name: str, with_lid: bool = False) -> Plate:
+  raise NotImplementedError(
+    "Deprecated. Use 'Cor_96_wellplate_320ul_Vb' instead."
+  )
+
```
rickwierenga: this is not something that has existed

rickwierenga:
did you verify these results on a physical machine? we find that the geometry based approach is not always accurate (due to surface tension and such)

pylabrobot/resources/corning/plates.py
margaretsantiago: You're right — removed. Thanks for the catch :)

margaretsantiago:
No physical verification yet — these are based on Corning's published dimensions and geometric modeling. The computed max volumes match Corning's published specs (365.8 µL vs 360 µL for flat-bottom, 319.97 µL vs 320 µL for V-bottom). I agree geometry alone won't capture surface tension effects. I plan to do empirical validation on Tecan EVO. If you have access to another lh, please let me know. I'd gladly validate those too. Would you prefer to hold this PR until I have bench data?

rickwierenga:
yes, we find that physically validating these height <> volume functions is important because the geometrical approach is unfortunately not super accurate.

rickwierenga:
happy to proceed with merging Cor_96_wellplate_320ul_Vb based on the technical drawing in the meantime if you want to do that

margaretsantiago:
Yes, please merge.

pylabrobot/resources/corning/plates.py
```diff
@@ -82,6 +102,75 @@ def Cos_96_EZWash(name: str, with_lid: bool = False) -> Plate:
...
+  """
+
+  # Corning MicroplateDimensions96-384-1536.pdf:
+  # top internal diameter 6.86mm, bottom internal diameter 6.37mm, well depth 11.12mm
+  # The V-bottom is modeled as a cone (apex at bottom) transitioning to a cylinder.
+  # Cone height back-calculated from published total volume (320uL) using d=6.37mm.
+  WELL_DIAMETER = 6.37  # diameter at cone-to-cylinder transition
+  CONE_HEIGHT = 1.62  # mm, back-calculated to match 320uL total volume
+  CYLINDER_HEIGHT = 11.12 - CONE_HEIGHT  # 9.50mm
+  WELL_DEPTH = 11.12
```
rickwierenga, line 130: the spec says 6.86 for the top diamater, let's use the widest part

where did you find 6.37?

how did you find the cone and cylinder height?

dx Is 14.38 - well_size/2

let's use the defined constants wherever possible

how did you find material_z_thickness?

https://www.corning.com/catalog/cls/documents/drawings/DWG00848.PDF

rickwierenga:
can you please also add a documentation entry?

pylabrobot/resources/corning/plates.py
margaretsantiago: Hi Rick,

Thank you for the thorough review. Going through each one:

1. "the spec says 6.86 for the top diameter, let's use the widest part” — that works. I’ll recalculate using 6.86 instead of 6.37. I’ve also created a separate file for my own reference to account for both top and bottom diameters based on manufacturer’s specs. This way I can continuously cross-reference various specs with bench data and physical tests, using different liquids. I plan to collect data on cone-height, cone-angles, height of V section. 

2. “where did you find 6.37?” From Corning’s website, [Corning Microplate Dimensions PDF](https://www.corning.com/catalog/cls/documents/drawings/MicroplateDimensions96-384-1536.pdf) lists top and bottom diameters across varying microplates. 

3. "dx Is 14.38 - well_size/2” — Sounds good. The use of 14.3 came directly from Corning’s website, [Corning Microplate Dimensions PDF](https://www.corning.com/catalog/cls/documents/drawings/MicroplateDimensions96-384-1536.pdf) 

4.  "how did you find the cone and cylinder height?” **Cone height** — was back calculated to match 320uL total volume. **Cylinder height** — I used Corning’s “well depth” measurement (see link above) and subtracted  cone height

5. "let's use the defined constants wherever possible” - Let’s do that. 

6. "how did you find material_z_thickness?” — I received data from a research student who measured a cross-section of a 96-well v-bottom plate that was going to be disposed. I do plan to take my own physical measurements as soon as the plates arrive. 

[CorningMicroplateDimensions.pdf](https://github.com/user-attachments/files/25657740/CorningMicroplateDimensions.pdf)

PR #904, opened 2026-02-21
Fix SpectraMax reliability in MolecularDevicesBackend

xbtu2:
This PR improves the reliability of the MolecularDevicesBackend for SpectraMax plate readers.

Key changes:
- Improved response decoding by using `errors="replace"`.
- Enhanced success verification by checking all response parts for "OK" instead of just the first one.
- Added defensive parsing to `get_status`, `read_error_log`, and `get_temperature` to handle malformed or short responses gracefully.


pylabrobot/plate_reading/molecular_devices/backend.py
```diff
@@ -326,19 +326,27 @@ async def close(self, plate: Optional[Plate] = None) -> None:
...
+    if len(res) > 1:
+      parts = res[1].split()
+    else:
+      parts = res[0].replace("OK", "").split()
+
+    if len(parts) >= 2:
+      return (float(parts[1]), float(parts[0]))  # current, set_point
+    elif len(parts) == 1:
+      return (float(parts[0]), float(parts[0]))
+    return (0.0, 0.0)
```
rickwierenga: why is `return (float(parts[0]), float(parts[0]))` valid?

`return (0.0, 0.0)` seems like an error case. I think the user should know when a command failed and decide what to do.

pylabrobot/plate_reading/molecular_devices/backend.py
```diff
@@ -326,19 +326,27 @@ async def close(self, plate: Optional[Plate] = None) -> None:
 
   async def get_status(self) -> List[str]:
     res = await self.send_command("!STATUS")
-    return res[1].split()
+    return res[1].split() if len(res) > 1 else res
 
   async def read_error_log(self) -> str:
     res = await self.send_command("!ERROR")
-    return res[1]
+    return res[1] if len(res) > 1 else res[0]
```
rickwierenga, line 338: this seems strange. have you observed this working correctly?

pylabrobot/plate_reading/molecular_devices/backend.py
xbtu2: you are right about this. someone suggested these fixes for me and i didn't look at them carefully. thanks!

rickwierenga:
sweet

PR #908, opened 2026-02-25
Expose `STARBackend.channel_measure_temperature()`

BioCam:


rickwierenga:
all report exactly 20.0C, so I don't think this works

BioCam:
> all report exactly 20.0C, so I don't think this works

ahh sad, our AC is set to 20C which is why I thought it was returning the correct values

If it doesn't return the correct values then the firmware command is not working and I will close the PR - I will test with different AC conditions and ask some more people to test in their labs before doing so

rickwierenga:
our labs can't be both exactly 20.0C on all channels 😅

BioCam:
yes, that is what I am saying

BioCam:
Closing this PR now due to the STAR not actually making these sensors usable - sad, would have been extremely useful to have Hamilton-built temperature sensors in every channel

PR #914, opened 2026-03-01
Add configurable liquid_color parameter to Visualizer 

BioCam:
260301_VIsualizer_liquid_color_parameter_default

- making liquid color configurable
- new default color

### Examples

```python
from pylabrobot.visualizer import Visualizer                                                                                           
                                                            
# Default (amber, CVD-friendly)                                                                                                        
vis = Visualizer(lh, open_browser=False)                  
                                                                                                                                       
# Custom color                                            
vis = Visualizer(lh, open_browser=False, liquid_color="5DADE2")

# With hash prefix (stripped automatically)
vis = Visualizer(lh, open_browser=False, liquid_color="#EF233C")

# Invalid → clear error
vis = Visualizer(lh, open_browser=False, liquid_color="red")
# ValueError: liquid_color must be a 6-character hex string (e.g. 'F39C12'), got 'red'
```


rickwierenga:
why not pick one good default and stick with that for now?

I think we should reserve liquid colors for the case where we want to track different liquid "types" in the future

BioCam:
Because people with different types of CVD cannot see different colours.

And the liquid is the only colour that depends on it's opacity.

I.e. giving the user power to make it what they need it to be :)

We could still make it track different liquid types in the future either way

pylabrobot/visualizer/visualizer_tests.py
```diff
@@ -60,6 +60,30 @@ def test_result_is_valid_json(self):
...
+  def test_invalid_hex_raises(self):
+    r = Resource(size_x=100, size_y=100, size_z=100, name="root")
+    with self.assertRaises(ValueError):
+      Visualizer(r, open_browser=False, liquid_color="nope")
+
+  def test_short_hex_raises(self):
+    r = Resource(size_x=100, size_y=100, size_z=100, name="root")
+    with self.assertRaises(ValueError):
+      Visualizer(r, open_browser=False, liquid_color="FFF")
+
```
rickwierenga, line 85: :(

rickwierenga:
ok yeah it's simple enough to keep and might help some people

PR #915, opened 2026-03-01
Support variable channel spacing via per-channel minimum Y spacing firmware retrieval

BioCam:
## The Problem

  All channel spacing in `STARBackend` was hardcoded to a single value (9mm / 8.9mm / 8990µm depending on the call site). This fails for:                           
                                                                                   
  - **4-channel single-rail machines** with 18mm spacing; reported PLR user configuration in #822
  - **Mixed 1mL + 5mL channel configurations** where adjacent channels have different physical constraints (e.g. 8.98mm between two 1mL channels, 17.96mm between a 1mL and 5mL channel) -> preparation for the new X1 channels arriving to consumers

mixed_channel_spacing

The constraint is **per-channel and symmetric**: each channel carries its own minimum spacing, and the safe distance between any two adjacent channels is the `max()` of both.

  ## PR Content/Solution

  **Infrastructure (`STAR_backend.py`)**
  - `_channel_minimum_y_spacing: float` → `_channels_minimum_y_spacing: List[float]` (per-channel)
  - `_min_spacing_between(i, j)` helper: `round(max(spacing[i], spacing[j]), 1)`
  - `channels_request_y_minimum_spacing()`: parallel `P<n>VY` queries via `asyncio.gather` at setup
  - All 11 consumer sites updated: `can_reach_position`, `position_channels_in_y_direction`, `get_channels_y_positions`,
  `probe_liquid_heights` safe bounds, CoRe gripper assert, validation

  **Simulation (`STAR_chatterbox.py`)**
  - `channels_minimum_y_spacing` constructor param (defaults to `[9.0] * num_channels`)
  - `channel_request_y_minimum_spacing()` / `channels_request_y_minimum_spacing()` overrides — return stored values without firmware calls

  **Generic LH (`utils.py`)** 
  - `min_spacing` parameter added to `get_wide_` and `get_tight_single_resource_liquid_op_offsets` (default unchanged)
  - Renamed constant to `GENERIC_LH_MIN_SPACING_BETWEEN_CHANNELS`

  **Tests (`STAR_tests.py`)**
  - 4 behavioral tests comparing 4ch/18mm vs 8ch/9mm: reachability bounds, validation rejection, and JY command payload differences

**NOTE: This PR is meant to bring NO BREAKING CHANGES** :)


pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -1552,10 +1558,16 @@ def setup_done(self) -> bool:
...
+  async def channel_request_y_minimum_spacing(
+    self, channel_idx: int, rounded: bool = True
+  ) -> float:
     """Request the minimum Y spacing for a given channel.
+
     Args:
       channel_idx: the channel index to query. (0-indexed)
+      rounded: If True (default), round the result to 1 decimal place. Hardware reports
+        exact values (e.g. 8.98) which round to the expected 9.0.
+
```
rickwierenga: just choose a behavior that works, a parameter is making it overly complex

I would lean towards not rounding and letting the user (caller) round if they need to

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -1570,20 +1582,44 @@ async def channel_request_y_minimum_spacing(self, channel_idx: int) -> float:
...
+    # frontmost channel can go to y=6, every channel behind it constrains its min Y
+    spacings = self._channels_minimum_y_spacing
+    min_y_pos = 6 + sum(spacings[i] for i in range(channel_idx + 1, self.num_channels))
     if position.y < min_y_pos:
       return False
 
-    # backmost channel can go to y=601.6, every channel before that is about 8.9 mm further forward
-    max_y_pos = 601.6 - 8.9 * channel_idx
+    # backmost channel can go to y=601.6, every channel in front constrains its max Y
+    max_y_pos = 601.6 - sum(spacings[i] for i in range(channel_idx))
```
rickwierenga: nit pick: `sum(spacings[:channel_idx])` is easier to read

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -4237,12 +4273,9 @@ async def core_check_resource_exists_at_location_center(
...
     center = location + resource.centers()[0] + offset
     y_width_to_gripper_bump = resource.get_absolute_size_y() - gripper_y_margin * 2
-    assert (
-      self._channel_minimum_y_spacing
-      <= y_width_to_gripper_bump
-      <= round(resource.get_absolute_size_y())
-    ), (
-      f"width between channels must be between {self._channel_minimum_y_spacing} and "
+    min_spacing = min(self._channels_minimum_y_spacing)
+    assert min_spacing <= y_width_to_gripper_bump <= round(resource.get_absolute_size_y()), (
```
rickwierenga: this is not strictly correct for frankenstars, we would need half of each channel's y spacing. it's fine with me since it's a narrow edge case, just flagging in case you want to go to that with this PR

rickwierenga: or _min_spacing_between!

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -1208,6 +1208,14 @@ def __init__(
...
     self._setup_done = False
 
+  def _min_spacing_between(self, i: int, j: int) -> float:
+    """Return the conservative minimum Y spacing required between channels *i* and *j*.
+
+    The constraint is the larger of the two channels' individual minimum spacings,
+    rounded up to 1 decimal place for safe movement.
+    """
+    return round(max(self._channels_minimum_y_spacing[i], self._channels_minimum_y_spacing[j]), 1)
+
```
rickwierenga, line 1366: conservative, yes

but why not half of each channel's spacing?

BioCam: What do you mean by this?

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -10115,26 +10150,32 @@ async def clld_probe_y_position_using_channel(
...
       if channel_idx == 0:  # safe default
         adjacent_y_pos = STARBackend.y_drive_increment_to_mm(13_714) + 9  # y-position=635 mm
       else:  #  previous channel
         adjacent_y_pos = await self.request_y_pos_channel_n(channel_idx - 1)
 
       max_safe_y_mov_dist_post_detection = (
-        adjacent_y_pos - detected_material_y_pos - self._channel_minimum_y_spacing
+        adjacent_y_pos
+        - detected_material_y_pos
+        - self._min_spacing_between(channel_idx, adjacent_idx)
```
rickwierenga: `adjacent_idx = max(channel_idx - 1, 0)` seems like a special case to me, not sure if we can just say `adjacent_idx = 0` when `channel_idx = 0` and expect the code to work.

strictly speaking `_min_spacing_between` should raise when this happens I think

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
rickwierenga: I thought the spacing returned by each channel is around that channel's center. so the space between two channels is not `2*radius_max` but rather `1/2*radius_n + 1/2*radius_m` where radius is their safe spacing

your implementation is indeed conservative as it says, but does it need to be?

pylabrobot/liquid_handling/backends/hamilton/STAR_tests.py
rickwierenga, line 1: nice

pylabrobot/liquid_handling/utils.py
```diff
@@ -21,6 +21,7 @@ def _get_centers_with_margin(dim_size: float, n: int, margin: float, min_spacing
 def get_wide_single_resource_liquid_op_offsets(
   resource: Resource,
   num_channels: int,
+  min_spacing: float = GENERIC_LH_MIN_SPACING_BETWEEN_CHANNELS,
 ) -> List[Coordinate]:
```
rickwierenga, line 26: can also be a list in the future but let's do this first and save that for a future PR ,..

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: I think you're thinking about center of channel i to periphery of channel i+1?

But it is always center to center distance, which is 9 mm for standard 1mL, and 18 mm for 1mL on single rail and 18 mm for 5 mL.

The firmware is overly precise because it uses steps. But when using the precise values the firmware is anyway rounding to 9 or 18 mm 😅

rickwierenga: oh you want to have max because the channels are mounted on the rail and not just in arbitrary space

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -4237,12 +4273,9 @@ async def core_check_resource_exists_at_location_center(
...
-      self._channel_minimum_y_spacing
-      <= y_width_to_gripper_bump
-      <= round(resource.get_absolute_size_y())
-    ), (
-      f"width between channels must be between {self._channel_minimum_y_spacing} and "
+    min_spacing = min(self._channels_minimum_y_spacing)
+    assert min_spacing <= y_width_to_gripper_bump <= round(resource.get_absolute_size_y()), (
+      f"width between channels must be between {min_spacing} and "
       f"{resource.get_absolute_size_y()} mm"
       " (i.e. the minimal distance between channels and the max y size of the resource"
```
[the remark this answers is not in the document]
rickwierenga: yes

pylabrobot/liquid_handling/utils.py
```diff
@@ -3,7 +3,7 @@
 from pylabrobot.resources.coordinate import Coordinate
 from pylabrobot.resources.resource import Resource
 
-MIN_SPACING_BETWEEN_CHANNELS = 9
+GENERIC_LH_MIN_SPACING_BETWEEN_CHANNELS = 9
```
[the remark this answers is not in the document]
rickwierenga: whatever, easy fix

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: the user doesn't have access to a lot of the usage of these parameters (e.g. going into the spread channels command at runtime), but the user does expect minimal spacing of 9 mm.

if forced to not give the user deeper insight into what the firmware returns via this parameter, I would prefer to use `math.ceil` which is conservative and safe, and always produces what the user expects -> 9 mm / 18 mm

rickwierenga: if that's what the method should method should return to be correct sure

otherwise the user can call round or ceil or whatever

rickwierenga:
we do not know what the minimum y position for 5mL channels in the front, but given that for 9mm channels it's 6, it'd probably be 15

BioCam:
> we do not know what the minimum y position for 5mL channels in the front, but given that for 9mm channels it's 6, it'd probably be 15

Why 15?

If `y = 6 mm` is the center of 1 mL channel index=-1 (`radius=4.5 mm`) then the frontfacing inaccessible y-distance is `1.5 mm`.

Since a 5 mL channel is assumed to be double the diameter (radius=9 mm): `9 mm + 1.5 = 10.5 mm`

rickwierenga:
merged main into this branch, including new changes from #876

PR #918, opened 2026-03-03
Addressable troughs and Star(let) waste block

cmoscy:
# STAR/STARlet waste as addressable resource, AddressableTrough, Trough_CAR_5R60_A00 fix

1. Waste spots as a first class resource from Nimbus Pattern

The STAR/STARlet in Venus waste has distinct physical positions; modeling them as first-class spots lets each channel use a defined location. Drop heights now use the per-spot resource location (with a +10 mm approach), matching Hamilton/Venus behavior.

2. Addressable Troughs for defined positional access 

When the handler assumed “one resource → spread channels,” troughs with internal dividers (e.g. Hamilton 60 mL) could be hit. We need explicit entry positions (e.g. A1–H1) so the robot only goes to valid spots. AddressableTrough provides those spots while sharing one volume pool. To support this, spread is only applied when the resource is explicitly multichannel-capable (`Trough`/`Trash`); trough spots are not, so they are used as explicit positions and no longer get auto-spread into dividers.

Maybe it's worth migrating to addressable versions of troughs in general, but will leave that up for discussion if this generates interest. 

3. Labware Definition Fix
`Trough_CAR_5R60_A00` site positions were wrong, missing 5th position; the definition is corrected so the fifth site and spacing match the hardware. 

Physically tested on Starlet in lab.

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -8,9 +8,10 @@ import sys
 import warnings
 from abc import ABCMeta
 from contextlib import asynccontextmanager, contextmanager
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import (
   Any,
+  Awaitable,
   Callable,
   Coroutine,
   Dict,
@@ -37,6 +38,7 @@ from pylabrobot.liquid_handling.backends.hamilton.base import (
   HamiltonLiquidHandler,
 )
 from pylabrobot.liquid_handling.backends.hamilton.common import fill_in_defaults
+from pylabrobot.liquid_handling.backends.hamilton.planning import group_by_x_batch_by_xy
 from pylabrobot.liquid_handling.errors import ChannelizedError
 from pylabrobot.liquid_handling.liquid_classes.hamilton import (
   HamiltonLiquidClass,
@@ -60,6 +62,7 @@ from pylabrobot.liquid_handling.standard import (
   SingleChannelDispense,
 )
 from pylabrobot.liquid_handling.utils import (
+  MIN_SPACING_EDGE,
   get_tight_single_resource_liquid_op_offsets,
   get_wide_single_resource_liquid_op_offsets,
 )
@@ -114,7 +117,7 @@ def need_iswap_parked(
 
   @functools.wraps(method)
   async def wrapper(self: "STARBackend", *args, **kwargs):
-    if self.iswap_installed and not self.iswap_parked:
+    if self.extended_conf.left_x_drive.iswap_installed and not self.iswap_parked:
       await self.park_iswap(
         minimum_traverse_height_at_beginning_of_a_command=int(self._iswap_traversal_height * 10)
       )
@@ -131,7 +134,7 @@ def _requires_head96(
 
   @functools.wraps(method)
   async def wrapper(self: "STARBackend", *args, **kwargs):
-    if not self.core96_head_installed:
+    if not self.extended_conf.left_x_drive.core_96_head_installed:
       raise RuntimeError(
         "This command requires a 96-head, but none is installed. "
         "Check your instrument configuration."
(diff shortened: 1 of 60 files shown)


pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -1739,15 +1739,27 @@ async def drop_tips(
...
     """
 
     self.ensure_can_reach_position(use_channels, ops, "drop_tips")
 
+    # New z formula (max_z+10, max_z) only for addressable waste spots; fallback to trash area
+    # uses the original magic numbers (59.9, 49.9).
+    all_trash = all(isinstance(op.resource, Trash) for op in ops)
+    is_waste_drop = all_trash and all(
+      getattr(op.resource, "category", None) == "waste_position" for op in ops
+    )
```
rickwierenga, line 1928: what is the difference between `is_waste_drop` and `all_trash`? we have historically used the trash object rather than category

pylabrobot/resources/hamilton/hamilton_decks.py
```diff
@@ -554,6 +592,51 @@ def get_trash_area96(self) -> Trash:
...
+      RuntimeError: If no waste tip spots were defined at deck initialisation.
+      ValueError: If ``n_channels`` exceeds the number of defined waste tip spots.
+    """
+    waste_block = self.get_resource("waste_block")
+
+    # Filter children by category, then sort by numeric suffix to guarantee order.
+    all_positions: List[Trash] = sorted(
+      [cast(Trash, child) for child in waste_block.children if child.category == "waste_position"],
+      key=lambda r: int(r.name.split("_")[-1]),
+    )
```
rickwierenga, line 623: rather than sorting by name, why not sort by `.location.y` directly?

pylabrobot/resources/hamilton/hamilton_decks.py
```diff
@@ -554,6 +592,51 @@ def get_trash_area96(self) -> Trash:
...
+      )
+
+    n_total = len(all_positions)
+    if n_channels > n_total:
+      raise ValueError(
+        f"Requested {n_channels} channels but only {n_total} waste tip spots are defined."
+      )
+
+    step = max(1, n_total // n_channels)
+    return [all_positions[i * step] for i in range(n_channels)]
```
rickwierenga, line 638: this will get arbitrary locations per channel, but I thought the point is to have one fixed position per channel. Otherwise it's easy to run into problems, eg discarding tips on channel 0 and 2 will leave 1 in the middle which this math does not account for.

also, given waste positions are already named `"waste_position_{I}"`, why not just fetch those?

this function could be as simple as

```python
return [self.get_resource(f"waste_position_{ch}") for ch in channels]
```

pylabrobot/liquid_handling/liquid_handler.py
```diff
@@ -767,6 +767,22 @@ async def discard_tips(
...
+        waste_spots = get_waste_positions(n)
+        return await self.drop_tips(
+          tip_spots=waste_spots,
+          use_channels=use_channels,
+          offsets=offsets,
+          allow_nonzero_volume=allow_nonzero_volume,
+          **backend_kwargs,
+        )
+      except (RuntimeError, ValueError):
+        pass  # No spots defined or n exceeds available spots; fall through to trash.
```
rickwierenga, line 784: I think it is worth making this a required method for all `Deck`s, similar to `get_trash_area96`. other decks can easily fall back onto their current trash, and LH will still handle the offsets for those (for the time being)

pylabrobot/resources/hamilton/hamilton_decks.py
rickwierenga: can be ignored if we go with https://github.com/PyLabRobot/pylabrobot/pull/918#discussion_r2876467568

pylabrobot/resources/hamilton/hamilton_decks.py
cmoscy: One fixed postion / channel I think makes the most sense actually. Good case with the non-contiguous tip selection, definitely a bug there with this approach. 

How about at setup, we get n_channels from the instrument, then use that to assign tip locations. 

This is nice because the 16 channel spacing can be a bit tight, so if you're an 8 channel instrument you're still getting to use the full waste block length spacing. 

Will write up a change tomorrow to check out, should be quick. Otherwise would just do 1-16.

pylabrobot/liquid_handling/liquid_handler.py
cmoscy: Nice! Yeah that would be smooth. That way liquid handler is a little more explicit, and the decks would either pass the channel positions appropriate for the instrument configuration, or get the waste block if not implemented. Could move this over to the deck resources and their specifics rather than liquid_handler if that sounds good.

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
cmoscy: A pullthrough from doing this change incrementally, will clean up. currently `is_waste_drop` is a more general "trash" resource, where `all_trash` was to discriminate for the individual waste positions if they were defined. Will be consolidated along with the other changes from the later comments.

cmoscy:
> thanks for making this PR!
> 
> the tip spot changes make sense, just a few points to discuss below
> 
> as for `multichannel_capable` and trough stuff: I would rather keep each PR self contained so we can merge them fast and have a nice clean history. would you mind making a follow up PR for those changes? my high level early feedback for that is: I would rather model the physical barrier locations than `multichannel_capable` on the containers, since `multichannel_capable` is a property of a given liquid handler rather than a resource itself. If we model that simple aspect, we should be able to work out where to place channels automatically and this stays close to the hardware which is good for ease of adding and hardware agnosticism.

No problem! I'll split out the the trough stuff and resubmit just the waste changes for now. Will work through comments above soon and submit that one clean. 

I like the feedback about the troughs, that makes sense. 

If I understand your preference correctly. I should just provide an accurate description of the addressable geometry, and let the liquid handler be 'smart' about how to arrange its channels to address it (how many in parallel / what positions). I'll tweak the trough changes to follow that general logic.

PR #919, opened 2026-03-03
Extract STAR backend into star package with mixin architecture

rickwierenga:
- Move STAR_backend.py into hamilton/star/ package
- Extract 96-head methods into STARBackend96HeadMixin (backend_96head.py)
- Extract error classes and helpers into errors.py
- Extract shared utilities (decorators, fw parsing, Head96Information) into shared.py
- Add STARBaseMixin typing base with bare annotations for mypy
- Backward-compat shim at old STAR_backend.py import path


rickwierenga:
decide it's not worth it right now, will do with the big refactor http://discuss.pylabrobot.org/t/updating-plr-api-for-machine-interfaces-discussion/445/

PR #924, opened 2026-03-06
Per-channel waste positions: assign at setup, use in discard_tips

cmoscy:
- Decks expose addressable waste via get_waste_positions(); STARlet deck provides multiple waste positions for use with discard/drop tips.

- At liquid handler setup, waste positions are assigned once per hardware channel using backend.num_channels and deck.get_waste_positions() (via get_waste_positions_for_n_channels). This mapping is stored on the liquid handler.

- discard_tips uses that mapping: each channel in use_channels goes to its assigned waste position, independent of how many channels are in the op (no longer derived from len(use_channels)).

- Single-trash decks (e.g. Opentrons, EVO) unchanged: one trash repeated per channel, same offset behavior as before. Backends unchanged; only the front-end mapping and when it is computed changed.

-  Tests added for setup population, tested on Starlet hardware


rickwierenga:
I'm actually thinking about targeting beta 2 for this PR, since it's changing the model which could be unexpectedly breaking for users. I hope to release beta 1 to the current version this week, and then merge this PR into main after.

cmoscy:
All good! If there's anything you'd like me to refine/change in the meantime just let me know.

PR #926, opened 2026-03-09
bioshake: add delay after stop shake

xbtu2:
The BioShake 3000 ELM firmware needs ~3 seconds after shakeOff for the motor to fully decelerate before the edge-locking mechanism (ELM) can operate. Without this, setElmUnlockPos returns e (error).

Fix: Added asyncio.sleep(sleep_time_after_shake) at the end of BioShake.stop_shaking() in bioshake_backend.py. This way all callers automatically get the required settling delay.


rickwierenga:
thanks!

fyi we updated ruff in  #917, so formatting is now slightly different

also I removed the parameter to keep the api simple. in this case, a good default should work for everyone

PR #407, opened 2025-02-24
Add support for Hamilton MLPrep

rickwierenga:
A (nearly) complete firmware implementation for the Hamilton MLPrep, including the ip layer, harp2 (hamilton robot protocol?) and hoi2. Tested the big five (setup, tip pickup, tip drop, aspirate, dispense) on ours. This is the world's first interface to the Prep that doesn't use the internal computer & touch screen: you can now properly program it.

Plug your computer into the robot's ethernet port (should be a passthrough to the internal board). Assign a static ip to your computer in the `192.168.100.x` range, but not `192.168.100.102` since that's the robot's static ip.

TODO:
- [ ] properly define the PrepDeck resource
- [ ] verify it works on other Preps, the harp addressing I haven't figured out 100% yet
- [ ] core grippers
- [ ] error handling
- [ ] manual channel movement works when specifying full xyz coordinates for a channel, but can we do individual dimensions like STAR?

PR #928, opened 2026-03-10
Hamilton Prep Support powered by tcp_backend 

cmoscy:
Hamilton Prep backend with basic commands. Swapped custom wire protocol in #407 with improved tcp_backend.


BioCam:
@cmoscy - you are an absolute legend !!!


https://github.com/user-attachments/assets/3096ceaf-4d2c-42c0-a841-6a016198d66f

cmoscy:
> @cmoscy - you are an absolute legend !!!
> 
>  IT_LIVES_compressed.mp4

Working on 8MPH next!

BioCam:
> Working on 8MPH next!

Ohh, is that not yet in this PR?
Because, mine is failing initialisation, but it always has with the previous backend done so too (eBay, am I right ... 😅 )

Its squeezer drive seems to fail initialisation, but with this new driver/backend I could overwrite its initialisation requirement, and can now use the independent channels with atomic commands (I think that makes it already more powerful than the official API with its "transfer" function lock?)

Unfortunately, I do not have the integrated PC so I cannot really do any work on figuring out more commands - but please keep all commands coming, I'm very happy to test everything and contribute with what I can figure out

cmoscy:
> > Working on 8MPH next!
> 
> Ohh, is that not yet in this PR? Because, mine is failing initialisation, but it always has with the previous backend done so too (eBay, am I right ... 😅 )
> 
> Its squeezer drive seems to fail initialisation, but with this new driver/backend I could overwrite its initialisation requirement, and can now use the independent channels with atomic commands (I think that makes it already more powerful than the official API with its "transfer" function lock?)
> 
> Unfortunately, I do not have the integrated PC so I cannot really do any work on figuring out more commands - but please keep all commands coming, I'm very happy to test everything and contribute with what I can figure out

Not fully, you'll see there's tip pickup and drop commands in the backend for the 8MPH. Playing with the best way(s) to expose the 8MPH to the liquid_handler resource before pushing out the full implementation with aspirate/dispense. 

I'll 100% keep you in the loop for the 8MPH work, it's awesome to have someone else to test it! 

Also, I'll see if I can get around to creating a reasonable demo tcp_backend notebook to send your way. HamiltonTCPBackend can be used directly for introspection. There's a few rough edges to it, but you can more or less get the whole api from the Prep once you know your way around.

pylabrobot/liquid_handling/backends/hamilton/prep_commands.py
```diff
@@ -0,0 +1,1741 @@
...
+      z_bottom_search_offset=z_bottom_search_offset,
+      z_bottom_offset=z_bottom_offset,
+    )
+
+
+@dataclass
+class LldParameters:
+  default_values: PaddedBool
+  z_seek: F32
+  z_seek_speed: F32
```
BioCam: I recommend naming this simply `channel_speed`

pylabrobot/liquid_handling/backends/hamilton/prep_commands.py
```diff
@@ -0,0 +1,1741 @@
...
+      bottom_search=False,
+      z_bottom_search_offset=z_bottom_search_offset,
+      z_bottom_offset=z_bottom_offset,
+    )
+
+
+@dataclass
+class LldParameters:
+  default_values: PaddedBool
+  z_seek: F32
```
BioCam: I recommend naming this `search_start_position`

BioCam:
@cmoscy, testing through the PR now step by step, a quick comment: 

I'd like to use the Prep "as the little brother of the STAR(let)"; i.e. I want the exact same protocols to work on both systems - this goes beyond PLR's current liquid_handler/frontend-backend abstraction because imo what makes Hamilton systems so powerful _is_ their backend powers.

But the Prep and STAR(let) have the same underlying backend-given powers.

-> I have started making some comments to align the `PrepBackend` method argument naming with the `STARBackend` to this aim :)

pylabrobot/liquid_handling/backends/hamilton/prep_commands.py
BioCam: note: not named `search_speed` because pLLD requires both `channel_speed` and `dispensing_drive_speed` - eliminating the meaning of "search_speed" imo

BioCam:
@cmoscy - quick question: do you have production code that is dependent on this PR content?

I've had a go at it this weekend and figured out a bunch of stuff, I have especially looked into LLD and how to align it with the STAR naming

But I some of my changes will break the current code by moving or renaming things, and I want to make sure you are okay with this first :)

BioCam:
But my MPH unfortunately appears to be genuinely broken - I will remove that dead weight after my plastic rail extension has finished printing so I can remove the channel shuttles without the bearings rolling all around my floor 😂

cmoscy:
> @cmoscy - quick question: do you have production code that is dependent on this PR content?
> 
> 
> But I some of my changes will break the current code by moving or renaming things, and I want to make sure you are okay with this first :)

Go for it! Nothing in production at present and anything that would be is tied to a specific commit anyway. 

> I've had a go at it this weekend and figured out a bunch of stuff, I have especially looked into LLD and how to align it with the STAR naming

That's awesome! Would love to take a look since the Prep seems like it represents a few of those concepts differently than the STAR (container definitions passed directly to firmware for liquid following rather than having a star/end z positions right?). There's a couple other concepts that are unique to Prep too. Did physical testing work out with the lld / following behavior you were working on? nothing that can't be bridged between the two but have been thinking on how to preserve compatibility between the two, while still giving access to specific backend if it provides additional control.

BioCam:
Awesome, then I will see this PR as a massive dev opportunity ;)

> Would love to take a look since the Prep seems like it represents a few of those concepts differently than the STAR (container definitions passed directly to firmware for liquid following rather than having a star/end z positions right?).

Yeah, that is an awesome feature of the Prep and gave me the idea for a new solution to the problem in general: 
**Add height_volume_data attribute to Container with piecewise-linear interpolation** #938 
(It also explains a bit more why Hamilton has historically always required Prep users to request new resource definitions to be added by Hamilton themselves - this empowers users to create their own defintions and still be able to use _all_ the power of the Prep... I can't wait to see all the 3D-printed resources people can use now 😅 )

260316_PLR_Container_segmentation_MINI



> Did physical testing work out with the lld / following behavior you were working on?

LLD yes, I already fixed for cLLD over the weekend and created the same `LLDMode` enum as in STAR.
Surface following distance I am still working on because I need this new PR I made + I also want human/direct-declared `surface_following_distance` to work as well, very useful and needed to use surface following distance on liquids/containers for which LLD fails.

cmoscy:
> Yeah, that is an awesome feature of the Prep and gave me the idea for a new solution to the problem in general:
> **Add height_volume_data attribute to Container with piecewise-linear interpolation** #938

This is awesome! Sounds like you're the perfect person to work the LLD and following sides out. Let me know if there's anything you'd like me to test on that front.

Based on where you're focused, I might see if I can slip the MPH8 aspirate/dispense commands into this PR before merge so things are (minimally) complete.

cmoscy:
Can ignore the review, clicked unintentionally, ha

PR #929, opened 2026-03-10
Add custom MFX carriers for DWP and bioshake

katestephenG:
Adds 3 new custom MFX carriers for DWP and bioshake.


rickwierenga:
thanks for the PR!

are these carriers people can buy from Hamilton? the mfx carriers (base) should be agnostic the modules on them. we have a docs page on creating carriers with different modules: https://docs.pylabrobot.org/resources/carrier/mfx-carrier/mfx_carrier.html.

Also, since the bioshake does not currently have a resource model yet, I do not think it works out of the box on an mfx carrier. You can make this work by assigning a custom `PlateHolder` in the meantime ...

PR #936, opened 2026-03-15
Add Micronic rack reader backend

alexjamesgodfrey:


pylabrobot/rack_reading/standard.py
```diff
@@ -0,0 +1,40 @@
...
+  DATAREADY = "dataready"
+
+
+@dataclass
+class RackScanEntry:
+  position: str
+  tube_id: Optional[str]
+  status: str
+  free_text: str = ""
+
```
rickwierenga, line 28: what is the format of position? Like A1 etc.?

are tube_id arbitrary strings or barcodes?

what is status in this data class?

what is free text?

pylabrobot/rack_reading/standard.py
```diff
@@ -0,0 +1,40 @@
...
+class RackScanResult:
+  rack_id: str
+  date: str
+  time: str
+  entries: List[RackScanEntry]
+
+
+@dataclass
+class LayoutInfo:
+  name: str
```
rickwierenga, line 40: what is this?

rickwierenga (commented):
cool! thanks for the PR

seems pretty straightforward. Just some noob questions about what these things represent

PR #949, opened 2026-03-20
Encapsulate pipette batch scheduling into backend-agnostic module

BioCam:
Extracts batch scheduling logic from `STARBackend` into `pipette_batch_scheduling.py` - a standalone, backend-agnostic module with full test coverage.

## The Problem

`probe_liquid_heights` relied on several tightly coupled private methods (`execute_batched`, `_probe_liquid_heights_batch`, `_compute_channels_in_resource_locations`, `_move_to_traverse_height`) embedded in `STARBackend`. The scheduling algorithm (X grouping, Y sub-batching) was inseparable from Hamilton-specific hardware calls, making it untestable in isolation and not easily reusable by other backends.

While per-channel spacing support was already merged (PRs #862, #870, #915), the scheduling layer had gaps: non-consecutive channel batches (e.g. [0,1,2,5,6,7]) left intermediate physical channels 3,4 unpositioned, Y batch feasibility used a single-pair check rather than full pairwise span validation, and there was no lookahead between batches.

### New module: `pipette_batch_scheduling.py`

`plan_batches()` partitions channel-position pairs into executable batches. Backend-agnostic - depends only on channel indices, positions, and spacing constraints.

### New capabilities

- **Phantom channel interpolation** - when a batch contains non-consecutive channels (e.g. [0,2,5]), the physically intermediate channels (1,3,4) are explicitly positioned at correct pairwise spacing. Previously they were left wherever they happened to be.
- **Pairwise span validation** - batch feasibility checks the full span between outermost channels via `_span_required` (sum of adjacent pairwise spacings), not just the gap between the candidate and the batch's lowest-Y member. This matters for mixed-channel instruments where `spacing(ch0->ch3)` is `s(0,1) + s(1,2) + s(2,3)`, not `3 * max(spacings)`.
- **Transition optimisation/lookahead** - between batches, idle channels are pre-positioned toward their next-needed Y coordinate by scanning forward through upcoming batches. This is a pure performance optimization that reduces Y travel time; it does not affect correctness.
- **Configurable X grouping** - replaces `round(x, 1)` (Python's banker's rounding) with `math.floor(x / tolerance) * tolerance` and exposes `x_grouping_tolerance` as a parameter.

### Structural changes

- **`probe_liquid_heights` rewritten** - replaces the 5-method delegation chain (`execute_batched` with callback closure -> `_probe_liquid_heights_batch` -> `_compute_channels_in_resource_locations` -> `_move_to_traverse_height`) with a single linear flow calling `plan_batches()`. The method is longer (262 vs 124 lines) but reads top-to-bottom without jumping between methods or files.
- **Replaces `planning.py`** - the old module provided X grouping and Y sub-batching but lacked phantom interpolation, span validation, and transition lookahead. Deleted along with its tests.
- **Bug fix** - spacings list sizing now covers `num_channels` (not just `max(use_channels)+1`), preventing `IndexError` in transition optimization.

### Not in scope

Detection parameter exposure (cLLD/pLLD kwargs) is intentionally deferred to a follow-up PR to keep this change focused on scheduling encapsulation.

## Preview
a small taste of the new functionalities (GitHub doesn't allow larger videos)

https://github.com/user-attachments/assets/10c0d58a-d4f6-4fdd-9ced-9c456c59dcac


pylabrobot/liquid_handling/pipette_batch_scheduling.py
```diff
@@ -0,0 +1,456 @@
...
+  if num_physical > 1 and num_physical % 2 != 0:
+    offsets = [o + Coordinate(0, 5.5, 0) for o in offsets]
+
+  return offsets
+
+
+def validate_probing_inputs(
+  containers: List[Container],
+  use_channels: Optional[List[int]],
+  num_channels: int,
```
rickwierenga: why does the batch scheduling know about probing?

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
rickwierenga, line 1: with the removal of `execute_batched`, the code becomes a lot less modular. and we do want to use `execute_batched` for other commands in the future

pylabrobot/liquid_handling/pipette_batch_scheduling.py
BioCam: Thanks for catching this! I renamed to `validate_channel_selections` since it validates channel indices against container lists, nothing probing-specific.

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: interesting, the current `execute_batched` seemed very limited to probing actions, which is why I removed it, but you're saying that it actually acts as an abstraction layer for _any_ function that is meant to be called after the channels have moved to their target locations - I really like this

I will work on an implementation

pylabrobot/liquid_handling/backends/hamilton/STAR_chatterbox.py
rickwierenga, line 1: why does CB need `probe_liquid_heights`? ideally it should share the same logic as the star backend equivalent. previously I just overrode the measurement bit because that part we can't know in CB

BioCam: because....
1. Chatterbox code needs to work completely interchangeably with its backend/driver counterpart without the need to say `if protocol_mode == "simulation"` every time.
2. I disagree with...

> previously I just overrode the measurement bit because that part we can't know in CB

...the simulation/chatterbox actually has a unique opportunity here:
I made it return what the liquid tracker returns - this enables true simulation :)

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: I returned it and tried to make it even more adaptive:

Screenshot 2026-03-22 at 20 15 18

BioCam: with small adjustment to this workflow we can use it for...

- probing (ztouch, cLLD, pLLD)
- aspirate
- dispense

pylabrobot/liquid_handling/backends/hamilton/STAR_chatterbox.py
rickwierenga: ok yes this makes sense

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -2290,120 +2105,197 @@ async def probe_liquid_heights(
...
     n_replicates: int = 1,
+    move_to_z_safety_after: bool = True,
     # Traverse height parameters (None = full Z safety, float = absolute Z position in mm)
     min_traverse_height_at_beginning_of_command: Optional[float] = None,
     min_traverse_height_during_command: Optional[float] = None,
     z_position_at_end_of_command: Optional[float] = None,
-    # Deprecated
-    move_to_z_safety_after: Optional[bool] = None,
+    # X grouping tolerance (mm) — containers within this distance share an X group
+    x_grouping_tolerance: Optional[float] = None,
```
rickwierenga, line 2314: comment should go in doctoring

pylabrobot/liquid_handling/pipette_batch_scheduling.py
```diff
@@ -0,0 +1,493 @@
...
+  if use_channels is None:
+    use_channels = list(range(len(containers)))
+  if len(use_channels) == 0:
+    raise ValueError("use_channels must not be empty.")
+  if not all(0 <= ch < num_channels for ch in use_channels):
+    raise ValueError(
+      f"All use_channels must be integers in range [0, {num_channels - 1}], got {use_channels}."
+    )
+  if len(use_channels) != len(set(use_channels)):
+    raise ValueError("use_channels must not contain duplicates.")
```
rickwierenga: duplicate channels were supported in the past

pylabrobot/liquid_handling/pipette_batch_scheduling.py
```diff
@@ -0,0 +1,493 @@
...
+def validate_channel_selections(
+  containers: List[Container],
+  use_channels: Optional[List[int]],
+  num_channels: int,
+) -> List[int]:
+  """Validate and normalize channel selection.
+
+  If *use_channels* is ``None``, defaults to ``[0, 1, ..., len(containers)-1]``.
+
+  Returns:
```
rickwierenga, line 155: this function should probably live outside of batch scheduling, and can also be used elsewhere in LH

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -2290,120 +2105,197 @@ async def probe_liquid_heights(
...
+        {ch: min_traverse_height_at_beginning_of_command for ch in use_channels}
+      )
+
+    # Compute Z positions
+    z_cavity_bottom: List[float] = []
+    z_top: List[float] = []
+    for resource in containers:
+      z_cavity_bottom.append(resource.get_location_wrt(self.deck, "c", "c", "cavity_bottom").z)
+      z_top.append(resource.get_location_wrt(self.deck, "c", "c", "t").z)
+
```
rickwierenga: list comprehension?

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -2056,229 +2053,47 @@ class PressureLLDMode(enum.Enum):
...
-    """
-    if traverse_height is None:
-      await self.move_all_channels_in_z_safety()
-    else:
-      if channels is None:
-        channels = list(range(self.num_channels))
-      await self.position_channels_in_z_direction(
-        {channel: traverse_height for channel in channels}
-      )
-
```
rickwierenga, line 2127: this pattern is recurring, even with liquid probing. why delete the helper?

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -2290,120 +2105,197 @@ async def probe_liquid_heights(
...
+              if lld_mode == self.LLDMode.GAMMA:
+                msg += " Consider using pressure-based LLD if liquid is believed to exist."
+              logger.warning(msg)
+            else:
+              raise result
+          elif isinstance(result, Exception):
+            raise result
+          else:
+            height = current_absolute_liquid_heights[ch_idx]
+          absolute_heights_measurements[ch_idx].append(height)
```
rickwierenga: this batch function is modifying code outside its scope, namely `absolute_heights_measurements`

pylabrobot/liquid_handling/pipette_batch_scheduling.py
```diff
@@ -0,0 +1,493 @@
...
+  for b in batches:
+    x_key = round(b.x_position, 1)
+    x_groups.setdefault(x_key, []).append(b)
+
+  print(f"{label}:")
+  xg_keys = list(x_groups.keys())
+  for xg_i, x_key in enumerate(xg_keys):
+    xg_batches = x_groups[x_key]
+    is_last_xg = xg_i == len(xg_keys) - 1
+    xg_branch = "\u2514" if is_last_xg else "\u251c"
```
rickwierenga: what are these unicode things?

pylabrobot/liquid_handling/pipette_batch_scheduling.py
```diff
@@ -0,0 +1,493 @@
...
+  """
+  sorted_chs = sorted(channels)
+  for k in range(len(sorted_chs) - 1):
+    ch_lo, ch_hi = sorted_chs[k], sorted_chs[k + 1]
+    cumulative = 0.0
+    for phantom in range(ch_lo + 1, ch_hi):
+      cumulative += _min_spacing_between(spacings, phantom - 1, phantom)
+      if phantom not in y_positions:
+        y_positions[phantom] = y_positions[ch_lo] - cumulative
+
```
rickwierenga, line 141: this can also be used in position_channels_in_y_direction

functions that modify input parameters are dangerous, it is better to return a new list and not modify the original

rickwierenga:
> Phantom channel interpolation - when a batch contains non-consecutive channels (e.g. [0,2,5]), the physically intermediate channels (1,3,4) are explicitly positioned at correct pairwise spacing. Previously they were left wherever they happened to be.

that is not entirely true since `position_channels_in_y_direction` would make space for them, but with this PR the positions are computed during planning which is nicer

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -2056,229 +2053,47 @@ class PressureLLDMode(enum.Enum):
...
-
-    return relative_to_well
-
-  def _get_maximum_minimum_spacing_between_channels(self, use_channels: List[int]) -> float:
-    """Get the maximum of the set of minimum spacing requirements between the channels being used"""
-    sorted_channels = sorted(use_channels)
-    max_channel_spacing = max(
-      self._min_spacing_between(hi, lo) for hi, lo in zip(sorted_channels[1:], sorted_channels[:-1])
-    )
-    return max_channel_spacing
```
rickwierenga, line 2238: how come this function is no longer needed?

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: Removed, agree

pylabrobot/liquid_handling/pipette_batch_scheduling.py
BioCam: Yes, thank you for catching this!
Indeed crucial for e.g. multi-aspirates
Changed it back

pylabrobot/liquid_handling/pipette_batch_scheduling.py
BioCam: I agree; but not sure where it should go. Do you have a preferred home for it?

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: done

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: Agreed it's a recurring pattern, but each site has slightly different channel sets and conditions now.
The helper was just an if/else around two one-liner firmware commands. 
It felt to me as if the indirection cost more clarity than it saved.

pylabrobot/liquid_handling/pipette_batch_scheduling.py
BioCam: They're box-drawing characters for print_batches; they can be either unicode or symbols directly.
I replaced unicode escapes with the actual characters — ├, └, │, ─ to make them mre readable.

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: Good catch, just refactored. `_probe_batch_heights` now returns its measurements and `execute_batched` collects them.
The callback no longer mutates outer state.

BioCam:
> > Phantom channel interpolation - when a batch contains non-consecutive channels (e.g. [0,2,5]), the physically intermediate channels (1,3,4) are explicitly positioned at correct pairwise spacing. Previously they were left wherever they happened to be.
> 
> that is not entirely true since `position_channels_in_y_direction` would make space for them, but with this PR the positions are computed during planning which is nicer

You're right, that was a mistake of me to phrase it like this! 
`position_channels_in_y_direction(make_space=True)` already handled intermediate channels reactively.

The difference is that `plan_batches` now computes their positions upfront during planning rather than relying on the make-space correction at movement time.
I like this because it enables deterministic positioning: phantom channels land at computed positions rather than wherever make_space pushes them relative to their previous state.

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: Because it only had two callers:
1. `_compute_channels_in_resource_locations` -> deleted and logic moved to `plan_batches`
2. `pierce_foil`

For `pierce_foil` I inlined the equivalent computation directly at the call site since it was a one-liner.

pylabrobot/liquid_handling/pipette_batch_scheduling.py
BioCam: Good point on both counts.

Changed `_interpolate_phantoms` to return a new dict instead of mutating the input.

Reuse in `position_channels_in_y_direction` is a natural next step but outside of the scope of this PR imo.

pylabrobot/liquid_handling/backends/hamilton/STAR_chatterbox.py
```diff
@@ -316,3 +330,73 @@ async def position_channels_in_y_direction(self, ys, make_space=True):
...
+      use_channels=use_channels,
+      targets=containers,
+      channel_spacings=self._channels_minimum_y_spacing,
+      x_tolerance=x_grouping_tolerance,
+      wrt_resource=self.deck,
+      resource_offsets=resource_offsets,
+    )
+
+    print_batches(batches, use_channels, containers, label="probe_liquid_heights plan")
+
```
rickwierenga: why not override execute_batched?

`probe_liquid_heights` has annoyingly much logic in the chatterbox

pylabrobot/liquid_handling/pipette_batch_scheduling.py
```diff
@@ -0,0 +1,496 @@
...
+def plan_batches(
+  use_channels: List[int],
+  targets: Union[List[Container], List[Coordinate]],
+  channel_spacings: Union[float, List[float]],
+  x_tolerance: float,
+  wrt_resource: Optional[Resource] = None,
+  resource_offsets: Optional[List[Coordinate]] = None,
+) -> List[ChannelBatch]:
+  """Partition channel–target pairs into executable batches.
+
```
rickwierenga, line 371: having targets and wrt_resource is complex. can we not resolve this using a helper function and just only accept coordinates here? this will also make it more robust for other operations in the future

pylabrobot/liquid_handling/pipette_batch_scheduling.py
```diff
@@ -0,0 +1,496 @@
...
+    y_pos = [c.y for c in coordinates]
+
+  # Normalize scalar spacing to per-channel list.
+  max_ch = max(use_channels)
+  if isinstance(channel_spacings, (int, float)):
+    spacings: List[float] = [float(channel_spacings)] * (max_ch + 1)
+  else:
+    spacings = list(channel_spacings)
+    if len(spacings) < max_ch + 1:
+      spacings.extend([spacings[-1]] * (max_ch + 1 - len(spacings)))
```
rickwierenga: the else case is a clear bug

pylabrobot/liquid_handling/pipette_batch_scheduling.py
```diff
@@ -0,0 +1,496 @@
...
+          continue
+        group_channels = [use_channels[i] for i in c_indices]
+        spread = compute_single_container_offsets(
+          container=containers[c_indices[0]],
+          use_channels=group_channels,
+          channel_spacings=channel_spacings,
+        )
+        if spread is not None:
+          for i, idx_val in enumerate(c_indices):
+            adjusted_y[idx_val] += spread[i].y
```
rickwierenga: this breaks when one channel goes into a container more than once

rickwierenga: also channel positioning in containers is probably logic that should not live in this god method

pylabrobot/liquid_handling/pipette_batch_scheduling.py
BioCam: Done; I've split it into two functions:

- `resolve_container_targets()` handles container-to-coordinate conversion (position lookup, auto-spreading for multi-channel-to-same-container)
- `plan_batches()` now only accepts `List[Coordinate]` - no more `wrt_resource`, `resource_offsets`, or `Union` target type

pylabrobot/liquid_handling/pipette_batch_scheduling.py
BioCam: yikes, yes you are absolutely right!

The silent padding was hiding caller errors. Changed channel_spacings from Union[float, List[float]] to `List[float]` across `plan_batches`, `resolve_container_targets`, and `compute_single_container_offsets`. Since these are new functions there's no backwards compatibility concern.

A too-short list now raises `ValueError` immediately. The scalar path and the padding logic are both gone.

pylabrobot/liquid_handling/pipette_batch_scheduling.py
BioCam: I love the term "god method" 😅 

I agree on both accounts:

1. Restored the duplicate-channel guard that `group_by_x_batch_by_xy` had - `_partition_into_y_batches` now skips batches where the physical channel is already assigned, so duplicate channels (e.g. multi-aspirate) are correctly serialized.
2. Extracted all container logic into a separate `resolve_container_targets()` function.
`plan_batches` now only accepts `List[Coordinate]` - no containers, no auto-spreading, no wrt_resource.

rickwierenga: > Restored the duplicate-channel guard that group_by_x_batch_by_xy had 

it had no guard, it just worked with repeated channels

rickwierenga:
sweet

BioCam:
Thank you @rickwierenga for the review process throughout and working on this together!

It is going to change the game in liquid handler usage

PR #951, opened 2026-03-20
feat: add applied biosystems plate adapter

jemancuso:
Add Applied Biosystems™ MicroAmp™ Splash-Free 96-Well Base, Item No.: 4312063
https://www.fishersci.com/shop/products/applied-biosystems-microamp-splash-free-96-well-base/4312063


BioCam:
Hi @jemancuso,

Thank you very much for contributing this to PyLabRobot!

A couple of points to get this merged:

1. PyLabRobot Resources are organised & named first by manufacturer. This avoids confusion about vendors and vendor-specific catalogue numbers masking the manufacturer's catalogue numbers.
Applied Biosystems is not a company (anymore) but a brand owned by Thermo Fisher Scientific (see organisational structure in the [PLR docs](https://docs.pylabrobot.org/stable/resources/library/thermo_fisher.html) and screenshot below).
That is the reason why PLR does not have and shouldn't have an "Applied Biosystems" folder
-> please remove the folder and move the definition into `/pylabrobot/resources/thermo_fisher`

2. The naming convention should be `<manufacturer (lowercase, PLR-convention)>_<brand (PLR-convention)_<throughput>_<resource_type>_<miscellaneous identifier (a bit of a wildcard)`
-> I would suggest renaming to `thermo_AB_96_plateadapter_MicroAmp`?

3. Please ensure that the first line of the docstrings specifies the catalogue number; example: 
Screenshot 2026-03-21 at 17 53 16
(this is important for parsers and docstring consistency)

4. Please provide the link to the manufacturer in the docstring too: https://www.thermofisher.com/order/catalog/product/4312063

> **Note:** You may notice some inconsistencies in naming conventions in the existing codebase - this is a known legacy issue that is actively being cleaned up.
Please follow the conventions outlined above rather than using existing definitions as a reference.

---

Screenshot 2026-03-21 at 17 42 14

BioCam:
@jemancuso, please feel free to ignore point 3 - it is a bit nitpicky 😂 
The reason I do this is to enable classic parsers based on first line of docstring regular expression identification, but it's not essential

pylabrobot/resources/__init__.py
```diff
@@ -2,6 +2,7 @@
 from .agenbio import *
 from .agilent import *
 from .alpaqua import *
+from .applied_biosystems import *
```
BioCam: please delete

jemancuso: oopsie

rickwierenga:
thanks @jemancuso and @BioCam

PR #961, opened 2026-03-25
Expose `STARBackend.move_channel_stop_disk_z()`

BioCam:


BioCam:
@rickwierenga, @ben-ray, @cmoscy, @jrast 

Should we call this `STARBackend.move_channel_stop_disc_z()` and apply "stop_disc" naming for everything that refers to it?
It seems clearer to me than "probe" because very little Hamilton documentation references "probe", as opposed to "stop disc":

Screenshot 2026-01-01 at 14 52 20

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -4530,11 +4530,82 @@ async def move_channel_y(self, channel: int, y: float):
...
+    bottom of the stop disc. In that case, this command is effectively
+    equivalent to :meth:`move_channel_probe_z` for the same numeric Z value.
+
+    When the firmware state indicates that a tip is mounted on the channel,
+    the same Z position instead refers to the physical end of the tip. In
+    this case, the numeric Z value used with this method may differ from the
+    probe Z position used with :meth:`move_channel_probe_z` for the same
+    physical height above the deck.
+    """
     await self.position_single_pipetting_channel_in_z_direction(
```
rickwierenga, line 4559: should we make this only for the case when tips are mounted, and ask people to use `move_channel_probe_z` for other cases? this conditional method behavior is confusing

BioCam: Yes I think you're right - should be a separate PR though

rickwierenga: Lets do it here since they go nicely together

Also we can consider renaming tips method to move tip bottom or something to make it extra clear

BioCam: Are you sure?

`STARBackend. move_channel_z()` is a core method and will therefore require an extended deprecation period away from currently expected behaviour

I use it in every aP

cmoscy: It's not uncommon to account for the length of a mounted tool/effector when calling move commands on other hardware too. So it seems like STARBackend. move_channel_z() is plenty clear if it stays with it's current behavior/naming. 

It is great to have the STARBackend.move_channel_probe_z() as a clear way to explicitly move wrt the stop disc though, regardless of tip state. 

Just my 2c

rickwierenga: sounds ambiguous to me to have `move_channel_z` and `move_channel_probe_z`. It's a bad API. We can cleanly transition this in the v1 change through the legacy module if we add this to `v1b1`. But also a simple deprecation warning for a method rename is easy enough

cmoscy: Fair enough, maybe something like adding an arg `with_respect_to ['end','stopdisc']`. 1 function, with a clear callout of what you're moving the channel with respect to?

rickwierenga: I generally like separate functions better than parameters fundamentally changing a functions behavior

rickwierenga: @BioCam can you rename+depr then merge?

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -4530,11 +4530,135 @@ async def move_channel_y(self, channel: int, y: float):
...
+    """
+    warnings.warn(
+      "move_channel_z is deprecated. Use move_channel_stop_disk_z() for moves without a tip attached "
+      "or move_channel_tool_z() when a tip/tool is attached.",
+      DeprecationWarning,
+      stacklevel=2,
+    )
     await self.position_single_pipetting_channel_in_z_direction(
       pipetting_channel_index=channel + 1, z_position=round(z * 10)
     )
```
rickwierenga, line 4562: why not use the method you tell people to use?

BioCam: Which one? We tell users to use one of two depending on whether there is a tip __or__ is no tip attached.

But for deprecation we must not change the behaviour of the deprecated method.
Switching to either of the two new methods would lead to a deviation of the established behaviour.

rickwierenga: I was thinking based on tip status, but yeah might as well have firmware continue to do that for now. let's ship it

PR #966, opened 2026-03-26
Add BTX plate definition

hazlamshamin:


pylabrobot/resources/btx/plates_tests.py
rickwierenga, line 1: test is not necessary for a plate definition :)

pylabrobot/resources/btx/plates.py
```diff
@@ -0,0 +1,61 @@
...
+  WellBottomType,
+)
+
+
+def BTX_96_wellplate_125ul_Fb_2mm(name: str) -> Plate:
+  """BTX 96-well disposable electroporation plate, 2 mm gap.
+
+  - BTX part no.: 45-0450 / 45-0450-M
+  - External dimensions: 127.8 x 85.5 x 15.9 mm
+  - Well pitch: 9.0 mm x 9.0 mm
```
BioCam: no need because this is the standard assumed
(see https://www.slas.org/SLAS/assets/File/public/standards/ANSI_SLAS_4-2004_WellPositions.pdf, page 4 & 5)

We only have to declare if a plate is deviating from the SLAS plate standard.

BioCam:
@hazlamshamin - this is very nice! Thank you for contributing this!

One final comment/recommendation: 

I have recently updated the way that `Container`s compute height/volume from each other (see [Add height_volume_data attribute to Container with piecewise-linear interpolation
#938](https://github.com/PyLabRobot/pylabrobot/pull/938) )

I have made a showcase PR in [Update Corning 3603 with empirical cLLD height_volume_data
#948](https://github.com/PyLabRobot/pylabrobot/pull/948).

I think it would be very powerful to add a `height_volume_data` dict for this `Plate`/`Well`.

If you do not have a  liquid handler that can perform the liquid height measurements automatically for you then I recommend, creating a simple dict:
```python
height_volume_data = {
    0.0 : 0.0,
    10.0 : 165.0 # maximum volume you measured that fits into the height of the well
}
```

This will essentially just calculate a perfectly linear interpolation between the absolute min and absolute max.
This will not be great for irregular geometries but it seems that the well is actually a pretty perfect cuboid anyway?

pylabrobot/resources/btx/plates_tests.py
hazlamshamin: whoops, removed and noted

pylabrobot/resources/btx/plates.py
hazlamshamin: handled, thanks!

hazlamshamin:
@BioCam wow that is a useful update i just noticed, and yeah this plate is cuboid, added the simple dict so max volume reflects 160µL i tested manually (with just eyes).

BioCam:
> max volume reflects 160µL i tested manually (with just eyes).

@hazlamshamin, just to quickly check:
you measured by pipetting 160 uL into the well and then visually checked that the surface of the liquid aligns with the top of the well?

BioCam:
> wow that is a useful update i just noticed

I'm glad you find it useful - I wish more liquid handlers would have the advanced sensing capabilities of Hamilton liquid handlers (i.e. both ztouch/force sensing, and liquid-level detection -> both needed for automated `height_volume_data` generation for a given `Container`).
But as shown in this example, we can generate fairly accurate `height_volume_data` by simply thinking through what the `Well` geometry is and a manual pipette :)

hazlamshamin:
> > max volume reflects 160µL i tested manually (with just eyes).
> 
> 
> 
> @hazlamshamin, just to quickly check:
> 
> you measured by pipetting 160 uL into the well and then visually checked that the surface of the liquid aligns with the top of the well?

yeap, that's the best i could do with this plate because the structure has lots of layers to properly measure the internal and external lines. plus the manufacturer does not release the details, which guess it's related to their patents maybe. 

i have used this definition with OT2 for 50-100uL (the common range for electroporation) and they works fine.

imageimageimage

BioCam:
Thank you, @hazlamshamin - great work !

PR #988, opened 2026-04-07
feat: add maintenance mode command for Inheco SCILABackend

linx5o:
Tested locally. Confirmed working. 
Device will shutdown after this command is sent.
Requires manual power on.


rickwierenga:
thanks!

PR #990, opened 2026-04-10
Add the pedestal height to the _P version of the L5AC plate carrier

tetsuwan-yoo:
PLT_CAR_L5AC_P_A00 is missing pedestal_size_z.

Measuring this inhouse with calipers results in -4.74 mm, same as PLT_CAR_L5AC_A00.


rickwierenga:
thanks!

PR #985, opened 2026-04-03
Use Aravis camera driver instead of PySpin for Cytation

vcjdeboer:
## Summary

- Adds an Aravis-based camera driver (`AravisCamera`) as a drop-in alternative to PySpin for FLIR BlackFly cameras on the Cytation. Eliminates the Spinnaker SDK dependency and the Python 3.10 version cap that PySpin imposes.
- Splits `cytation.py` into `cytation1.py` and `cytation5.py` — each device creates its own driver internally based on a `camera` parameter (`"aravis"` or `"pyspin"`).
- Adds `CytationAravisDriver` (connection + optics + camera) and `CytationAravisMicroscopyBackend` (capture orchestration) following the STARDriver pattern.

## New files

| File | Role |
|------|------|
| `aravis_camera.py` | Standalone BlackFly driver using Aravis/GenICam via PyGObject |
| `aravis_simulated.py` | Simulated camera for testing without hardware |
| `cytation_aravis_driver.py` | Driver layer: BioTekBackend serial + AravisCamera, owns connections and optics |
| `cytation_aravis_microscopy.py` | MicroscopyBackend: capture orchestration, calls driver methods |
| `cytation1.py` | Cytation 1 device class (split from cytation.py) |
| `cytation5.py` | Cytation 5 device class (split from cytation.py) |

## Modified files

| File | Change |
|------|--------|
| `cytation.py` | Keeps CytationBackend (PySpin), re-exports device classes for backwards compat |
| `__init__.py` | Updated exports |

## Test plan

- [x] Hardware-tested on Cytation 1 with BlackFly BFLY-U3-13S2M via Aravis on macOS
- [x] Camera connect/disconnect, exposure, gain, image capture verified
- [x] Single capture (brightfield 2.5x) and multi-image tiling (2x2) working
- [x] Filter/objective discovery matches firmware (DAPI, GFP, RFP + objectives)
- [ ] CI: Aravis is an optional dependency (PyGObject + system library) — no import at module level, guarded by `HAS_ARAVIS`
- [ ] PySpin path unchanged — existing users unaffected

**Note:** Some CI checks may fail due to pre-existing lint issues on v1b1 (e.g., EL406 unused import) that are not from this PR.

PR #992, opened 2026-04-13
Migrate the full codebase to structured concurreny

burnpanck:
This is my go at https://discuss.pylabrobot.org/t/structured-concurrency/475.
With 136 changed files and about 3k changed lines, certainly not a small refactor. Luckily, the majority of them are changes of`await asyncio.sleep` to `await anyio.sleep`. I have reviewed every single changed line in this refactor, and many of the more intricate refactors are hand-written and agent-verified, rather than the other way around.

All tests pass locally, though unfortunately, many of the more intricate changes are in the reader loops of the hardware interfaces, which are often untested or partially replaced by mocks, so a significant testing gap remains. An earlier commit from this refactor was battle-tested on a Hamilton STAR for a couple of days, and confirmed to significantly increase robustness of the code with respect to aborted protocols due to errors and cancellation. With this refactor, you should be able to expect from your PyLabRobot liquid handling system that it resumes into a new lifespan context without a hitch, no matter how you exited the previous one.

I encourage everyone to run their own LLMs/Agents over this refactor to look for issues or changes in behavior (other than improvements with respect to structured concurrency) and for course test on your favorite piece of hardware. Usually, it should be enough to ping-pong a few commands around the hardware, if that works, you are good. The many hardware-specific commands didn't change, just the low-level firmware command handling.

Since the main branch has moved forward in the meantime, I will try to periodically merge main (every couple of days). Once it's read to merge, we can also squash the commit history of this refactor.

Things that are still TODO:
- The visualizer
- Documentation changes, docstrings and error messages referencing `setup`.


burnpanck:
Uh oh, I see you test again python 3.9. Is that a hard requirement? It's EOL since last year...

rickwierenga:
sick PR, first glance is looking good but will respond properly in a bit

> Uh oh, I see you test again python 3.9. Is that a hard requirement? It's EOL since last year...

any reason to not support it? I generally want to support as many versions and configurations as possible, but if there's something super nice that requires 3.10+ LMK

burnpanck:
> any reason to not support python 3.9?

I'm not aware of anything concrete at this point, I can give it a try.

IHMO, supporting outdated versions hurts code quality, because best practices change over time. The cognitive effort to properly phase-out old versions increases significantly the older these versions get. It's a compromise that projects need to balance when choosing their support policy. Personally, I'm used to developing for PSF active support versions (i.e. ±2 versions), and only guaranteeing extended support if an essential dependency requires it. In projects with OEM vendor dependencies (PySpin, harvesters, looking at you), unfortunately, that happens often.

rickwierenga:
> supporting outdated versions hurts code quality, because best practices change over time

agree in general, but what's the concrete example here? (for example 3.8 doesn't have asyncio.to_thread)

all else being equal, might as well support 3.9

> PySpin

not for much longer! thanks to @vcjdeboer #985

burnpanck:
> agree in general, but what's the concrete example here?

Now that I think of it, [PEP 654 (`except*` and exception groups)](https://peps.python.org/pep-0654/) was made specifically to support structured concurrency. My refactor doesn't use any, because there is relatively little error-recovery built-in in the code-base; but it would be the modern technical basis of `ChannelizedError`. It's python 3.11 though.

burnpanck:
Finally passes all the checks. The typechecker uncovered one or two real bugs, some of which were actually already in the pre-refactor codebase (one I recall: `pylabrobot.io.ftdi.FTDI` accidentally inherited from `io.IOBase` from the standard library instead of `pylabrobot.io.io.IOBase`).

rickwierenga:
I like the transition to aynio and related changes

> TODO: Design question: Do we need kwargs? We could elevate

yes, setup definitely needs kwargs. `setup` currently does to things: initialize the connection to the machine, and initialize the machine itself. For the machine initialization, you may for example choose to skip a certain part like in STAR setup you can say `skip_autoload=True` if you won't use it within the script.

burnpanck:
Ok, kwargs are in fact already mostly implemented in this PR, just not yet declared public API. Specifically, in this branch, a machine exposes:
- `lifespan(**kwargs)`: Async context manager managing the lifespan of the "active live" of that machine within pylabrobot; meaning that inside this context, all machine invariants are attempted to be respected, and if they can't,
you are guaranteed to receive an exception. `**kwargs` are passed to machine implementations, which usually mean they apply to that second part of the lifespan's task: Initialising the machine to a certain state. We should probably elevate this to an API requirement: kwargs do not affect behaviour once the context is fully entered, only the state you might observe at that point. I believe anything else is a footgun, because you can't introspect the effect.
- `__aenter__/__aexit__`: Always delegates to `lifespan()`, but doesn't offer `**kwargs`.
- `setup(**kwargs)/stop()`: cause `lifespan` to be entered on a global management task (which is automatically generated on `asyncio`, but that can't be done on `trio` - we'd need an explicit API for that). Note that through this API, you lose some of the desirable properties of structured concurrency: A clear owner for errors that are unrelated to any particular frontend call. A clear owner of the `contextlib.Context` which applies to the machine lifespan (though today, PLR doesn't play that well with separate task contexts anyway, e.g. looking at `with lh.use_channels`).

burnpanck:
With this latest commit, I changed `_lifespan(**kwargs)` to `lifespan(**kwargs)` and thereby made it part of the public API.

rickwierenga:
the diff is too massive compared to refactored main. Please reopen the PR against main in the future

PR #993, opened 2026-04-14
make sure liquid handler can restore state before connecting

burnpanck:
Currently, attempting to restore a liquid handler state before the first `.setup()` call may fail because the serialized state may contain head state information, but the liquid handler doesn't know about it's heads yet. However, restoring the state after `.setup()` is semantically wrong: Both `.setup()` and `.stop()` may actively change the effective state, so, neither of these calls should intervene between a safe/restore cycle. This PR solves this by initializing the head state upon load if no head state has been initialized before. That is a) the correct thing to do, and b) still allows a backend to update the frontend with new state information if it is capable of tip presence detection.


rickwierenga:
thanks!

rickwierenga:
apologies for late response

PR #999, opened 2026-04-18
Nimbus v1 tcp aligned

cmoscy:
Mostly @rickwierenga's nimbus v1 port. Aligned with updated tcp_client, introspection, and enriched HoiError+Channelized error driver.


pylabrobot/hamilton/liquid_handlers/prep/prep_commands.py
rickwierenga, line 1: is this supposed to be in this PR? happy either way

cmoscy: Not intentionally, but also likely won't change with the prep since it's just command shapes. Must have let that one through since this is a cherry pick. (had both up to try and make them a little more consistent with each other + star) Feel free to remove it if you'd like to keep it pure.

rickwierenga: ill just keep it, v1b1 is for dev anyway

rickwierenga:
One big thing this PR adds is liquid classes to the Nimbus, which hasn't existed in v0. I actually do not like liquid classes, because they are redundant with the parameters that you can already pass through standard python (whether directly through the capability or through backend params). A "liquid class" is just a 'set of parameters'. It's a relic from venus and anti pythonic imo, so long term I hope to get rid of them, in favor of just storing these 'sets of parameters' in `NimbusPIPAspirateParams` (or nested `NimbusPIPBackend.AspirationParams` to be more closely aligned with the star and other backends etc.)

Like you can just store it as follows

```python
liquid_class = NimbusPIPBackend.AspirationParams(
  x=1,
  y=2,
)
```

this is essentially equivalent to a "liquid class" concept in lab automation, and removes the ambiguity (with separate an hlc parameter, you have three sources of information: explicitly passed, hlc, and backend default. not good)

would that work for you?

cmoscy:
> One big thing this PR adds is liquid classes to the Nimbus, which hasn't existed in v0. I actually do not like liquid classes, because they are redundant with the parameters that you can already pass through standard python (whether directly through the capability or through backend params). A "liquid class" is just a 'set of parameters'. It's a relic from venus and anti pythonic imo, so long term I hope to get rid of them, in favor of just storing these 'sets of parameters' in `NimbusPIPAspirateParams` (or nested `NimbusPIPBackend.AspirationParams` to be more closely aligned with the star and other backends etc.)
> 
> Like you can just store it as follows
> 
> ```python
> liquid_class = NimbusPIPBackend.AspirationParams(
>   x=1,
>   y=2,
> )
> ```
> 
> this is essentially equivalent to a "liquid class" concept in lab automation, and removes the ambiguity (with separate an hlc parameter, you have three sources of information: explicitly passed, hlc, and backend default. not good)
> 
> would that work for you?

That works for me. Agree that could be a lot better and direct. Does this also include the volume correction curves that come along? Since some of those defaults can be useful as a starting point/reference. 

Want me to pull the liquid_class wiring from nimbus and prep for now? Mainly had it just for consistency with star (+vol corrections)

rickwierenga:
that is an unsolved question :) open to suggestions. I think passing a function/calibration curve would be neat, or if no function/curve is passed just use the value directly, with the default being the current default hlc curve.

I am saying it more from a "long term I want to get rid of liquid classes and figure out a better way" point of view, so would prefer to not add it in new code and rather focus on getting it out of old code if that makes sense.

cmoscy:
> that is an unsolved question :) open to suggestions. I think passing a function/calibration curve would be neat, or if no function/curve is passed just use the value directly, with the default being the current default hlc curve.
> 
> I am saying it more from a "long term I want to get rid of liquid classes and figure out a better way" point of view, so would prefer to not add it in new code and rather focus on getting it out of old code if that makes sense.

Totally makes sense. Maybe I can play around a bit and see if there's any solutions there. If I get any ideas I'll put them up on the forum

cmoscy:
Just to clarify, did you want me to pull hlc-related behavior out before merging this? Or just work on moving away from hlc behaviors as this gets built out alongside prep from this branch?

rickwierenga:
Did that already actually, just forgot to push ... will do when I get home

Do you want to keep working on this branch or should I merge it? I thought this just adds/refactors TCP and some nimbus stuff and then we rebase the prep branch onto v1b1 after merging this one. But either way works

cmoscy:
> Did that already actually, just forgot to push ... will do when I get home
> 
> Do you want to keep working on this branch or should I merge it? I thought this just adds/refactors TCP and some nimbus stuff and then we rebase the prep branch onto v1b1 after merging this one. But either way works

Ah, no prob. Merging is probably easiest. I already split off the prep into a separate branch from here anyway. But whatever is most convenient.

cmoscy:
When there's a good time. Is there anything else you'd want reviewed/changed before merging this in? 

Prep v1 builds on top of this, so would be easiest to merge this -> then open prep pr. 

But can always rebase prep_v1 if that's best too. Just let me know. 

Asking since Prep is ready :)

pylabrobot/hamilton/liquid_handlers/nimbus/driver.py
```diff
@@ -20,30 +23,68 @@
...
     port: int = 2000,
-    read_timeout: float = 30.0,
+    read_timeout: float = 300.0,
     write_timeout: float = 30.0,
     auto_reconnect: bool = True,
     max_reconnect_attempts: int = 3,
+    connection_timeout: int = 600,
+    error_codes: Optional[Dict[Tuple[int, int, int, int, int], str]] = None,
   ):
+    merged_error_codes = {**NIMBUS_ERROR_CODES, **(error_codes or {})}
```
rickwierenga: I do not understand why error codes would be passed through init

pylabrobot/hamilton/tcp/client.py
```diff
@@ -0,0 +1,747 @@
...
+              action.name,
+              action,
+              len(entries),
+              summary,
+            )
+            raise HoiError(
+              exceptions=entry_errors,
+              entries=list(entries),
+              raw_response=response_message.hoi.params,
+            )
```
rickwierenga, line 660: I think we should always `raise_on_error`. If the caller wants to not raise they can try/except. This is more readable and simple

pylabrobot/hamilton/tcp/messages.py
```diff
@@ -75,9 +69,9 @@ class HoiParams:
     [0x03|0x00|0x04|0x00|100][0x0F|0x00|0x05|0x00|"test\0"][0x1C|0x00|...array...]
 
     params = (HoiParams()
-              .i32(100)
-              .string("test")
-              .u32_array([1, 2, 3])
+              .add(100, I32)
+              .add("test", Str)
+              .add([1, 2, 3], U32Array)
```
rickwierenga, line 74: why this change in API? I really liked the previous api!

pylabrobot/hamilton/tcp/messages.py
```diff
@@ -89,365 +83,361 @@ def _add_fragment(self, type_id: int, data: bytes, flags: int = 0) -> "HoiParams
...
-    Format: [element0][element1]... (NO count prefix - count derived from DataFragment length)
 
-    Note: BOOL_ARRAY uses flags=0x01 in the DataFragment header (unlike other types which use 0x00).
-    """
-    writer = Writer()
-    for val in values:
-      writer.u8(1 if val else 0)
-    return self._add_fragment(HamiltonDataType.BOOL_ARRAY, writer.finish(), flags=0x01)
+def inspect_hoi_params(params: bytes) -> List[dict]:
+  """Inspect raw HOI params bytes fragment-by-fragment for debugging.
```
rickwierenga, line 294: `params` as the name for the buffer is a bit weird because it also returns parameters, just a different format. Maybe `param_buffer`?

pylabrobot/hamilton/tcp/introspection.py
```diff
@@ -669,16 +1744,56 @@ async def get_enums(self, address: Address, interface_id: int) -> List[EnumInfo]
...
+
+    Example:
+      raw, fragments = await intro.get_structs_raw(mph_addr, 1)
+      for i, f in enumerate(fragments):
+        print(f\"{i}: type_id={f['type_id']} len={f['length']} decoded={f['decoded']!r}\")
+    """
+    command = GetStructsCommand(address, interface_id)
+    result = await self.backend.send_command(command, ensure_connection=False, return_raw=True)
+    (params,) = result
+    return params, inspect_hoi_params(params)
```
rickwierenga: this method is not called anywhere afaict. Also, why does it need to return both paras and the dicts? don't the dicts contain all information? if not, then maybe it should be parsed by the caller.

rickwierenga: this goes down a rabbit hole with inspect_hoi_params having untyped dictionaries but maybe it's not even used?

pylabrobot/hamilton/tcp/introspection.py
```diff
@@ -39,98 +142,119 @@ def resolve_type_id(type_id: int) -> str:
...
-_COMPLEX_TYPE_IDS = {60, 61, 64, 78, 81, 82, 85}  # Types that need additional bytes
+# Rows = firmware scalar or array kinds; columns = In, Out, InOut, RetVal
+# (HoiParameterType.Direction). Source: vendor protocol reference mHoiParamTypes[31,4].
+
+
+@dataclass(frozen=True)
+class _HoiTypeRow:
+  """One row in vendor mHoiParamTypes[31,4] with readable display metadata."""
+
+  dotnet_name: str
```
rickwierenga: why is dotnet name here?

pylabrobot/hamilton/tcp/introspection.py
```diff
@@ -39,98 +142,119 @@ def resolve_type_id(type_id: int) -> str:
...
+    "enum",
+    "enum[]",
+    "HoiResult",
+  }
+)
+
+_HOI_PARAM_DIRECTION: tuple[str, ...] = ("In", "Out", "InOut", "RetVal")
+
+
+def _build_introspection_maps() -> tuple[dict[int, str], set[int], set[int], set[int], set[int]]:
```
rickwierenga: this type is very complex and might need a data class

pylabrobot/hamilton/tcp/introspection.py
```diff
@@ -192,6 +302,347 @@ class ObjectInfo:
...
+    return "\n".join(self.format_lines())
+
+
+@dataclass
+class FirmwareTree:
+  """Structured firmware tree produced by introspection traversal."""
+
+  roots: List[FirmwareTreeNode] = field(default_factory=list)
+
+  def format(self) -> str:
```
rickwierenga: are there really multiple tree roots? that is super cursed

pylabrobot/hamilton/tcp/introspection.py
```diff
@@ -39,98 +142,119 @@ def resolve_type_id(type_id: int) -> str:
...
+
+  dotnet_name: str
+  display_name: str
+  ids: tuple[int, int, int, int]  # [In, Out, InOut, RetVal]
+
+
+_HOI_TYPE_ROWS: tuple[_HoiTypeRow, ...] = (
+  _HoiTypeRow("i8", "i8", (1, 17, 9, 25)),
+  _HoiTypeRow("i16", "i16", (3, 19, 11, 27)),
+  _HoiTypeRow("i32", "i32", (5, 21, 13, 29)),
```
rickwierenga: what are these ints? I thought `HamiltonDataType` but for example 17 doens't exist

pylabrobot/hamilton/tcp/introspection.py
```diff
@@ -1,29 +1,132 @@
...
+    self,
+    command: TCPCommand,
+    *,
+    ensure_connection: bool = True,
+    return_raw: bool = False,
+    raise_on_error: bool = True,
+    read_timeout: Optional[float] = None,
+  ) -> Any: ...
+
+  async def resolve_path(self, path: str) -> Address: ...
```
rickwierenga: why does this exist? I can only see it changing in testing, but that might be better done using a mock. I don't think in "normal" operation it would change what the "backend" is? that could really simplify things

pylabrobot/hamilton/tcp/introspection.py
```diff
@@ -280,35 +870,173 @@ class EnumInfo:
...
+# minute(U8), second(U8), millisecond(U16). Wire format confirmed via
+# GetDeckCalibration on PipettorCalibration.
+_NETWORK_STRUCTS[3] = StructInfo(
+  struct_id=3,
+  name="DateTime",
+  fields={
+    "year": ParameterType(type_id=5),  # U16
+    "month": ParameterType(type_id=4),  # U8 (padded)
+    "day": ParameterType(type_id=4),
+    "hour": ParameterType(type_id=4),
```
rickwierenga: this problem exists in a lot of places, just commenting here, but ints are really confusing for type ids because it's not readable. why not just use HamiltonDataType.UI16 for example?

rickwierenga (commented):
there is a lot to learn here so just a first pass

pylabrobot/hamilton/tcp/introspection.py
cmoscy: 100% valid, addressed here, but will do a more comprehensive search and apply

pylabrobot/hamilton/tcp/introspection.py
cmoscy: 1. This works as a standalone interface to arbitrary TCP-based Hamilton instruments that may or may not have the instrument specific backend implemented yet. The MPE2 for example from last meeting. Or when I first connected to the prep.
2. Just a centralized place to hold tcp logic, gets wrapped into a dependency injection (.driver) once there is a concrete backend for a specific instrument. 

So not just for testing, otherwise I'd agree a mock it 100% a better design. 

Given this, do you still think a consolidation/mock is the better pattern?

pylabrobot/hamilton/tcp/introspection.py
cmoscy: The introspection type id system is separate from the wire for interface 0 methods. And type ids for introspection actually encode data type AND directionality. IN/OUT/INOUT/ReturnValue. 

I hate it but is part of the introspection-specific typing system

pylabrobot/hamilton/tcp/introspection.py
cmoscy: Mistake on my part, addressed in changes

pylabrobot/hamilton/tcp/introspection.py
cmoscy: Agreed, addressed in changes

pylabrobot/hamilton/tcp/messages.py
cmoscy: Regressed as part of a larger refactor that reduced duplication, but totally agree. Re-implemented in changes.

pylabrobot/hamilton/tcp/introspection.py
rickwierenga: Yes, I understood that part and it's obviously the right choice to share an object for all hamilton tcp stuff.

If the firmware interfaces are as similar as I understand, shouldn't it be possible to have a lower level "driver" own the io object and just expose these `send_command` command and get_root_object_addresses etc. live on there? And then the `NimbusDriver` just owns that object, and the `MPEDriver` will have an instance of it as well.

pylabrobot/hamilton/tcp/introspection.py
rickwierenga: So if that is the case, then wouldn't it be possible to delete `FirmwareTree` and just keep `FirmwareTreeNode`?

pylabrobot/hamilton/tcp/introspection.py
rickwierenga: sweet

pylabrobot/hamilton/tcp/introspection.py
cmoscy: Ah, actually my mistake, this can be refactored out as you're describing. Been staring at it so long it sort of merged into HamiltonTCPClient in my mind. Doh

rickwierenga: love that name! let's have HamiltonTCPClient, and then NimbusDriver.client and Other.client in the future. And this can be nicely mocked out for testing.

pylabrobot/hamilton/tcp/client.py
cmoscy: Will trace through this logic carefully, as the introspection interface originally needed this (as did error resolution) when I had the partial type / incomplete communication layer and needed fallbacks for things that couldn't be 100% resolved. It could probably be simplified now, but will get back to you on that

pylabrobot/hamilton/tcp/messages.py
rickwierenga: nice!

pylabrobot/hamilton/tcp/introspection.py
rickwierenga: I see, that's odd.

Is there a structure to this? would be more readable than the list of ints 😂

pylabrobot/hamilton/liquid_handlers/nimbus/driver.py
cmoscy: Agreed, addressed in changes. Now baked in to each respective backend

pylabrobot/hamilton/tcp/introspection.py
cmoscy: I still had some validation/old reverse engineering code in here. Have cleaned that up and removed this in the changes.

pylabrobot/hamilton/tcp/introspection.py
cmoscy: You're suggesting we just build the interface tree using FirmwareTreeNode as the building block? 

If I understand correctly that totally agree, much simpler with current behavior

rickwierenga: yes, if it's really a tree then all nodes should be equal or at least be able to use the same class

pylabrobot/hamilton/tcp/introspection.py
rickwierenga: wondering if we can get rid of ParameterType entirely and just use HamiltonDataType. Is it correct that ParameterType just provides some more information about the type? If so we can just store/compute that on the type directly

pylabrobot/hamilton/tcp/introspection.py
cmoscy: From the raw introspection responses, it's all ints. But open to anything that you'd think is an improvement here.

pylabrobot/hamilton/tcp/introspection.py
cmoscy: Got it, will simplify that

pylabrobot/hamilton/tcp/introspection.py
rickwierenga: is this loaded from the machine or defined somewhere? do we know what these numbers map to?

pylabrobot/hamilton/tcp/introspection.py
cmoscy: Will take a look! Probably a good idea.

PR #1000, opened 2026-04-18
v1b1 changes

rickwierenga:
https://discuss.pylabrobot.org/t/updating-plr-api-for-machine-interfaces-discussion/445


pylabrobot/hamilton/liquid_handlers/star/pip_channel.py
```diff
@@ -0,0 +1,1378 @@
...
+  # -- C0:RA  request X position ------------------------------------------------
+
+  async def request_x_pos(self) -> float:
+    """Request current X-position of this channel (mm).
+
+    All PIP channels share the same X arm, so this returns the arm position.
+    """
+    resp = await self.driver.send_command(
+      module="C0",
+      command="RA",
```
BioCam: this is incorrect on multiple levels:

1. `C0 RA` is an EEPROM read command (returns calibration constants set during service);
it doesn't query live motor position, which is what's needed here.

3. `pn` is not a valid field for `C0 RA` - per spec it accepts only `id` and `ra=<param>`.
The firmware silently ignores `pn`, and without `ra=` it echoes the last-queried parameter
(which I verified: after `ra=kg` it returns `kg328`; after `ra=kf` it returns `kf3682`).
So this method silently leaks whatever EEPROM constant some prior unrelated RA call touched, rather than returning a position.
Independently, `fmt="ra#####"` cannot match any real RA reply - responses always lead with the parameter name (e.g. `kg<value>`), never literal `ra`.

5. based on PLR:main this command should be `C0 RX`

BioCam: fixed with #1012

rickwierenga: thanks!

PR #1011, opened 2026-04-26
Expose `iSWAPBackend.rotation_drive_request_position()`

BioCam:
It is time to re-start "Epic: Tame the iSWAP" ... a series of tasks/PRs aimed at making this unpredictable beast more reliable, predictable and ready for our agents.

This is PR 0 of many to come... 

Adds `iSWAPBackend.rotation_drive_request_position()` - returns the iSWAP rotation drive position as a `Coordinate` in deck space - plus the supporting `rotation_drive_request_x_offset()` read of EEPROM `kg` via `C0:RA`.
Also renames a few drive methods so naming reflects what they actually return.

## Problem

The iSWAP arm is a 2-link planar manipulator (L1 = L2 = 138 mm) whose base — the **rotation drive** — sits at a fixed offset from the X-arm carriage.
Every kinematic calculation on the iSWAP — forward kinematics from joint angles to gripper pose, inverse kinematics from a target plate pose to joint angles, reachable-workspace and collision checks — starts from that base point in deck coordinates.
Without it, there is no anchor to put the arm geometry on the deck.

Today the rotation drive's deck position has to be reconstructed by hand from three separate reads:

1. X-arm carriage center,
2. the per-machine X offset stored as EEPROM `kg` (set during service, currently exposed nowhere in PLR),
3. rotation-drive Y from `request_gripper_y_position`, and the 13 mm vertical drop from the rotation-drive plane to the gripper finger plane to recover Z.

Screenshot 2026-04-26 at 20 07 38

This is the foundational primitive for moving iSWAP path planning out of C0's firmware macros (`PG`, `PP`) and into PLR — the larger "tame the iSWAP" goal.
Until joint 0 is queryable as a `Coordinate`, none of the downstream kinematic work can land cleanly.

Separately, the existing `request_{rotation,wrist}_drive_position_increments` methods return an **angle** in motor increments, not a spatial position — the name was misleading and now actively collides with the new `rotation_drive_request_position()`, which really does return a `Coordinate`.

## PR content / solution

**New on `iSWAPBackend`:**
- `rotation_drive_request_x_offset() -> float` — reads EEPROM `kg` via `C0:RA` (mm, /10). Per-machine, set at service.
- `rotation_drive_z_offset_above_finger = 13.0` — class constant for the rotation-drive-to-finger vertical offset.
- `rotation_drive_request_position() -> Coordinate` — composes X-arm center − `kg`, rotation-drive Y, finger Z + 13 mm. 
Raises `RuntimeError` if no iSWAP is installed.

**Renames (angle ≠ position):**
- `request_rotation_drive_position_increments` → `rotation_drive_request_angle_increments`
- `request_rotation_drive_orientation`         → `rotation_drive_request_orientation`
- `request_wrist_drive_position_increments`    → `request_wrist_drive_angle_increments`
- `rotate_rotation_drive`                      → `rotation_drive_rotate`

**Legacy backend (`legacy/.../STAR_backend.py`):**
- 4 deprecation shim bodies updated to delegate to the new method names.
- Existing `"""Deprecated: use ..."""` docstrings left untouched (no new deprecation messaging introduced).

### DIFF
--- a/pylabrobot/hamilton/liquid_handlers/star/iswap.py
+++ b/pylabrobot/hamilton/liquid_handlers/star/iswap.py
@@ -49,6 +49,7 @@ class iSWAPBackend(OrientableGripperArmBackend):
     self.traversal_height = traversal_height
     self._version: Optional[str] = None
     self._parked: Optional[bool] = None
+    self._rotation_drive_x_offset: Optional[float] = None  # cached in _on_setup
 
   @property
   def version(self) -> str:
@@ -95,6 +96,9 @@ class iSWAPBackend(OrientableGripperArmBackend):
 
     if self._version is None:
       self._version = await self._request_version()
+
+    if self._rotation_drive_x_offset is None:
+      self._rotation_drive_x_offset = await self._rotation_drive_request_x_offset()
 
   async def _request_version(self) -> str:
     """Request the iSWAP firmware version from the device."""
@@ -298,7 +302,7 @@ class iSWAPBackend(OrientableGripperArmBackend):
       f"Expected one of {list(wrist_orientation_to_motor_increment_dict)}."
     )
 
-  async def rotation_drive_request_x_offset(self) -> float:
+  async def _rotation_drive_request_x_offset(self) -> float:
     """Read the X-offset i.e. X-axis center <-> iSWAP rotation drive, in mm.
 
     Stored in the master EEPROM as parameter ``kg`` (set via ``C0:AG`` —
@@ -306,6 +310,7 @@ class iSWAPBackend(OrientableGripperArmBackend):
     Previously measured to be 32.8 mm by contributor;
     per-machine calibrated during service. Required for deriving the
     iSWAP rotation drive's deck X coordinate from the X-arm carriage center.
+    Cached on the backend as ``_rotation_drive_x_offset`` during setup.
     """
     resp = await self.driver.send_command(module="C0", command="RA", ra="kg", fmt="kg###")
     return cast(int, resp["kg"]) / 10.0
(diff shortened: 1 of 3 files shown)


pylabrobot/hamilton/liquid_handlers/star/iswap.py
```diff
@@ -287,17 +287,49 @@ async def request_wrist_drive_orientation(self) -> "iSWAPBackend.WristDriveOrien
...
+    x_arm_center = await self.driver.left_x_arm.request_position()  # type: ignore[union-attr]
+    iswap_x_offset = await self.rotation_drive_request_x_offset()
+    rotation_drive_y = await self.rotation_drive_request_y()
+    finger_loc = (await self.request_gripper_location()).location
+
+    return Coordinate(
+      x=x_arm_center - iswap_x_offset,
+      y=rotation_drive_y,
+      z=finger_loc.z + self.rotation_drive_z_offset_above_finger,
+    )
```
rickwierenga, line 336: I think it is worth considering making `rotation_drive_request_position` private, just loading it in `_on_setup` and then storing it locally so we don't have to call this firmware command every time.

BioCam: I'm not sure I understand what you mean:

Every time the STAR's X-arm moves this position is different, same for when the iSWAP performs any action, or channels require the iSWAP to move out of the way?

rickwierenga: the offset should be constant right?

BioCam: ohhh you don't mean making `rotation_drive_request_position` private,
but instead you mean making `rotation_drive_request_x_offset` private and cached during setup?

yes, that makes sense - will implement

rickwierenga: yes sorry

rickwierenga:
can I or you merge this?

rickwierenga:
added to v0 branch in #1014, this was targeting v1. we will add it to v1 when syncing v0 into v1

PR #1014, opened 2026-04-27
Expose `STARBackend.iswap_rotation_drive_request_position()`

BioCam:
Another one for the "Tame the iSWAP" (Part 3)

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -1367,6 +1367,7 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
     self._extended_conf: Optional[ExtendedConfiguration] = None
     self._channel_traversal_height: float = 245.0
     self._iswap_traversal_height: float = 280.0
+    self._iswap_rotation_drive_x_offset_mm: Optional[float] = None
     self.core_adjustment = Coordinate.zero()
     self._unsafe = UnSafe(self)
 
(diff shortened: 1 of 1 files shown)


pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -10026,85 +10032,179 @@ async def iswap_put_plate(
...
+    # explicitly when the drive sits at the parking stop rather than the W3
+    # work stop.
+    # TODO: add PARKED_LEFT reference for STAR(let)s that park on the left.
+    rotation_reference_positions = {
+      STARBackend.RotationDriveOrientation.LEFT: -29068,
+      STARBackend.RotationDriveOrientation.FRONT: 0,
+      STARBackend.RotationDriveOrientation.RIGHT: 29068,
+      STARBackend.RotationDriveOrientation.PARKED_RIGHT: 29500,
     }
+    tolerance_incr = 1700  # ~5 deg at 0.00310 deg/incr (iSWAP W-drive resolution)
```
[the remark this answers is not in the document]
BioCam: Will be addressed in a follow up PR (already designed but out of scope here)

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -1763,6 +1764,11 @@ async def set_up_iswap():
           minimum_traverse_height_at_beginning_of_a_command=int(self._iswap_traversal_height * 10)
         )
 
+        if self._iswap_rotation_drive_x_offset_mm is None:
+          self._iswap_rotation_drive_x_offset_mm = (
+            await self._iswap_rotation_drive_request_x_offset()
+          )
+
```
rickwierenga, line 1768: isn't it always none?

BioCam: Yes, because during class instantiation it has been set to `None` - then during iswap_setup information is retrieved and populated.

same Pythonic pattern as: 
```python
self._iswap_parked: Optional[bool] = None
self._num_channels: Optional[int] = None
self._core_parked: Optional[bool] = None
self._extended_conf: Optional[ExtendedConfiguration] = None
...
```

rickwierenga: yes, so then we always want to set it on setup (unconditionally)

PR #1018, opened 2026-04-29
Expose X-axis movement with acceleration control for X-arm and iSWAP rotation drive

BioCam:
Tame the iSWAP - Part 6

Hi everyone,

The standard `C0`-based X-arm movement along the X-axis does not allow any form of speed nor acceleration control.

This is an obstacle e.g. for...
- traversing tips with slippery liquids along the X dimension
- generating smoother motion for the iSWAP

This PR exposes per-call acceleration control by going one layer down to the `X0` module, which provides absolute position commands with parametric acceleration (`XP`, with acceleration level `lr` and current protection `lw`).

new methods:

```python
STARBackend.x_arm_move(x, acceleration_level=3, current_protection_limiter=7)

STARBackend.iswap_rotation_drive_move_x(x, acceleration_level=3, current_protection_limiter=7)

STARBackend.x_arm_request_firmware_version()
```

Note:

I apparently have a very old X-arm.
Ideally we would use an X-arm movement command that actually enables precise speed and acceleration parameters.
But my firmware `X0RFid0143rf1.4S 2012-04-25` does not enable this.

I have made `STARBackend.x_arm_request_firmware_version()` to enable a user with more modern X-arm firwmare/hardware to make the `STARBackend.x_arm_move()` more powerful and adapt it's allowed argument input dynamically to the available firmware (as we have done for other commands, e.g. 96-head move y)

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -1816,44 +1816,41 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
   # X-Arm
   # -----------------------------------------------------------------------
 
-  async def x_arm_request_firmware_version(self):
-    """"""
-    
-    resp = await self.send_command(
-        module="X0",
-        command="RF",
-    )
-    fw_datetime = self._parse_firmware_version_datetime(resp)
+  async def x_arm_request_firmware_version(self) -> Tuple[str, datetime.date]:
+    """Request the X-arm firmware version and build date.
 
-    return resp.split("rf")[-1].split(" ")[0], fw_datetime
+    Returns:
+      A tuple of (version_string, build_date), e.g. ("1.0S", date(2009, 6, 24)).
+    """
+
+    resp = await self.send_command(module="X0", command="RF")
+    version = resp.split("rf")[-1].split(" ")[0]
+    build_date = self._parse_firmware_version_datetime(resp)
+    return version, build_date
 
   async def x_arm_move(
-    self,                                                                       
+    self,
     x: float,
-    acceleration_level: int = 3,                                                
+    acceleration_level: int = 3,
     current_protection_limiter: int = 7,
-    ):
+  ):
     """Move the X-arm to an absolute X position.
-                                                                                
+
     Args:
-      x: Target X coordinate in mm. Must be between 90.0 and 1350.0.             
-      acceleration_level: Acceleration index (hardware units). Must be an       
-        integer between 1 and 5 (inclusive). Default: 3.                        
-      current_protection_limiter: Motor current limit (0-7, hardware units).    
-        Default: 7.                                                             
-    """           
-                                                                                
-    if 90.0 <= x >= 1350.0:
-        raise ValueError(f"x must be between 90.0 and 1350.0 mm; got {x}")
-    if isinstance(acceleration_level, int) and 1 <= acceleration_level > 5:
-        raise ValueError(                                                                            
-              f"acceleration_level must be an integer between 1 and 5; got {acceleration_level}"                                                         
-            )             
-    if isinstance(current_protection_limiter, int) and 0 <= current_protection_limiter > 7:
-        raise ValueError(                                            
-      f"current_protection_limiter must be an integer between 0 and 7; got  {current_protection_limiter}"                                                 
-    )             
-                                                                                
+      x: Target X coordinate in mm. Must be between 90.0 and 1350.0.
+      acceleration_level: Acceleration index (hardware units), 1-5. Default 3.
+      current_protection_limiter: Motor current limit (hardware units), 0-7. Default 7.
+    """
+
+    if not (90.0 <= x <= 1350.0):
+      raise ValueError(f"x must be between 90.0 and 1350.0 mm, is {x}")
+    if not (1 <= acceleration_level <= 5):
+      raise ValueError(f"acceleration_level must be between 1 and 5, is {acceleration_level}")
+    if not (0 <= current_protection_limiter <= 7):
+      raise ValueError(
+        f"current_protection_limiter must be between 0 and 7, is {current_protection_limiter}"
+      )
+
     return await self.send_command(
       module="X0",
       command="XP",
@@ -1861,7 +1858,6 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
       lr=str(acceleration_level),
       lw=str(current_protection_limiter),
     )
-
 
   # # # # Single-Channel Pipette Commands # # # #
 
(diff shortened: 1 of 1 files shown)


pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -1812,6 +1812,53 @@ def setup_done(self) -> bool:
...
+      raise ValueError(
+        f"current_protection_limiter must be between 0 and 7, is {current_protection_limiter}"
+      )
+
+    return await self.send_command(
+      module="X0",
+      command="XP",
+      la=f"{round(x * 10):05}",
+      lr=str(acceleration_level),
+      lw=str(current_protection_limiter),
```
rickwierenga: doesn't this depend on the firmware version?

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -9845,6 +9892,39 @@ async def iswap_rotation_drive_request_position(self) -> Coordinate:
...
+    current_protection_limiter: int = 7,
+  ):
+    """Move the iSWAP rotation drive to an absolute X position (deck coordinates).
+
+    Thin wrapper around `x_arm_move` that translates rotation-drive X into
+    X-arm carriage X using the cached `kg` offset.
+
+    Args:
+      x: Target rotation-drive X coordinate in mm.
+      acceleration_level: Acceleration index (hardware units), 1-5. Default 3.
```
rickwierenga, line 9908: can we put the acceleration index table in the doctoring?

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: `X0XP` doesn't to my knowledge - but as mentioned in the description, I can't verify on device for more modern firmware

rickwierenga:
also does this replace position_left_x_arm_? I do not see why we would keep both

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: I don't understand, what do you mean by this?

BioCam:
> also does this replace position_left_x_arm_? I do not see why we would keep both

It does, but this is a new command and therefore untested outside of one lab.

`position_left_x_arm_` is the foundation of `move_channel_x` which (at least for me) is used in every single script I have ever written.

Replacing it instead of adding next to it first, testing and verifying it, and then replacing once verified across labs seemed too risky to me.

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
rickwierenga: table `4.2 X drive accelerate values`

```
+---------------------+-------------+---------------------------+-------------------------+-------------------------+
| acceleration index  | start value | acceleration increase per | top limit of accel. inc.| top limit of accel. inc.|
|                     |             | 1 meter (reduced by 100)  | (low weight / default)  | (high weight)           |
+---------------------+-------------+---------------------------+-------------------------+-------------------------+
| 0 (init. move only) | 100         | 1                         | 100                     | 100                     |
| 1                   | 300         | 2                         | 500                     | 375                     |
| 2                   | 400         | 6                         | 1000                    | 750                     |
| 3                   | 500         | 10                        | 1500                    | 1125                    |
| 4                   | 600         | 14                        | 2000                    | 1500                    |
| 5                   | 700         | 18                        | 2500                    | 1875                    |
+---------------------+-------------+---------------------------+-------------------------+-------------------------+
```

rickwierenga: otherwise it is not clear what `acceleration_level` means

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -9845,6 +9892,39 @@ async def iswap_rotation_drive_request_position(self) -> Coordinate:
...
+    x_max = 1350.0 - kg
+    if not (x_min <= x <= x_max):
+      raise ValueError(f"x must be between {x_min} and {x_max} mm, is {x}")
+
+    return await self.x_arm_move(
+      x=x + kg,
+      acceleration_level=acceleration_level,
+      current_protection_limiter=current_protection_limiter,
+    )
+
```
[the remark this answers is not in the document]
BioCam: this generates x_arm errors which we want to avoid - and instead have clean informative errors for the rotation drive/iSWAP SCARA base

rickwierenga:
either it works or it doesn't. has this new one been tested and verified?

BioCam:
> either it works or it doesn't. has this new one been tested and verified?

yes it has been verified on one device - do you know it will work on newer ones?

rickwierenga:
I am testing it

in the meantime, it also ignores the `left_side_panel_installed` which constrains the range of x movement. Ideally we add a `min_x` function that returns the value rather than it living in every individual function

rickwierenga:
tested on STAR with 

-  X-arm firmware: version='1.4S' build_date=2012-04-25 

2 STARLets

- 1.4S, build date 2012-04-25
- 1.4S, build date 2012-04-25

and it works nicely, has parity with position_left_x_arm_

i cannot install 5.1 firmware on any of the star(let)s i have tried

```python
import asyncio
from pylabrobot.liquid_handling.backends.hamilton.STAR_backend import STARBackend
from pylabrobot.resources.hamilton import STARDeck


async def main():
  star = STARBackend()
  lh = LiquidHandler(backend=star, deck=STARDeck())
  await lh.setup(
    skip_instrument_initialization=True,
    skip_pip=True,
    skip_autoload=True,
    skip_iswap=True,
    skip_core96_head=True,
  )

  try:
    version, build_date = await star.x_arm_request_firmware_version()
    print(f"X-arm firmware: version={version!r} build_date={build_date}")

    for x_mm in (200.0, 600.0, 400.0):
      print(f"\n--- target {x_mm} mm ---")

      await star.position_left_x_arm_(x_position=int(round(x_mm * 10)))
      pos_legacy = await star.request_left_x_arm_position()
      print(f"after position_left_x_arm_({int(x_mm*10)}): RX = {pos_legacy} mm")

      await star.position_left_x_arm_(x_position=int(round((x_mm + 100) * 10)))

      await star.x_arm_move(x=x_mm)
      pos_new = await star.request_left_x_arm_position()
      print(f"after x_arm_move({x_mm}): RX = {pos_new} mm")

      delta = abs(pos_legacy - pos_new)
      print(f"delta = {delta} mm")
  finally:
    await lh.stop()


asyncio.run(main())
```

rickwierenga:
do we have a doc for 5.1? what does it say for this command?

rickwierenga:
i can change the acceleration precisely by overwriting the `ax` parameter using command `X0AA` (analogous to `slow_iswap`)

base:

```python
>>> import time
>>> await backend.x_arm_move(x=100)
>>> t0 = time.time()
>>> await backend.x_arm_move(x=500)
>>> t1 = time.time()
>>> print(t1 - t0)
1.865234375
```

to update acceleration: first request the current value

```python
>>> await backend.send_command("X0", "RA", ra="ax")
'X0RAid0063ax100 01 0100 300 02 0375 400 06 0750 500 10 1125 600 14 1500 700 18 1875'
```

Then you can overwrite it

```python
ax = "100 01 0100 300 02 0375 400 06 0750 500 18 1125 600 14 1500 700 18 1875"
await backend.send_command("X0", "AA", ax=ax)
```

Here I changed 10 -> 18 (after 500) for the "acceleration increase per 1 meter (reduced by 100" (https://github.com/PyLabRobot/pylabrobot/pull/1018#discussion_r3163635129). I chose the third because the default acceleration_level for `x_arm_move` is 3.

this makes the arm move faster!

```python
>>> import time
>>> await backend.x_arm_move(x=100)
>>> t0 = time.time()
>>> await backend.x_arm_move(x=500)
>>> t1 = time.time()
>>> print(t1 - t0)
1.6652910709381104
```

i do not see parameters for velocity directly though... You can however control the PID constants (`kp`,`ki` and `kd`) which i havent tried yet

rickwierenga:
6 blocks of "start value (offset)
acceleration increase
top limit of acceleration increase"

BioCam:
@rickwierenga: I don't understand

> Here I changed 10 -> 18 (after 500) 

Isn't both the request and set 18 here?

Also, what does 18 refer to here, i.e. what unit does it take?

rickwierenga:
fixed, accidentally pasted the same string twice

the unit given is "acceleration increase per 1 meter (reduced by 100 for long integer arithmetic)", so jerk. I do not know more than that. Looking at the other lines in the firmware doc (like XF command) I see 150mm/s^2, so I am assuming it is that since the numbers are about the same size.

BioCam:
Do you think we should finish this PR as is and figure out what these units mean and how to best use them in an upgrade PR?

rickwierenga:
let's name it `experimental_` until we at least figure out the acceleration ramps. when it is no longer experimental we can replace position_left_x_arm_ with this

rickwierenga:
and then merge as experimental yeah

BioCam:
Happy to - let's call them experimental_ for now and figure out what further control we can achieve

BioCam:
fundamentally the x move only takes 5 "levels"

but you are saying we can change what these 5 EEPROM registers map to:

i.e. we should definitely bound to the set 1 level for lower bound and 5 level for upper bound?

but I am still not sure what these setting actually mean in terms of physical units

rickwierenga:
> fundamentally the x move only takes 5 "levels"

it takes one of 5 levels, which just points to the saved list (see table). each list item is 3 values (which I think are start accel, jerk, stop accel)

AA allows us to override that list with custom values. it probably resets on power cycle but persists between commands, so reseting them is necessary (similar to `slow_iswap`)

so index 1 through 5 really just points to one of 3 values

the best would be for the user to specify the value rather than choosing from one of 5

> i.e. we should definitely bound to the set 1 level for lower bound and 5 level for upper bound?

probably, if 1 and 5 are defined as the true min and max but I can imagine we can safely go beyond those

> but I am still not sure what these setting actually mean in terms of physical units

I am not sure either

PR #1020, opened 2026-04-30
Expose `STARBackend.iswap_rotation_drive_move_y()`

BioCam:
### Tame the iSWAP - Part 7

The center of the rotation drive is the "base of the SCARA part of the iSWAP".

In the previous parts I've exposed requesting the position information of it, then added it's movement in X.

In this PR, I'm adding its movement in Y.

Because the iSWAP is not alone on the rail / its X-coordinate, this is the PR in which I am adding the **iSWAP safety features in the Y dimension**:

-> from now on you can no longer crash your iSWAP into ...

- the back of the arm
- channel_0

Because it is tedious to always think about arm positions I added `make_space` argument to:

```python
  async def iswap_rotation_drive_move_y(
    self,
    y: float,
    speed: float = 220.0,
    acceleration_level: int = 2,
    current_protection_limiter: int = 7,
    make_space: bool = False,
  )
```
...adapted from `position_channels_in_y_direction`'s `make_space`

To make this calculation possible I took the iSWAP channel diameter to be `30.5` mm, as described in the operator manual - see #1011 

---

Please let me know whether you think we should make `make_space: bool = False` by default; it might help people but does introduce implicit behaviour - though expected behaviour?

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -9871,6 +9871,64 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
     iswap_y_pos = resp["ry"][1]  # 0 = FW counter, 1 = HW counter
     return round(STARBackend.y_drive_increment_to_mm(iswap_y_pos), 1)
 
+  # Vertical drop from the iSWAP rotation drive plane to the gripper finger
+  # plane. R0 RZ is calibrated to the finger plane; the rotation drive sits
+  # 13 mm above it.
+  iswap_rotation_drive_z_offset_above_finger_mm = 13.0
+
+  async def iswap_rotation_drive_request_z(self) -> float:
+    """Request iSWAP rotation drive Z position (deck coordinates), in mm.
+
+    Adds the 13 mm structural offset to the gripper finger plane (C0 QG).
+    """
+    if not self.extended_conf.left_x_drive.iswap_installed:
+      raise RuntimeError("iSWAP is not installed")
+    finger_plane_z = (await self.request_iswap_position()).z
+    return finger_plane_z + STARBackend.iswap_rotation_drive_z_offset_above_finger_mm
+
+  async def iswap_rotation_drive_request_position(self) -> Coordinate:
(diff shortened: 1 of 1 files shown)


pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -9927,6 +9929,85 @@ async def experimental_iswap_rotation_drive_move_x(
...
+      acceleration_level: Acceleration index, 1 or 2. Default 2.
+      current_protection_limiter: Motor current limit, 0-7. Default 7.
+      make_space: If True, reposition pipetting channels in a single
+        synchronous JY move when channel 0 is in the way and can be cleared.
+        If False, raise so the caller decides.
+    """
+    if not self.extended_conf.left_x_drive.iswap_installed:
+      raise RuntimeError("iSWAP is not installed")
+
+    iswap_radius = STARBackend.iswap_rotation_drive_diameter / 2
```
rickwierenga: `rotation_drive_radius`?

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -9927,6 +9929,85 @@ async def experimental_iswap_rotation_drive_move_x(
...
+      raise ValueError(
+        f"speed must be between "
+        f"{STARBackend.y_drive_increment_to_mm(speed_min)} and "
+        f"{STARBackend.y_drive_increment_to_mm(speed_max)} mm/sec, "
+        f"got {speed} mm/sec"
+      )
+
+    if not (1 <= acceleration_level <= 2):
+      raise ValueError(f"acceleration_level must be between 1 and 2, got {acceleration_level}")
+
```
rickwierenga, line 10005: this is 6250 steps / s2 or 2 * 6250

so 6250 * 0.046302082 = 289.388 mm/s or twice that?

BioCam: For level 1:
6250 * 0.046302082 = 289.388 mm/s

For level 2 twice that

Is how i understood it

rickwierenga: cool

PR #1021, opened 2026-04-30
Expose iswap rotation drive move z

rickwierenga:
follow up to #1018 and #1020 

@BioCam


rickwierenga:
stacked on #1020

BioCam:
### Tame the iSWAP - Part 8

thank you @rickwierenga !

BioCam:
Have you checked whether is iswap gripper z-height + 13?

rickwierenga:
is that something we defined or is that native to the firmware? what's the best way to check? here i am using the firmware frame for z

BioCam:
> is that something we defined or is that native to the firmware? what's the best way to check? here i am using the firmware frame for z

The easiest way to check is to use this move z command, and the check the position of the iSWAP - currently still set to the gripper finger center

BioCam:
We know this distance from the operator manual, se #1011

BioCam:
Cross-verified on our STAR, command works as expected:

Screenshot 2026-05-01 at 13 28 44

PR #1022, opened 2026-04-30
Expose iSWAP rotation drive request/set angle

rickwierenga:
## Summary
- Adds `STARBackend.iswap_rotation_drive_request_angle()` returning the current rotation drive angle in degrees (signed, 0 deg = FRONT).
- Adds `STARBackend.iswap_rotation_drive_set_angle(angle, speed, acceleration, current_limit)` to rotate Joint 1 to an arbitrary angle within ~±93°. The wrist drive is held at its current position.

Branched off #1021; will rebase onto main once that lands.

## Test plan
- [ ] On a STAR with iSWAP installed, call `iswap_rotation_drive_set_angle(0)`, `45`, `-45` and verify the drive moves to the expected orientation.
- [ ] Confirm `iswap_rotation_drive_request_angle()` reports the same angle that was just set (within ~0.01°).
- [ ] Verify out-of-range angles (e.g. 100) raise `ValueError`.

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -10253,6 +10253,11 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
     response = await self.send_command(module="R0", command="RW", fmt="rw######")
     return cast(int, response["rw"])
 
+  async def iswap_rotation_drive_request_angle(self) -> float:
+    """Query the iSWAP rotation drive angle in degrees (signed, 0 deg = FRONT)."""
+    increments = await self.request_iswap_rotation_drive_position_increments()
+    return increments * STARBackend.iswap_rotation_drive_deg_per_increment
+
   async def request_iswap_rotation_drive_orientation(self) -> "RotationDriveOrientation":
     """Request the iSWAP rotation drive orientation.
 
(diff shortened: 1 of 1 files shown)


pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -10253,6 +10253,11 @@ async def request_iswap_rotation_drive_position_increments(self) -> int:
     response = await self.send_command(module="R0", command="RW", fmt="rw######")
     return cast(int, response["rw"])
 
+  async def iswap_rotation_drive_request_angle(self) -> float:
+    """Query the iSWAP rotation drive angle in degrees (signed, 0 deg = FRONT)."""
+    increments = await self.request_iswap_rotation_drive_position_increments()
+    return increments * STARBackend.iswap_rotation_drive_deg_per_increment
```
BioCam: This is not always true:

FRONT is indeed always perfectly front-facing, i.e. the linkage_1/rotation drive is perfectly aligned to the back of the X-arm
-> this is what matters because it is one of the components that defines the relative Orientation, specifically yaw (rotation about the z-axis), of the gripper compared to the carriers and plates it will be picking up from.

But FRONT is not always 0 increments in the rotation drive encoder!

Instead the firmware-stored increment value for FRONT is set during the calibration process - my STAR has a FRONT value of 156 increments.

Using always 0 increments here will result in incorrect gripper orientation.

Instead I recommend we set 0 degrees, -90 and +90 degrees based on the firmware-predefined positions - i.e. build a dynamic, firmware-corrected degree system at runtime.

BioCam: I wonder now whether we should keep `STARBackend.iswap_rotation_drive_deg_per_increment` at all or just compute its value at runtime? 🤔

BioCam:
#### Tame the iSWAP - Part 10

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -10332,6 +10337,42 @@ async def rotate_iswap_rotation_drive(self, orientation: RotationDriveOrientatio
     else:
       raise ValueError(f"Invalid rotation drive orientation: {orientation}")
 
+  async def iswap_rotation_drive_set_angle(
```
BioCam: `set_` is misleading terminology here - based on our [PLR command prefix terminology](https://discuss.pylabrobot.org/t/standardised-plr-command-prefix-proposal/403/4) this implies changing a memory value.

But this is an action command - maybe `iswap_rotation_drive_rotate_to_angle()` ?

Note: A PLR standard/convention for increment vs angle naming of methods would be useful here)

PR #1024, opened 2026-05-02
`STARBackend`: anchor iSWAP move_y to EEPROM parking pose

BioCam:
#### Tame the iSWAP - Part 10

Replaces `extended_conf.pip_maximal_y_position` (PIP channel envelope, 606.5 mm  default) with the EEPROM-stored iSWAP parking pose (`py["parking"]`, per-machine) as the upper bound in `iswap_rotation_drive_move_y`.

The previous validator rejected legitimate iSWAP positions past the PIP channel reach, blocking parking moves on every machine where parking sits beyond `pip_maximal_y_position`.

Parking value is cached at `set_up_iswap()` alongside `_iswap_rotation_drive_x_offset_mm`.
Lower bound and the `make_space` channel-collision logic below are unchanged.


rickwierenga:
working for me

PR #1026, opened 2026-05-05
Improved ODTC Backend in v1b1

cmoscy:


cmoscy:
More thorough physical testing this week, but wanted to get on here so I don't forget. 

Improved ODTC backend in v1b1 API. Implementation of the `Thermocycling` capability with `Protocol` standard. Includes improvements from #894 discussion.

PR #1029, opened 2026-05-11
Replace Tecan tip geometry prints with warnings

norle:
## Summary

Replaces direct `print()` diagnostics for invalid Tecan tip geometry with `warnings.warn(...)`.

Some Tecan tip factories currently instantiate tips with `total_tip_length <= 0`. Those diagnostics were printed directly to stdout, which makes resource/catalog introspection noisy and hard for downstream apps to suppress.

This change emits a `UserWarning` instead, preserving the diagnostic while making it suppressible/capturable through Python’s warnings system.

## Changes

- Add a warning in `TecanTip.__init__` when `total_tip_length <= 0`
- Remove direct `print()` calls from affected Tecan tip factories
- Add a regression test asserting the warning is emitted without stdout output

## Testing

```bash
pytest pylabrobot/resources/tecan/tip_creators_tests.py


rickwierenga:
thanks!

PR #1031, opened 2026-05-11
fix linting from #1029

norle:
#1029 introduced a linting error, this should fix it. sorry for not checking before.


rickwierenga:
All good thanks for fixing

PR #1033, opened 2026-05-12
Add Molecular Devices SpectraMax Gemini EM backend

brad-usredoxlabs:
## Summary

Adds a Molecular Devices SpectraMax Gemini EM plate reader backend, including exports, backend tests, and user-facing documentation.

## Validation

- `python3 -m unittest pylabrobot.plate_reading.molecular_devices.backend_tests`
- Direct Linux hardware smoke test over `/dev/ttyUSB0`: setup, status, temperature, open, close
- Direct Linux hardware validation over `/dev/ttyUSB0`: fluorescence, luminescence, TRF endpoint, TRF kinetic, emission spectrum, excitation spectrum, fill wellscan, and partial-region fluorescence

## Hardware Results

- Fluorescence: 1 read, 8 x 12
- Luminescence: 1 read, 8 x 12
- TRF endpoint: 1 read, 8 x 12
- TRF kinetic: 2 reads, each 8 x 12
- Emission spectrum: 2 reads, each 8 x 12
- Excitation spectrum: 2 reads, each 8 x 12
- Fill wellscan: 9 reads, each 6 x 6, with scan point/x/y metadata
- Partial fluorescence B2:G7: 1 read, 6 x 6

`nan` values appeared as expected where the reader returned blanks.

## Notes

Hardware validation was run on a Linux laptop directly connected to the SpectraMax Gemini EM through a USB-to-RS-232 adapter. `ruff` was not run locally because `python3 -m ruff` is not installed in this environment.


rickwierenga:
Pushed `0d80c88` addressing four review findings:

1. **State leak between reads** — `read_fluorescence` now calls `_set_wellscan_mode(False)`. Previously it was the only read method that didn't, so a prior `read_fluorescence_wellscan` would leave `WELLSCANMODE=ON` and a subsequent plain fluorescence read would inherit that state.

2. **Silent `focal_height` ignore** — `read_fluorescence` now raises `NotImplementedError` when `focal_height is not None`, matching the existing behavior of `read_luminescence` and `read_fluorescence_wellscan`. Tests that passed `focal_height=0` to `read_fluorescence` have been updated to drop the argument, with a new `test_fluorescence_rejects_focal_height` covering the strict path.

3. **Plate-specific wellscan defaults** — `read_fluorescence_wellscan` now derives `center_x`/`center_y` from `plate.get_item(0).location` (same logic as `_set_plate_region`) when not supplied. The previous defaults (`14.380`, `20.235`) were captured from one dev plate and silently wrong for any other geometry. The OEM-validated sequence test now passes those values explicitly to preserve the captured command stream.

4. **Order-sensitive well comparison** — `_assert_full_plate` and `_get_well_region` now compare by `{id(well) for ...}` set semantics. Previously `wells != plate.get_all_items()` used Python list equality, so a reordered full-plate list would fail the fast-path and either raise `NotImplementedError` (wellscan) or fall through to the rectangular check unnecessarily.

All 69 tests in `backend_tests.py` pass locally.

rickwierenga:
@brad-usredoxlabs really good PR!

pushed some changes and converted the hello world from markdown to notebook so it's easier to run. could you please verify this works?

also renamed the non-standard-PLR methods to `experimental_` because we want to figure out a good call+return type system for them. In this way, it is clear to users this function might change in the future while still being useful today.

rickwierenga:
thanks and congrats on first PR!

PR #1037, opened 2026-05-13
Expose `STARBackend.iswap_wrist_drive_{request,rotate_to}_angle()`

rickwierenga:
## Summary

Mirror of PR #1022 (rotation-drive angle methods), but for the wrist drive (Joint 2). Anchored on the per-machine EEPROM `pt` table so the four named stops round-trip exactly:

| Orientation | Angle |
|---|---|
| `RIGHT` | -90 deg |
| `STRAIGHT` | 0 deg |
| `LEFT` | +90 deg |
| `REVERSE` | +180 deg |

User-facing 0 deg is the calibrated `STRAIGHT` stop (wrist aligned with the arm). The motor's raw zero sits between `STRAIGHT` and `LEFT` and has no physical meaning, so it's not a useful reference.

### API

- `iswap_wrist_drive_request_angle() -> float` — current wrist angle in deg, signed from `STRAIGHT`.
- `iswap_wrist_drive_rotate_to_angle(angle, speed=20_000, acceleration=145, current_limit=5)` — `angle` accepts either a `WristDriveOrientation` member (uses the exact EEPROM increment) or a float in degrees. Rotation drive is held at its current position.

### Implementation notes

- Three piecewise-linear segments: `[RIGHT, STRAIGHT]` → `[-90, 0]`, `[STRAIGHT, LEFT]` → `[0, +90]`, `[LEFT, REVERSE]` → `[+90, +180]`. Angles outside `[-90, +180]` extrapolate using each boundary segment's slope.
- No new firmware queries — `_iswap_wrist_drive_request_predefined_increments()` and the setup-time `_iswap_wrist_drive_predefined_increments` dict already exist (added in #1022).
- Hardware-range validation uses the existing `iswap_wrist_drive_{min,max}_increment` class constants (±30000 incr ≈ ±152 deg motor).

## Test plan

- [ ] Static: existing CI (lint, type-check) passes.
- [ ] On a real STAR with iSWAP installed:
  - [ ] `request_angle()` at each named stop returns exactly -90/0/+90/+180 deg.
  - [ ] `rotate_to_angle(WristDriveOrientation.STRAIGHT)` after a manual nudge returns the wrist to STRAIGHT exactly.
  - [ ] `rotate_to_angle(45.0)` puts the wrist halfway between STRAIGHT and LEFT; `request_angle()` reports ~45.0.
  - [ ] `rotate_to_angle(-135.0)` raises `ValueError` for out-of-range (or extrapolates if within motor limits — verify behavior).
  - [ ] Rotation drive does not move during a wrist-only rotate.

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -10715,11 +10715,15 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
       wrist_position_deg = STARBackend._iswap_wrist_drive_increments_to_angle(
         wrist_position_increments
       )
+      min_deg = STARBackend._iswap_wrist_drive_increments_to_angle(
+        STARBackend.iswap_wrist_drive_min_increment
+      )
+      max_deg = STARBackend._iswap_wrist_drive_increments_to_angle(
+        STARBackend.iswap_wrist_drive_max_increment
+      )
       raise ValueError(
-        f"angle {angle} maps to {wrist_position_increments} incr "
-        f"({wrist_position_deg:.2f} deg), outside hardware range "
-        f"[{STARBackend.iswap_wrist_drive_min_increment}, "
-        f"{STARBackend.iswap_wrist_drive_max_increment}]"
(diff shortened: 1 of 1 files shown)


BioCam:
**Tame the iSWAP - Part 12**

PR #1042, opened 2026-05-15
Resource catalog 3d docs

norle:
Replaced resource catalog with a more interactive list, lmk what you think. I added 3d models to the labware. I'm thinking of making a gui with a more sophisticated simulation/visualization (basically making biolint opensource but less complicated) proabbly in a new repo.

## Summary

This PR replaces the old per-manufacturer resource-library listing with a single Resource Catalog page backed by generated geometry metadata and inline 3D previews.

#### Catalog + 3D infrastructure
- Added catalog index build extension:
  - `docs/_exts/pylabrobot_labware_catalog.py`
- Added frontend assets:
  - `docs/_static/plr_labware_catalog.js`
  - `docs/_static/plr_labware_catalog.css`
  - `docs/_static/plr_geometry_viewer.js`
- Added catalog page:
  - `docs/resources/catalog.md`

#### Geometry generation support
- Added geometry module/tests:
  - `pylabrobot/resources/geometry.py`
  - `pylabrobot/resources/geometry_tests.py`
- Exported geometry utilities:
  - `pylabrobot/resources/__init__.py`
- Updated API docs:
  - `docs/api/pylabrobot.resources.rst`

#### Resource docs restructuring
- Renamed Resource Library wording to Resource Catalog:
  - `docs/resources/index.md`
- Removed old manufacturer-page listing from nav and redirected legacy URLs to catalog:
  - `docs/conf.py`
- Updated notebook links that referenced removed library pages:
  - `docs/resources/carrier/plate-carrier/plate_carrier.ipynb`
  - `docs/user_guide/01_material-handling/heating_shaking/hamilton.ipynb`
- Updated docs workflow/build files as part of catalog pipeline:
  - `.github/workflows/docs.yml`
  - `Makefile`

## Why

The old manufacturer page tree was redundant and harder to navigate. The new catalog centralizes metadata, search/filtering, and 3D previews in one place.

## Testing

- Ran docs build (`make docs`) and resolved broken cross-references from removed library pages by updating links + redirects.
- Ran `make test` locally; run reaches an existing stall point at:
  - `pylabrobot/visualizer/visualizer_tests.py::VisualizerSetupStopTests::test_setup_stop`

## Notes

This branch consists of:
- `35f426db` geometry
- `ea1972e7` remade resource catalog with 3d models
- `712eff09` remove old resource libarry
- `2c214a79` replace resource libary with resource catalog


rickwierenga:
this is sick, we've been wanting to replace the static pages with a searchable/filterable list for a long time

BioCam:
This is amazing! Thank you @norle !!

BioCam:
Hi @norle - thanks for this, this is a great improvmenet!
I've pushed a series of commits onto the branch to build on it:

**3D viewer:** true mm rulers + Blender-colored XYZ axes, a tuned default camera, size readout (X×Y×Z + well/tip count), pan (MMB / Shift-drag), a reset/home button, full ±90° pitch, tick decimation, a chamfered orientation cube with on-surface labels, depth-correct plate-outline vs. wells, well bottom-type rendering (flat/V/U), and a projection/render perf pass (per-frame transform cache + rAF-coalesced renders).

**Catalog:** renamed the "Vendor" filter/field to "Manufacturer" end-to-end, since the value is sourced from each library page's manufacturer title.
Also, quick note: we do not name resources in PLR based on vendor (anyone who sells something) because they constantly change, but instead on the original equipment manufacturer.

One thing I'd like to raise rather than change unilaterally:
commit `2c214a7` renames **"Resource Library" → "Resource Catalog"** across the landing page, captions, and the PLR-RL term. I'd gently suggest keeping the existing **"Resource Library"**:

- "Catalog" carries a commercial/storefront connotation, whereas this collection is explicitly open-source and crowd-sourced/peer-reviewed — "Library" fits that ethos and avoids implying a vendor catalog.
- The feature is modeled after Opentrons, which calls theirs the **"Labware Library"**, not a catalog.
- It's the long-standing PLR term — **"Resource Library (PLR-RL)"** is still what `main` and the wider docs/community use; this PR is the first to drop it.

Please let us know what you think :)

docs/_static/plr_geometry_viewer.js
```diff
@@ -737,6 +737,12 @@
       };
     }
 
+    rulerUiScale() {
+      // Keep rulers a bit larger at baseline and grow with zoom so labels and
+      // axis strokes stay legible while inspecting details.
+      return clamp(1.1 + (this.zoom - 1) * 0.28, 1.0, 2.0);
+    }
+
```
BioCam, line 745: absolutely love it!

norle:
Hi!

I like all the changes, I think it looks a lot better now!

About the catalog naming, it was my and my AIs halucination, at first I thought of not replacing the previous version of the library in the docs so I named it different and then I remove the previous version. I renamed it back to resource library now, I hope I reverted everything.

I noticed some things about the labware. I am not sure if these are visualization issues or resource issues.

First the tips get rendered weird
image

This troughplate doesnt look right.
image

This wellplate's wells look a bit off.
image

And this trough plate has non rectangular troughs but the irl image looks like it doesn't
image

BioCam:
> First the tips get rendered weird
> image

Interestingly, this is actually very accurate, it is just the very first time I have seen PLR's current TipRack accurately rendered:

The TipRack model has always been a bit odd: it declared the position of the TipSpot at the bottom of the tips which it houses.
But in reality this has multiple issues, including (1) the TipRack itself doesn't actually differ (for Hamilton) tipracks, so the model lies, (2) TipSpots render below the tipracks as you can see in your 3D renders.

I believe @rickwierenga fixed this TipRack issue in v1. It would be very interesting to merge your PR into main AND v1 and see how the render for TipRacks differs with the new TipRack architecure :)

BioCam:
I will have a look at the other issues you flagged - this is fantastic because it really flags up the inconsistencies in PLR's Resource Library and forces us to address them.

E.g. another issue: it clearly highlights when the resource definition is actually incorrect in the code...
Screenshot 2026-05-17 at 11 48 05

...the container inside the plate cannot be closed up by the resource; it can only be level with the top or protruding above it :)

BioCam:
This also flags a wider issue:

1. Two sources of truths for resource information: the Python function VS the docs md file -> I recommend we choose only the Python definitions docstring as sources of truth (they are easier to generate, access at runtime and live right next to the code)
2. There is no resource definition docstring standard
3. We are missing _hundreds_ of definitions in the docs: this PR initially only covered 95, I fixed a couple of md file errors to get it to about 105, but there over 300 definitions in the code, and we haven't yet figured out whether to move all Tecan and Opentrons premade definitions into our standard

BioCam:
> This troughplate doesnt look right.
> image

All these issues in the rendering are actually not rendering issues but correct rendering flags!

These appear to be real errors people made when creating these definitions:
- here we see the definition having flipped the x and y values for the container AND
- because the container defaults to cross-section circle, it renders incorrectly -> no cross-section explicitly declared in this definition as shown here:

Screenshot 2026-05-17 at 17 09 26

...showcasing why your 3D render is highly valuable and improves the quality of the PLR definition by identifying mistakes made!

norle:
> 1. Two sources of truths for resource information: the Python function VS the docs md file -> I recommend we choose only the Python definitions docstring as sources of truth (they are easier to generate, access at runtime and live right next to the code)

Using the Python definitions would make much more sense, I didn't know the difference was so large! However, I saw that @rickwierenga has mentioned about switching to toml for labware definitions (https://discuss.pylabrobot.org/t/toml-instead-of-py-files-for-labware/500, https://discuss.pylabrobot.org/t/april-24th-9am-pt/489/4). We could then rely on the toml definitions instead. Is this something that's already happening in V1? I would think that switching to toml would be out of scope of this PR.

BioCam:
I'm not sure whether an official decision has been made on transitioning all definitions to toml format yet.
I think it makes sense but details would have to be figured out, including what Python version would then still be supported because standard library-based toml support has been introduced in Python 3.11 - i.e. PLR would stop supporting the prior Python version it currently supports
(a natural consequence of living code at some point, just has to be a clear decision and communicated)

But either way that is a separate big change across the codebase.

And first, we have to identify and fix any incorrect definitions and then the naming inconsistencies, and handle all the deprecations.

I've fixed the incorrect plate definition functions you highlighted in the separate PR mentioned above.
(Having this system in place to identify these issues faster is fantastic and improves definitions quality)

norle:
The geometry was already generated based on the Python code but only labware with md doc files was shown. I chagned it so that all of them are seen. I also allow for Opentrons json to be downloaded. I also changed catalog to library in some places that was still wrong (wonder how that happened hehe). This is a report about the geometry and docs 
Network-enabled local report:

labware items identified: 415
items with markdown enrichment: 104
items with geometry: 336
items without geometry: 79
failures recorded: 79
Main reasons for missing geometry:

31 undefined lids: NotImplementedError: This lid is not currently defined.
29 Tecan/Hamilton carriers: pedestal_size_z must be provided.
2 Opentrons definitions: HTTP 404 from upstream JSON URL.
17 deprecated/alias/unsupported definitions raising NotImplementedError or ValueError.
Opentrons specifically improved: most Opentrons resources now get geometry when network access is available. Remaining Opentrons failures:

opentrons_24_aluminumblock_nest_1_5ml_screwcap -> HTTP 404
opentrons_24_tuberack_nest_0_5ml_screwcap -> HTTP 404

BioCam:
Yeah, I've been diving through the chaos since you started this PR :)

Screenshot 2026-05-18 at 23 54 57

This is genuinely amazing because it is deeply needed and you started it!

Background info: when PLR was first developed I believe the Opentrons, Tecan and initially the Hamilton information was straight used to generate definitions... i.e. might have never been used or verified by a human on a device.

Hamilton resources are now pretty-well battle-tested.

But Opentrons and Tecan might not be, simply to fewer people using PLR for their devices.

I therefore disregard them mostly in my analysis.

PR #1047, opened 2026-05-19
Add HighRes Biosolutions MicroSpin centrifuge backend

c-reiter:
# Add HighRes Biosolutions MicroSpin centrifuge backend

This PR adds an integration for the HighRes Biosolutions **MicroSpin** automated microplate centrifuge, plus an in-process mock TCP server that emulates the device's wire protocol for hardware-free development.

It also refactors the `pylabrobot.centrifuge` module to a per-vendor folder layout matching `pylabrobot.plate_reading`.

### Backend — `pylabrobot.centrifuge.highres.MicroSpinBackend`

- Speaks the MicroSpin's ASCII command/response protocol over TCP/1000 (port configurable; matches the device's web-UI `SERVER_PORT` setting).
- Reverse-engineered from the MicroSpin User Manual (HighRes doc 1058675 Rev C, §6.6) and from the firmware's own `list all` / `help` / `settings` introspection commands.
- Maps the abstract `CentrifugeBackend` API to the right wire commands. The four maintenance-only lock primitives (`lockdoor`, `unlockdoor`, `locknest`, `unlocknest`) raise `NotImplementedError` per the contributor guide: the higher-level `open <bucket>` and `spin` commands handle locking internally, and exposing the primitives would let callers put the device into half-managed states. An escape hatch via `backend.send_command(...)` exists for service techs who really need them.
- MicroSpin-specific helpers: `home()`, `is_homed()`, `abort()`, `clear_button_abort()`, `get_status()`, `get_version()`, `get_errors()`, `wait_for_spindle_stopped()`, and `reset()`.
- **`reset()`** issues `abort` → `clearbuttonabort` → `status`. The third step is the real "we are stopped" gate: the first two return `OK!` immediately on the wire (just acknowledgements), but `status` is queued behind any active motion by the firmware and only answers once the rotor is genuinely stopped.
- Spin parameters are validated; the 0-1 acceleration/deceleration fractions are converted to the integer 0-100 percentages the firmware expects.
- Calling `spin()` with `g < 30` emits a `UserWarning` — empirically the spindle-stopped sensor sometimes fails to latch at very low G, causing every subsequent command to time out.

### Mock server — `pylabrobot.centrifuge.highres.MicroSpinMockServer`

A localhost asyncio TCP server speaking the same wire protocol, usable two ways:

```python
async with MicroSpinMockServer() as srv:
    cf = MicroSpin(name="mock", host=srv.host, port=srv.port)
    await cf.setup()
    # ... drive it like a real device ...
```

```bash
python -m pylabrobot.centrifuge.highres.mock_server --port 1000
nc 127.0.0.1 1000
```

- Implements the firmware's "`status` blocks until the spindle has truly stopped" semantics — this is what makes `reset()` work as a synchronisation primitive.
- `simulate_low_g_hang` flag reproduces the firmware quirk above so the recovery path can be tested deterministically.

## Refactor: per-vendor folder layout

The `pylabrobot.centrifuge` module now mirrors `pylabrobot.plate_reading`:

```
pylabrobot/centrifuge/
├── __init__.py             (re-exports all vendors)
├── access2.py              (deprecation shim → agilent/)
├── backend.py
├── centrifuge.py
├── centrifuge_tests.py
├── chatterbox.py
├── standard.py
├── vspin_backend.py        (deprecation shim → agilent/)
│
├── agilent/                ← NEW (Agilent VSpin + Access2 loader)
│   ├── __init__.py
│   ├── access2.py
│   └── vspin_backend.py
│
└── highres/                ← NEW (HighRes Biosolutions MicroSpin)
    ├── __init__.py
    ├── microspin.py
    ├── microspin_backend.py
    ├── microspin_tests.py
    ├── mock_server.py
    └── mock_server_tests.py
```

**Existing imports keep working.** `pylabrobot.centrifuge.vspin_backend` and `.access2` re-export from `.agilent.*` and emit `DeprecationWarning`, identical to the pattern used by `plate_reading.biotek_backend` etc.

## Other fixes

- Imported `unittest.mock` in `centrifuge_tests.py` (pre-existing bug that prevented the test class from running on Python 3.12).

## Tests

| File | Count | What |
|---|---|---|
| `microspin_tests.py` | 23 | Stub-based: protocol edge cases (bad ACK, EOF), argument validation, `reset()` error branches, timeout extension logic, serialization. |
| `mock_server_tests.py` | 24 | Real-TCP integration: every backend method against the mock, plus the status-blocks-during-motion gate and the low-G hang reproduction. |
| `centrifuge_tests.py` | 13 | Unchanged front-end tests. |

**60 new centrifuge tests** all pass in ~3 s. Full project suite still passes (1643 tests).

## Docs

- New user-guide notebook: `docs/user_guide/01_material-handling/centrifuge/highres_microspin.ipynb`
  - Network configuration (direct-attached, DHCP, IP discovery)
  - Connecting from PLR with port customisation
  - Homing, status, version, bucket positioning, door operation
  - Spinning with a pre-spin checklist quoting the manual
  - The low-G hang warning + recovery path
  - `reset()` and the three-step abort/cba/status sequence
  - **Developing without hardware** — section pointing developers at the mock server
- `docs/api/pylabrobot.centrifuge.rst` updated to the multi-vendor pattern (backends grouped under `agilent.*` and `highres.*` paths).

## Contributor-guide compliance

| Check | Status |
|---|---|
| `make test` | ✅ 1643 passed, 1 skipped |
| `make lint` (ruff) | ✅ |
| `make format-check` (ruff format + import order) | ✅ |
| `make typecheck` (mypy `--check-untyped-defs`) | ✅ |
| `make docs` (strict, warnings-as-errors) | ✅ |
| Google-style docstrings on all functions/classes | ✅ |
| Type hints | ✅ |
| Pre-commit hooks (ruff-format, ruff-check, typos) | ✅ |
| OS-agnostic (no Win/Mac/Linux-specific code) | ✅ (pure stdlib asyncio) |

## Safety note

This integration was developed against a physical HighRes MicroSpin during the reverse-engineering phase, but **the spin command itself was never executed by the integration code** — all wire-level behaviour in the test suite runs against the mock server. The first real spin on any given unit should follow the manual's commissioning checklist (§§5.3, 7, 8): tie-wraps removed, buckets seated and balanced, chamber clean, air pressure 70-135 psi, door closed.

## Backwards compatibility

No breaking changes. The deprecation shims preserve all existing import paths from `pylabrobot.centrifuge.vspin_backend` and `pylabrobot.centrifuge.access2`.


rickwierenga:
thanks for the PR this is sick! first high res device in PLR

rickwierenga:
while the vendor refactor is a good idea, we are actually about to move to a new architecture (see #1000, https://discuss.pylabrobot.org/t/updating-plr-api-for-machine-interfaces-discussion/445) and I do not want to introduce more deprecation warnings before then..

rickwierenga:
> This integration was developed against a physical HighRes MicroSpin during the reverse-engineering phase, but the spin command itself was never executed by the integration code 

will you be able to test this on a machine?

rickwierenga:
> There is no separate "open the door without a bucket" workflow.

nice, this will be useful for the refactor (we usually/always to have tweak the abstraction layer when adding a 2nd machine with a particular capability). for now yes I think the not implemented errors are correct to minimize changes to the existing Centrifuge in the v0 api

rickwierenga:
can I rewrite this using `pylabrobot.io.Socket`? this is a thin wrapper around StreamReader and StreamWriter you already use, but has nice logging and [validation](https://docs.pylabrobot.org/stable/user_guide/machine-agnostic-features/logging-and-validation/validation.html)

c-reiter:
> thanks for the PR this is sick! first high res device in PLR

thanks!

> while the vendor refactor is a good idea, we are actually about to move to a new architecture...

makes sense - happy with the revert

> will you be able to test this on a machine?

yep, just ran the integration script end-to-end on the real unit. 300g 5s spin, door auto-closed, no new errors on the stack. all good

> for now yes I think the not implemented errors are correct...

thanks

> can I rewrite this using `pylabrobot.io.Socket`?

go for it

rickwierenga:
> makes sense - happy with the revert

already done in https://github.com/PyLabRobot/pylabrobot/pull/1047/commits/b45f61fb41ae16e40b8d1d5e389c7e62a4767d7b :)

rickwierenga:
rewrote using plr.io.socket layer. would you please be able to just confirm everything works as expected still? should be fine since I was able to test with the mock server

PR #1046, opened 2026-05-18
feat(spark): add spectrum scan support for absorbance and fluorescence

xbtu2:
## Summary

Add `read_absorbance_spectrum()` and `read_fluorescence_spectrum()` methods to `ExperimentalSparkBackend`, enabling wavelength-sweep spectral scanning on the Tecan Spark plate reader.

## Key Design Decision

The Spark firmware natively supports wavelength range sweeps via `WAVELENGTH=FROM~TO:STEP` syntax in its filter commands. This implementation leverages the native sweep rather than issuing per-wavelength commands, which is both faster and simpler.

### `controls/optics_control.py`
- `wavelength` param type: `Optional[int]` → `Optional[Union[int, str]]` to accept firmware range strings (e.g. `'5000~5500:100'`)

### `spark_backend.py`
- Added `read_absorbance_spectrum()` and `read_fluorescence_spectrum()` public methods
- Extracted shared helpers to reduce duplication:
  - `_run_measurement()`: shared background-read → scan → cleanup orchestration
  - `_setup_absorbance()` / `_setup_fluorescence()`: shared instrument configuration

### `spark_processor.py`
- Added `process_absorbance_spectrum()` and `process_fluorescence_spectrum()` for multi-wavelength TDCL data
- Extracted shared helpers:
  - `_find_key()`: dynamic TDCL key lookup (replaces repeated `next()` patterns)
  - `_extract_measurement_pairs()`: unifies nested_mult and flat block handling
  - `_extract_fluo_calibration()`: shared dark/bright/K extraction
  - `_reshape_to_rows()`: shared wavelength→rows×cols reshaping
  - `_get_dark_for_wl()`: per-wavelength dark value lookup with fallback

### Tests
- 14 new tests covering spectrum scan methods and parameter validation
- All 25 tests pass

## Physical Hardware Validation

Tested on a physical Tecan Spark via USB/IP forwarding:

| Scan | Range | Result |
|------|-------|--------|
| Absorbance spectrum | 500–550nm, 10nm steps | 6 wavelength points, OD ≈ 0 (empty well) ✅ |
| Fluorescence spectrum | ex 440–480nm → em 545nm | 5 wavelength points, 353→213 RFU ✅ |

PR #1048, opened 2026-05-19
feat(spark): add fluorescence emission spectrum scan

xbtu2:
## Summary

Add `experimental_read_fluorescence_emission_spectrum()` to `ExperimentalSparkBackend`, which sweeps the **emission** wavelength while keeping excitation fixed. This complements the existing `experimental_read_fluorescence_spectrum()` (excitation sweep) from #1046.

### `spark_backend.py`
- Widened `_setup_fluorescence` `em_wavelength` type from `int` → `Union[int, str]` to accept firmware range strings
- Added `experimental_read_fluorescence_emission_spectrum()` — mirrors the excitation sweep method but passes the range string to the emission filter instead

### `spark_backend_tests.py`
- Added happy-path test verifying correct wavelength mapping and data structure
- Added validation tests for start ≥ end and step ≤ 0

## Physical Hardware Validation

Tested on a physical Tecan Spark via USB/IP:

| Scan | Parameters | Result |
|------|-----------|--------|
| Emission spectrum | ex=440nm fixed, em=500–560nm, 10nm steps | 7 wavelength points, 3654→1483 RFU ✅ |

The emission decay curve shape (monotonically decreasing from peak toward red) confirms correct wavelength mapping and calibration.

### DIFF
--- a/pylabrobot/plate_reading/tecan/spark20m/spark_backend.py
+++ b/pylabrobot/plate_reading/tecan/spark20m/spark_backend.py
@@ -161,7 +161,7 @@ class ExperimentalSparkBackend(PlateReaderBackend):
   async def _setup_fluorescence(
     self,
     ex_wavelength: Union[int, str],
-    em_wavelength: int,
+    em_wavelength: Union[int, str],
     bandwidth: int,
     num_reads: int,
     gain: int,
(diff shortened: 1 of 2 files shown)


rickwierenga:
<3

PR #1052, opened 2026-05-20
Refactor `STARBackend.move_iswap_y` for smooth motion

BioCam:
Replaces the C0 QG read + C0 GY relative-step-with-splitting pattern with a single R0 YA absolute move via `iswap_rotation_drive_move_y`. The rotation drive Y carriage and the gripper translate rigidly together in Y, so the algorithm reads the current gripper Y from `iswap_request_pose` (FK-based, correct in all states), reads the current rotation drive Y direct from R0 RY, and issues one `iswap_rotation_drive_move_y` for `current_rotation_drive_y + (target_gripper_y - current_gripper_y)`.

Four behavioural improvements:

- **Corrects a parked-state landing bug.** From PARKED state, the previous code read the gripper Y from `request_iswap_position` (C0 QG), which actually returns the rotation drive XY (not the gripper XY). The old algorithm therefore landed the **rotation drive** at `y_position` instead of the gripper - up to ~276 mm wrong depending on arm orientation. The new path uses FK-based `iswap_request_pose`, correct in all states.
- Smooth motion. The previous path split moves > 99.9 mm into multiple GY commands, each with its own decel-and-stop in between. The new path issues one R0 YA with a single continuous motion profile.
- Caller-controllable motion parameters. `speed` (mm/sec), `acceleration_level`, `current_protection_limiter`, and `make_space` are now optional kwargs with the same defaults as `iswap_rotation_drive_move_y`. The previous path used Hamilton's firmware-internal GY defaults with no PyLabRobot-side override. Additive and non-breaking - existing positional call sites remain valid.
- Inherits the safety machinery of `iswap_rotation_drive_move_y` (Y hardware range, channel-0 collision detection with optional `make_space=True` auto-clearing, speed/accel/current-limit range checks).

Part 18 of the "Tame the iSWAP" epic:

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -9918,8 +9918,24 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
         `speed`, `acceleration_level`, or `current_protection_limiter` is
         outside its valid range.
     """
-    current_gripper_y = (await self.iswap_request_pose()).location.y
-    current_rotation_drive_y = await self.iswap_rotation_drive_request_y()
+    # Read the full joint state once so the current rotation-drive Y and the
+    # FK-derived current gripper Y come from the same snapshot - avoids the
+    # redundant R0 RY read and any race window between the two values.
+    joints = {
+      STARBackend.iSWAPAxis(k): v for k, v in (await self.iswap_request_joint_state()).items()
+    }
+    current_rotation_drive_y = joints[STARBackend.iSWAPAxis.Y]
+    current_gripper_y = STARBackend._iswap_fk(
+      joints=joints,
+      link_1_length=self.iswap_information.link_1_length,
+      link_2_length=self.iswap_information.link_2_length,
+      wrist_straight_angle=STARBackend._iswap_wrist_drive_increments_to_angle(
+        self.iswap_information.wrist_drive_predefined_increments[
+          STARBackend.WristDriveOrientation.STRAIGHT
+        ]
+      ),
+    ).location.y
+
     await self.iswap_rotation_drive_move_y(
       y=current_rotation_drive_y + (y_position - current_gripper_y),
       speed=speed,
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -9918,24 +9918,8 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
         `speed`, `acceleration_level`, or `current_protection_limiter` is
         outside its valid range.
     """
-    # Read the full joint state once so the current rotation-drive Y and the
-    # FK-derived current gripper Y come from the same snapshot - avoids the
-    # redundant R0 RY read and any race window between the two values.
-    joints = {
-      STARBackend.iSWAPAxis(k): v for k, v in (await self.iswap_request_joint_state()).items()
-    }
-    current_rotation_drive_y = joints[STARBackend.iSWAPAxis.Y]
-    current_gripper_y = STARBackend._iswap_fk(
-      joints=joints,
-      link_1_length=self.iswap_information.link_1_length,
-      link_2_length=self.iswap_information.link_2_length,
-      wrist_straight_angle=STARBackend._iswap_wrist_drive_increments_to_angle(
-        self.iswap_information.wrist_drive_predefined_increments[
-          STARBackend.WristDriveOrientation.STRAIGHT
-        ]
-      ),
-    ).location.y
-
+    current_gripper_y = (await self.iswap_request_pose()).location.y
+    current_rotation_drive_y = await self.iswap_rotation_drive_request_y()
     target_rotation_drive_y = current_rotation_drive_y + (y_position - current_gripper_y)
     try:
       await self.iswap_rotation_drive_move_y(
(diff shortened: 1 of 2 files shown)


pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -9875,23 +9875,57 @@ async def move_iswap_z_relative(self, step_size: float, allow_splitting: bool =
...
+      RuntimeError: if iSWAP is not installed, or if `setup()` has not
+        populated `iswap_information`.
+      ValueError: if `y_position` is outside the Y hardware range; if the
+        target requires channel 0 to be moved and `make_space=False`; if
+        the move is unreachable even with channel repositioning; or if
+        `speed`, `acceleration_level`, or `current_protection_limiter` is
+        outside its valid range.
+    """
+    current_gripper_y = (await self.iswap_request_pose()).location.y
+    current_rotation_drive_y = await self.iswap_rotation_drive_request_y()
```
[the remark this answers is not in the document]
BioCam: good catch - and actually highlights that we should add follow up PRs to simplify by adding more powerful firmware commands, e.g. found one that does correctly retrieve at least rotation drive and wrist drive increment position in one go

I have applied the fix suggest here but non-destructive simplification refactors that aim to minimise the number of firmware commands will be built as improvement PRs after this one

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -9875,23 +9875,57 @@ async def move_iswap_z_relative(self, step_size: float, allow_splitting: bool =
...
-      step_size=y_position - loc.y,
-      allow_splitting=True,
+  async def move_iswap_y(
+    self,
+    y_position: float,  # TODO: by convention should be just 'y' in v1
+    speed: float = 220.0,
+    acceleration_level: int = 2,
+    current_protection_limiter: int = 7,
+    make_space: bool = False,
+  ):
```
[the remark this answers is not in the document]
BioCam: addressed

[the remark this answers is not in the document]
BioCam: zero of the 27 chatterbox method overrides currently use **kwargs, all mirror signatures explicitly :)

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -9875,23 +9875,83 @@ async def move_iswap_z_relative(self, step_size: float, allow_splitting: bool =
...
+        the move is unreachable even with channel repositioning; or if
+        `speed`, `acceleration_level`, or `current_protection_limiter` is
+        outside its valid range.
+    """
+    # Read the full joint state once so the current rotation-drive Y and the
+    # FK-derived current gripper Y come from the same snapshot - avoids the
+    # redundant R0 RY read and any race window between the two values.
+    joints = {
+      STARBackend.iSWAPAxis(k): v for k, v in (await self.iswap_request_joint_state()).items()
+    }
```
rickwierenga: why is this needed? 

any time a function returns something and you always apply the same transformation, that transformation should probably just exist in that function

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -9875,23 +9875,83 @@ async def move_iswap_z_relative(self, step_size: float, allow_splitting: bool =
...
+    current_gripper_y = STARBackend._iswap_fk(
+      joints=joints,
+      link_1_length=self.iswap_information.link_1_length,
+      link_2_length=self.iswap_information.link_2_length,
+      wrist_straight_angle=STARBackend._iswap_wrist_drive_increments_to_angle(
+        self.iswap_information.wrist_drive_predefined_increments[
+          STARBackend.WristDriveOrientation.STRAIGHT
+        ]
+      ),
+    ).location.y
```
rickwierenga: this is a lot of boiler plate to get the current y location. is there no easier way

BioCam: only with tradeoffs:

I initially set it to `STARBackend.iswap_request_pose()` - i.e. only one command - but this then lead to one redundant firmware call: this firmware command has to call the rotation_drive Y-position inside it, but we then have to also call rotation_drive Y-position outside of it too, to compute the position the rotation drive has to be sent to

-> more code vs one less firmware command tradeoff

(same for the x and z movements)

rickwierenga: why not have a function that returns the pose in both cartesian AND joint space? other arms have a similar problem. could even be a private method

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: full reason below - forward kinematics requires joint information to compute position

But if you think a single redundancy (i.e. calling request Y-position) is better because it is much easier to read, I'm happy to change it :)

rickwierenga: it could return a dictionary where the keys are STARBackend.iSWAPAxis. so far every method I have seen casts the keys to STARBackend.iSWAPAxis

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
BioCam: this is what the forware-kinematics / iswap_request_pose does, no?

rickwierenga: yes but this would just be in one place rather than copying this pattern everywhere

BioCam: I don't understand, how so?

BioCam: changed to using `iswap_request_pose()` directly - trading the one extra firmware command for readability

rickwierenga (approved):

PR #1065, opened 2026-05-27
Implement simple mechanis simulator for Hamilton STAR

burnpanck:
This PR adds a `STARSimulator` which tries to keep track of drive positions for all mayor drives on the STAR (including 96 head and iSWAP). Our use-case are protocol runtime estimates, so the only thing done with the actual drive movements are `await anyio.sleep(move_time)`. The PR depends on the structured concurrency branch, mainly because that one enables running on the `trio` event loop. Trio has that amazing `MockClock`, which fast-forwards whenever everything is sleeping. Note that the PR also bumps the minimum python version to 3.10, because we extensively use the 3.10 `match` statement. While it could certainly be written without that, I have no interest in writing code using outdated syntax. I anticipate that this may be a blocker from actually merging this as-is, but perhaps it's useful to @BioCam in his iSWAP quest anyway.


burnpanck:
I should perhaps also add that the mechanical simulation is still rather crude. We don't do liquid level detection or surface following. We assume all movements are using the same max-speed and acceleration values (per drive type). We don't stagger the pip channels. The main goal is to have at least some estimate rather than none, but there is no intent to be 100 % accurate.

PR #1072, opened 2026-06-04
v1: Brooks PF400: read gripper width limits from the controller; restrict freedrive default to installed axes

### LINKED ISSUE #1071: PreciseFlex `brooks` driver: freedrive and gripper fail on a standard (no-rail) arm
Two defects in the v1b1 `pylabrobot.brooks` PreciseFlex driver, on a physical arm. Both hardcode values the controller can report, so they fail on standard hardware.

**Terms:** axes are 1 `BASE` (Z), 2 `SHOULDER`, 3 `ELBOW`, 4 `WRIST`, 5 `GRIPPER`, 6 `RAIL` (optional, absent on many arms). Freedrive = hand-move mode (`freemode <axis>`, `0` = all). Per-axis soft limits are DataIDs 16077 (max) / 16078 (min), read via `request_parameter`.

**Tested on:** a standard no-rail PreciseFlex 400 (5 axes); GPL 5.1D4 / TCS 3.0D4; `pylabrobot/brooks/precise_flex.py`.

## Issue 1: `start_freedrive_mode()` frees a non-existent axis -> `-2800`

With no args it defaults to freeing `[BASE, SHOULDER, ELBOW, WRIST, RAIL]` (axes 1, 2, 3, 4, 6). On a no-rail arm, `freemode 6` (the rail) returns `-2800 Parameter Mismatch`, so the plain "enter freedrive" call fails.

Cause: the default axis list hardcodes the rail (axis 6), ignoring `has_rail` / whether that axis exists in the joint set.

Fix: default to the installed axes (drop the rail when absent), or send `freemode 0`. Workaround: `start_freedrive_mode(free_axes=[Axis.BASE, Axis.SHOULDER, Axis.ELBOW, Axis.WRIST])`.

## Issue 2: gripper width checked against hardcoded limits -> `-1012`

`move_gripper(width)` maps mm to units as `closed_gripper_position + (width - min_gripper_width)`, with `min/max_gripper_width` hardcoded to 60/145. The arm's real gripper limits (DataIDs 16078/16077, axis 5) are 69-134. A `width` that passes the 60-145 check can map to units outside the real limits, and the move is rejected with `-1012 Joint out-of-range` rather than a clear client-side error.

Fix: take `min/max_gripper_width` from the gripper-axis soft limits, and validate the computed units against them (named error) before sending.

## Note (firmware behaviour, not a bug): freedrive clears homing

After `home()`, freedriving and hand-moving the arm leaves it "not homed" (`-1021`) on the next commanded move - free mode invalidates the homed reference. A teach-then-replay flow must re-home after `stop_freedrive_mode()` (captured joint angles stay valid). The driver could surface this rather than passing through the raw error.

BioCam:
Two defects in the v1b1 `pylabrobot.brooks` PreciseFlex driver, both from hardcoding values the controller reports. Verified on a physical no-rail PreciseFlex 400. Closes #1071.

**Freedrive frees a non-existent axis.** `start_freedrive_mode()` defaulted to `[BASE, SHOULDER, ELBOW, WRIST, RAIL]`; on a no-rail arm `freemode <rail>` returns `-2800`, so entering freedrive fails. It now frees the four always-present axes and adds the rail only when `has_rail` is set; passing an explicit `free_axes` is unchanged.

**Gripper width checked against hardcoded limits.** `move_gripper` validated against `min/max_gripper_width` hardcoded to 60/145, but the gripper's real soft limits (gripper-axis DataIDs 16078/16077) differ - so an accepted width can map to units the controller then rejects with a late `-1012`. `_on_setup` now reads those soft limits (best-effort, falling back to the defaults) and `move_gripper` validates the computed units with a clear `ValueError` before sending.

### DIFF
--- a/pylabrobot/brooks/precise_flex.py
+++ b/pylabrobot/brooks/precise_flex.py
@@ -374,6 +374,22 @@ class PreciseFlexArmBackend(OrientableGripperArmBackend, HasJoints, CanFreedrive
   async def _on_setup(self, backend_params: Optional[BackendParams] = None) -> None:
     await super()._on_setup(backend_params=backend_params)
     await self.stop_freedrive_mode()
+    # Read the gripper-axis soft limits so width validation reflects this arm
+    # rather than the hardcoded defaults (the servo gripper's real range differs).
+    try:
+      soft_min = [float(v) for v in (await self.request_parameter(16078)).split(",")]
+      soft_max = [float(v) for v in (await self.request_parameter(16077)).split(",")]
+      gripper_index = int(Axis.GRIPPER) - 1
+      self._gripper_soft_min = soft_min[gripper_index]
+      self._gripper_soft_max = soft_max[gripper_index]
+      self.min_gripper_width = soft_min[gripper_index]
+      self.max_gripper_width = soft_max[gripper_index]
+    except Exception as exc:  # best-effort; fall back to the class defaults
+      logger.warning(
+        "[PreciseFlex %s] could not read gripper soft limits, using defaults: %s",
+        self.driver.io._host,
+        exc,
+      )
 
   async def _request_state(
     self,
@@ -418,9 +434,13 @@ class PreciseFlexArmBackend(OrientableGripperArmBackend, HasJoints, CanFreedrive
     """Get the current speed percentage of the arm's movement."""
     return await self.request_profile_speed(self.profile_index)
 
-  # Physical jaw range for the PF400 servoed gripper.
+  # Physical jaw range for the PF400 servoed gripper. Overridden at setup from the
+  # gripper-axis soft limits when readable; these are the fallback defaults.
   min_gripper_width: float = 60.0
   max_gripper_width: float = 145.0
+  # Gripper-axis soft limits in firmware units, read at setup; None until then.
+  _gripper_soft_min: Optional[float] = None
+  _gripper_soft_max: Optional[float] = None
 
   def _mm_to_firmware_units(self, width_mm: float) -> float:
     """Convert a jaw width (mm) to the firmware's native position unit.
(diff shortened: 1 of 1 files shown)


pylabrobot/brooks/precise_flex.py
```diff
@@ -374,6 +374,22 @@ def __init__(
...
+      self._gripper_soft_min = soft_min[gripper_index]
+      self._gripper_soft_max = soft_max[gripper_index]
+      self.min_gripper_width = soft_min[gripper_index]
+      self.max_gripper_width = soft_max[gripper_index]
+    except Exception as exc:  # best-effort; fall back to the class defaults
+      logger.warning(
+        "[PreciseFlex %s] could not read gripper soft limits, using defaults: %s",
+        self.driver.io._host,
+        exc,
+      )
```
rickwierenga, line 392: should this go into a config object?

pylabrobot/brooks/precise_flex.py
```diff
@@ -750,13 +778,13 @@ async def start_freedrive_mode(
...
-      Axis.WRIST,
-      Axis.RAIL,
-    ]:
+    if free_axes is None:
+      # Free the always-present positioning axes; add the rail only when fitted -
+      # freemode on an absent axis returns -2800 on a no-rail arm.
+      free_axes = [Axis.BASE, Axis.SHOULDER, Axis.ELBOW, Axis.WRIST]
+      if self._has_rail:
+        free_axes.append(Axis.RAIL)
+    for axis in free_axes:
```
rickwierenga, line 787: nice!

pylabrobot/brooks/precise_flex.py
BioCam: It definitely should - follow up PR with responsibility if introducing this will be up in minutes (but out of scope for this bug fix)

PR #1089, opened 2026-06-12
`STARBackend`: add basic 96-head aspirate/dispense

BioCam:
Adds `head96_basic_aspirate` and `head96_basic_dispense` - minimal 96-head liquid commands that own the dispensing-drive (piston) and Z-drive motion of the aspirate / dispense action (the firmware runs both in parallel). They compute no positions from geometry; the only thing they read is the tip overhang, to express `minimum_height` in tip-bottom space. Builds on #1084 (the `Head96Information` ranges + converters), #1086 (the stop-disk / tip-bottom Z reference), and #1087 (the dispensing-drive default speed).

They are the direct, full-control counterparts to `aspirate96` / `dispense96`: instead of a resource and liquid class they take the raw piston volume, height, surface-following and flow directly - no liquid class, no air-transport / blow-out / mix, no curve correction. The head is rigid, so both pin the full channel pattern (no per-channel selection).

```python
  @_requires_head96
  async def head96_basic_aspirate(
    self,
    volume: float, # piston movement volume (!= liquid volume for most liquid types)
    flow_rate: Optional[float] = None, # uL/s, None -> defaults to 96-head's default rate, firmware-variant-dependent
    minimum_height: Optional[float] = None, # mm, None -> defaults to 99.98 mm for safety reasons
    surface_following_distance: float = 0.0,
    enforce_minimum_height: bool = True,
    enforce_requires_tip: bool = True, # firmware command itself does not, extra tip presence check has to be performed, might no be wanted if you pick up something that is 
  ):
```

```python
  @_requires_head96
  async def head96_basic_dispense(
    self,
    volume: float,
    flow_rate: Optional[float] = None,
    minimum_height: Optional[float] = None,
    stop_back_volume: float = 0.0,
    surface_following_distance: float = 0.0,
    stop_flow_rate: Optional[float] = None,
    enforce_requires_tip: bool = True,
  ):
```

- `minimum_height` is the lowest the tip end descends to, in tip-bottom space, converted to the firmware stop-disk `zh` via the tip overhang (stop disk - tip bottom, read once move-free). `None` defaults to the deck floor (`MINIMUM_CHANNEL_Z_POSITION`); with no tip the overhang is 0, so it is guarded against the firmware Z floor (`z_range[0]`) directly.
- Both validate every parameter against `Head96Information` ranges and convert human units to firmware increments.
- `enforce_requires_tip` raises if the head holds no tips; the firmware allows tipless `PA`/`PB` (moving air), so it is opt-out, not pinned.
- Thin by design: settling (`to`) and pre-wetting (`dc`) are pinned off - settling is its own basic command, and pre-wetting's interaction with surface following is unverified.

A `STARChatterboxBackend` `head96_request_tip_presence` stub is added so `enforce_requires_tip` works in simulation (it reads the tip tracker and raises if tip tracking is disabled).

Validated on hardware: a `surface_following_distance=8 mm` aspirate then dispense moved the tip exactly 8 mm down and back, repeatably with no drift.

Tests: chatterbox wire-string assertions for `PA` and `PB` (no-tip and with-tip / overhang paths), the `minimum_height` deck-floor default, default flow-rate / stop-flow-rate resolution, the no-tips `RuntimeError`, and an out-of-range-volume guard.

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -8428,6 +8428,160 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
 
   # -------------- 3.10.3 Liquid handling using CoRe 96 Head --------------
 
+  @_requires_head96
+  async def head96_basic_aspirate(
+    self,
+    volume: float,
+    *,
+    flow_rate: Optional[float] = None,
+    minimum_height: float,
+    surface_following_distance: float = 0.0,
+    enforce_minimum_height: bool = True,
+    enforce_requires_tip: bool = True,
+  ):
+    """Aspirate on the 96-head with surface following (the firmware drives Z and the dispensing drive
+    in parallel).
+
+    The direct, full-control counterpart to `aspirate96`: it takes the height / surface-following /
+    flow directly rather than a resource and liquid class, and computes no positions. Acts on the whole
+    head (rigid - no per-channel selection) and assumes the head already holds tips. Values are given
+    in human units and converted to firmware increments. This is a basic, thin command: it does not
+    settle after aspirating (settling is a separate basic command).
+
+    Args:
+      volume: The volume to aspirate through each channel, uL.
+      flow_rate: The dispensing-drive speed, uL/s; None leaves the firmware default.
+      minimum_height: The lowest Z (mm) the head descends to, the end of the aspiration stroke.
+      surface_following_distance: The Z travel during aspiration, mm; 0 keeps the head in place so it
+        cannot drive into the container bottom.
+      enforce_minimum_height: If True, clamp any deeper target to minimum_height; if False, surface
+        following may carry the head past it.
+      enforce_requires_tip: If True, raise if the head holds no tips; if False, allow aspirating air.
+
+    Raises:
+      RuntimeError: If 96-head is not installed, or enforce_requires_tip and the head holds no tips.
+      AssertionError: If firmware info missing or a parameter is out of range.
+    """
+    assert self._head96_information is not None, "96-head information not loaded; run setup()"
+    info = self._head96_information
+    volume_max = info.dispensing_drive_range[1]
+    flow_max = info.dispensing_drive_speed_range[1]
+    z_min, z_max = info.z_range
+    surface_following_max = self._head96_z_drive_increment_to_mm(9999)
+
+    assert 0 <= volume <= volume_max, f"volume must be between 0 and {volume_max} uL"
+    if flow_rate is not None:
+      assert 0 <= flow_rate <= flow_max, f"flow_rate must be between 0 and {flow_max} uL/s"
+    assert 0 <= surface_following_distance <= surface_following_max, (
+      f"surface_following_distance must be between 0 and {surface_following_max} mm"
+    )
+    assert z_min <= minimum_height <= z_max, (
+      f"minimum_height must be between {z_min} and {z_max} mm"
+    )
+
+    if enforce_requires_tip:
+      has_tips = await self.head96_request_tip_presence()
+      if not has_tips:
+        raise RuntimeError(
+          "96-head has no tips (firmware reports none); pick up tips before aspirating"
+        )
+
+    to_z = self._head96_z_drive_mm_to_increment
+    to_vol = self._head96_dispensing_drive_uL_to_increment
+    params = {
+      "pm": "F" * 24,  # all 96 channels; the rigid head has no per-channel selection
+      "dj": "1" if enforce_minimum_height else "0",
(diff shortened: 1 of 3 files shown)


pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -8428,6 +8442,189 @@ async def discard_tips_core96(
 
   # -------------- 3.10.3 Liquid handling using CoRe 96 Head --------------
 
+  @_requires_head96
+  async def head96_basic_aspirate(
+    self,
+    volume: float,
+    flow_rate: Optional[float] = None,
+    minimum_height: Optional[float] = None,
```
rickwierenga, line 8450: why do we need a public method named `head96_basic_aspirate`? what does it do? how is it different from `aspirate96`?

BioCam: `head96_basic_aspirate` is the true atomic command - i.e. it cannot be broken down further into smaller commands: it is solely the parallel motion of the dispensing_drive up in z and the pipette's z_drive down in z if surface following is active.

i.e. it is massively simpler than `aspirate96` which performs coordinated positioning of the head96 with required iSWAP parking, channel raising +
+ head96 movement to traverse height
+ blowout air volume actuation
+ computation of z movement 
+ lldmode/liquid level search actuation
+ settling time actuation
+ swap speed actuation
+ movement to traverse height actuation
...all independent actuations that are actually separate from the _pure_ aspirate action

We have reached the limits of what PLR can accomplish while maintaining complex firmware commands.
As a result we now need to build the true atomic/indivisible commands into PLR - due to the long-term established usage of the term "atomic" I have instead chosen "basic" here for these commands.

BioCam: This is in direct analogy to what Opentrons has done: divide commands into basic commands and complex commands.

Basic commands then enable anyone to quickly build their own complex function when needed from predictable, minimal input/argument basic methods.

pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -8428,6 +8442,189 @@ async def discard_tips_core96(
...
 
+  @_requires_head96
+  async def head96_experimental_aspirate(
+    self,
+    volume: float,
+    flow_rate: Optional[float] = None,
+    minimum_height: Optional[float] = None,
+    surface_following_distance: float = 0.0,
+    enforce_minimum_height: bool = True,
+    enforce_requires_tip: bool = True,
```
rickwierenga: why does `enforce_minimum_height` exist? why not just have `minimum_height` none when you dont want it changed?

rickwierenga: cleaner name

```suggestion
    requires_tip: bool = True,
```

BioCam: Initially I chose against this because _a_ pattern (though inconsistent) in PLR is that `None` indicates a default value will be used, not the absence of an action altogether.

For the _complex command_ `aspirate96` e.g. `minimum_height = None` should compute the cavity_bottom of the target container and setting it as the `minimum_height` that is sent in the firmware command.

But in this basic/simple/experimental command we do not presume the existence and targeting of a container at all, and as a consequence, there is no cavity_bottom that the default `None` can compute, leaving its meaning ambiguous.

As a result, we need to ask what pattern are people meant to expect from PyLabRobot:
1. `minimum_height = None` -> **default value** is automatically chosen for parameter
2. `minimum_height = None` -> parameter is is automatically **disabled**
?

We could argue that if no cavity_bottom exists None is set to the `disable` which is what you are proposing?
The only potential issue I see here is inconsistency in API design.

BioCam: On second thought there is another reason besides semantics/expectability:

There is a safety cost to collapsing `None` to set both independent firmware parameters like this:

Right now `minimum_height=None` resolves to the **deck floor**, so a caller who does not think about it at all still gets the protective hard-stop by default.
If `None` instead meant "don't clamp," the do-nothing default would become "no floor at all" - i.e. a user could still set a surface following distance that _crashes the tips into the deck_, something we must avoid to ensure device safety.

Keeping the two separate lets both defaults point at safety: `None` keeps its safe deck-floor meaning and `enforce_minimum_height=True` keeps the clamp on.

BioCam: Convinced: in our private conversation -> minimum_height enforcement should ALWAYS be active:

If `None` it should simply compute `max(deck_height, min_head96_stop_disk_height-tip_overhang)`

which can be overwritten by the PLR programmer to a higher max than this.

i.e. `enforce_minimum_height` becomes redundant

PR #1094, opened 2026-06-19
`STARBackend`: add 96-head capacitive LLD Z-probe

BioCam:
Adds `head96_probe_z_using_clld` - a capacitive LLD (cLLD) liquid-surface probe for the 96-head, the head counterpart of the single-channel `clld_probe_z_height_using_channel`. It runs a downward cLLD search on the stop-disk Z drive (`H0 ZL`) and returns the detected surface as a tip-bottom Z height. Builds on #1086 (the stop-disk / tip-bottom Z reference pair) and #1088 (the per-drive speed/acceleration scoping).

Positions are tip-bottom referenced like `head96_move_tool_z`: the tip overhang (stop disk minus tip bottom, a rigid constant for the mounted tip) maps `lowest_immers_pos` and `start_pos_search` to the firmware's `zh` / `zc`, and the return value maps the stopped position back to tip-bottom. `tip_len` and `approach_speed` default to `None`, resolving to the measured tip length and `head96_z_drive_speed_default`. A firmware error during the search retracts the head to Z-safety before re-raising.

The `ZL` wire format changed between the 2008 and 2013 firmware command sets - the 2013 set adds the `lm` sensor-selection field and widens the acceleration / current-limit fields - so the parameters are formatted per the head's reported firmware date, in the documented field order; `lm` selection requires 2013+ firmware.

Validated on hardware: the cLLD search is accepted (`er00`) across `lm` modes. An earlier 6-digit `zc` field width was rejected (`er32`) and is now 5 digits.

Tests: the assembled wire string (guarding the `zc` width and the tip-overhang offset), the no-tip guard on both tip-length paths, and the Z-safety retract on firmware error.

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -8611,6 +8611,8 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
     start_pos_search: Optional[float] = None,
     speed: float = 10.0,
     acceleration: float = 300.0,
+    approach_speed: Optional[float] = None,
+    current_protection_limiter: int = 15,
     lld_sensor: Literal["G11 or H12", "A1 or B2", "any", "all"] = "any",
     detection_edge: int = 10,
     detection_drop: int = 2,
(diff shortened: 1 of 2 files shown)


pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -8601,6 +8602,209 @@ async def head96_move_tool_z(self, z: float, speed: Optional[float] = None):
...
+    if not z_speed_min <= approach_speed <= z_speed_max:
+      raise ValueError(
+        f"approach_speed must be between {z_speed_min} - {z_speed_max} mm/sec, is {approach_speed}"
+      )
+    if not z_speed_min <= speed <= z_speed_max:
+      raise ValueError(f"speed must be between {z_speed_min} - {z_speed_max} mm/sec, is {speed}")
+    if not z_accel_min <= acceleration <= z_accel_max:
+      raise ValueError(
+        f"acceleration must be between {z_accel_min} - {z_accel_max} mm/sec**2, is {acceleration}"
+      )
```
rickwierenga, line 8686: maybe at some point we can abstract this now that many features use the same pattern

BioCam: Yes I agree, probably 2-3 more commands and we'll be able to see a nice abstraction to apply :)

PR #1096, opened 2026-06-20
`STARBackend`: create 96-head `mix96`

BioCam:
Creates `mix96` - in-place mixing with the rigid 96-head, built on the `head96_experimental_aspirate` / `head96_experimental_dispense` primitives. It positions the head over a target and runs repeated surface-following aspirate / dispense cycles, taking the mix parameters directly rather than a liquid class. Builds on #1089 (the experimental aspirate / dispense), #1081 (the `head96_move_x` acceleration control), and #1086 / #1088 (the tip-bottom Z reference and per-drive speed / acceleration scoping).

```python
  @_requires_head96
  @need_iswap_parked
  async def mix96(
    self,
    mix: Mix,
    resource: Optional[Union[Plate, Container, List[Well]]] = None,  # aspirate96-style target
    a1_coordinate: Optional[Coordinate] = None,        # or an explicit channel-A1 deck target
    minimum_traverse_height_start: Optional[float] = None,  # abs Z before X/Y; None -> Z-safety
    offset: Coordinate = Coordinate.zero(),            # added to the target; offset.z lifts the floor
    lld_mode: Optional[LLDMode] = None,                # None -> LLDMode.OFF (only OFF supported)
    descent_speed: float = 80.0,                       # mm/s, fast descent to just above the well
    swap_speed: float = 5.0,                           # mm/s, careful descent into the well and exit
    settling_time: float = 0.0,                        # s, wait after the last cycle before exit
    minimum_traverse_height_end: Optional[float] = None,    # abs Z after mixing; None -> Z-safety
  ):
```

The target is declared in exactly one of two ways, like `aspirate96`: an explicit `a1_coordinate`, or a `resource` (a Plate - head A1 over well A1 - a Container, or a list of Wells) plus `offset`. All Z is tip-bottom referenced: the declared Z (or a well's cavity_bottom) is the floor, the deepest the tip end reaches, and `offset.z` lifts the whole stroke off it.

Motion: parks the iSWAP, traverses to a start height, moves X/Y (X acceleration dialled down below y=200 mm, the low-Y zone where the head is cantilevered off the X-drive), descends in two stages (`descent_speed` to just above the well, then `swap_speed` into it), runs `mix.repetitions` symmetric cycles between the floor and `surface_following_distance` above it, waits `settling_time`, then retracts at `swap_speed` before the traverse out.

```python
# explicit channel-A1 deck coordinate
await star.mix96(Mix(volume=50, repetitions=3, flow_rate=100), a1_coordinate=Coordinate(500, 150, 100))

# aspirate96-style resource target (head A1 over well A1)
await star.mix96(Mix(volume=50, repetitions=3, flow_rate=100), resource=plate)
```

`lld_mode` currently accepts only `LLDMode.OFF`; liquid-level-driven auto surface following (computing the surface-following distance from a cLLD probe) is planned as a follow-up.

Validated on hardware: the surface-following primitive this builds on moved the tip exactly 8 mm down and back, repeatably with no drift (#1089).

Tests: two minimal orchestration assertions for the highest-risk paths - the experimental command's `minimum_height` resolves to the tip-bottom floor (a well's cavity_bottom plus `offset.z`), and the careful descent starts surface following at floor + `surface_following_distance` so the stroke spans [floor, floor+sf] and never drives below the floor.

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -9111,16 +9111,17 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
   ):
     """Position the 96-head over a target and mix in place.
 
-    Moves X/Y over the target, descends into the well, then runs ``mix.repetitions`` symmetric
-    aspirate / dispense cycles: each aspirate follows the surface down by
-    ``surface_following_distance`` and each dispense follows it back up, so the tip oscillates
-    without drifting. Returns to a traverse height when done.
+    Raises the single channels to safe Z, then moves X/Y over the target and descends into the
+    well, then runs ``mix.repetitions`` symmetric aspirate / dispense cycles: each aspirate
+    follows the surface down by ``surface_following_distance`` and each dispense follows it
+    back up, so the tip oscillates without drifting. Returns to a traverse height when done.
 
     Z targets are in tip-bottom space (the target Z is where the tip end goes, not the stop
     disk). Declare the target in exactly one of two ways:
-      - ``a1_coordinate``: an explicit deck Coordinate for the channel-A1 position.
-      - ``resource`` (+ ``offset``): like ``aspirate96`` - a Plate (head A1 over well A1), a
-        Container, or a list of Wells.
+
+    - ``a1_coordinate``: an explicit deck Coordinate for the channel-A1 position.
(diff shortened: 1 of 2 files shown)


rickwierenga:
Split `mix96` into a coordinate-based workhorse `mix96_at_coordinate(a1_coordinate, z_top=...)` and a thin `mix96(resource, offset=...)` convenience wrapper that resolves the channel-A1 target and container top from the resource and delegates.

PR #1103, opened 2026-06-23
Add MicroSpin centrifuge docs

c-reiter:
## Summary
- add HighRes MicroSpin to the machines centrifuge table
- add MicroSpin factory/error/mock-server entries to centrifuge API docs
- update README centrifuge examples and clarify backend-specific positioning/door-lock behavior

## Note on the generic centrifuge docs
- `_centrifuge.md` now lists the current common `Centrifuge` frontend API.
- `start_spin_cycle` still exists but is deprecated in favor of `Centrifuge.spin`, so the guide now points users to `spin`.
- `rotate_distance` is not a `Centrifuge` frontend method and is not used by the current Agilent VSpin backend. VSpin low-level positioning/calibration is done with `VSpinBackend.go_to_position(...)`; the guide now mentions that backend-specific method instead.

## Tests
- `sphinx -b dummy docs /tmp/pylabrobot-docs-check -W --keep-going`

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,16 +83,30 @@ data = await pr.read_luminescence()
 
 For Cytation5, use the `Cytation5` backend.
 
-### Centrifuges ([docs](https://docs.pylabrobot.org/0.2.1/user_guide/01_material-handling/centrifuge/_centrifuge.html))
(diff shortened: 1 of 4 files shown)


rickwierenga:
thanks!

PR #1106, opened 2026-06-23
v1: `STARBackend`: drop unused `HamiltonSTARDeck` import

BioCam:
Removes an unused `HamiltonSTARDeck` import in `star.py` (keeping `HamiltonCoreGrippers`). It trips `F401` and so the `v1b1` Linting check fails for every PR based on `v1b1`, not just this one. No behaviour change; nothing imports the symbol from `star.star`.


BioCam:
Closing - redundant one-line lint fix opened in error; the maintainer already has this fix locally and will push it directly.

PR #1109, opened 2026-06-23
Add Agilent BenchCel 4R storage backend

c-reiter:
# Add Agilent BenchCel 4R storage backend

This PR adds an integration for the Agilent **BenchCel 4R** microplate handler, modeled as a PyLabRobot storage machine, plus an in-process mock TCP server that emulates the device's binary wire protocol for hardware-free development.

It follows the usual PLR driver structure: a reverse-engineered backend, a faithful mock server, hardware-free tests, and a user guide.

### Backend — `pylabrobot.storage.agilent.BenchCel4RBackend`

- Speaks the BenchCel's **framed binary protocol over TCP/7612** (port/host configurable). Each application frame is `[1 byte command_id][2 byte LE payload_length][payload]`.
- Reverse-engineered from Agilent VWorks packet captures and live tests against a 4-stacker BenchCel (firmware 3.2.20.0). This is **not** vendor wire-protocol documentation; the public Agilent Quick Guide (G5400-90003) documents operation/safety but not the Ethernet protocol.
- Modeled as an `IncubatorBackend`, since the BenchCel is open, sequential stacker storage. The firmware addresses **whole stackers**, so PLR racks track the expected plate order/content per stacker.
- Implemented commands: home / home-motors (`0x47`/`0x48`), stacker load/unload (`0x60`/`0x61`), pick/downstack + place/upstack (`0x62`/`0x63`), move-to-target (`0x65`), relative jog (`0x66`), pneumatic stacker clamps (`0x67`), full-open/close robot grippers (`0x6a`), save teachpoint (`0x73`), **labware-settings push + commit (`0x7d`/`0x9f`)**, stacker sensors (`0x7e`), current position (`0x85`), general/arm status (`0x87`), axis bounds (`0x99`), and device-error frames (`0x02`).
- Decoded status helpers: `request_stacker_sensors`, `request_all_stacker_sensors`, `request_arm_status`, `request_general_status`, `request_axis_bounds`, `request_current_position`.
- Optional `source_ip` to bind a specific local NIC on multi-homed control hosts.
- **No files are written on the host.** The device has no known command to read teachpoints back; if you need to persist the numeric teachpoints you wrote, keep them in your own protocol/config.

#### Narrowed `dangerously_` API

[...]

PR #1110, opened 2026-06-24
`Plate`: add optional `stacking_z_height` parameter

c-reiter:
## Summary

Adds an optional `stacking_z_height` parameter to the `Plate` class — the vertical pitch (in mm) between two identical plates stacked directly on top of each other (i.e. the height one plate adds to a stack: `size_z` minus the overlap with the plate below).

This mirrors the existing `NestedTipRack.stacking_z_height` and gives plates the same notion of stacking pitch that nested tip racks already have. It's required by plate-stacker devices (e.g. the Agilent BenchCel I'm adding in #1109) and defaults to `None` when unknown, so nothing changes for existing labware.

Per the discussion with @rickwierenga: option (b) — add it as an optional param on `Plate` rather than a new subclass — split out into its own PR.

## Changes

- `Plate.__init__`: new `stacking_z_height: Optional[float] = None` parameter, stored on the instance.
- `Plate.__eq__`: since `stacking_z_height` is a physical dimension, plates that differ in it are different labware and now compare unequal. `__hash__` is restored to `Resource`'s (defining `__eq__` otherwise sets it to `None`), so plates stay hashable, and `stacking_z_height` is added to `__repr__`.
- `Plate.serialize()`: new override so `stacking_z_height` round-trips through `deserialize`. This also fixes a pre-existing gap where `plate_type` was silently dropped on serialization (because `Plate` had no `serialize` override, so the constructor arg was never persisted — `copy()` lost it too).
- `ResourceStack`: bare plates stacked in the z direction now **nest** by their stacking pitch. A plate placed directly on another bare plate sinks in by `size_z - stacking_z_height`, so a stack of `N` identical plates is `size_z + (N - 1) * stacking_z_height` tall (both the reported `get_size_z()` and the child placement). Gated on the value being set, and skipped when the lower plate wears a lid — so existing behavior is unchanged (no library plate sets the value yet). This parallels the lid-nesting `ResourceStack` already did.
- Docs: documented `stacking_z_height` (what it is, how to measure it, why it's needed, and the `ResourceStack` nesting behavior) on the Plate resource page and the ResourceStack notebook, and added it to the example plate definition in the contributor guide.
- Tests: `Plate` — default-is-`None`, value-stored, serialize/deserialize round-trip (covering both `stacking_z_height` and `plate_type`), equality differs on `stacking_z_height`, hashability. `ResourceStack` — two/three plates nest, no-nesting without the value (back-compat), and no nesting onto a lidded plate.

## Notes

- No `Plate` subclasses exist, so adding the keys to `serialize()` (which `deserialize` feeds back as constructor kwargs) is safe.
- Backwards compatible: deserializing older JSON without these keys falls back to the constructor defaults; no existing library plate sets `stacking_z_height`, so `ResourceStack` behavior is unchanged for all current labware.

### DIFF
--- a/pylabrobot/resources/plate.py
+++ b/pylabrobot/resources/plate.py
@@ -80,6 +80,7 @@ class Plate(ItemizedResource["Well"]):
     lid: Optional[Lid] = None,
     model: Optional[str] = None,
     plate_type: Literal["skirted", "semi-skirted", "non-skirted"] = "skirted",
+    stacking_z_height: Optional[float] = None,
   ):
     """Initialize a Plate resource.
 
@@ -89,6 +90,10 @@ class Plate(ItemizedResource["Well"]):
       lid: Immediately assign a lid to the plate.
       plate_type: Type of the plate. One of "skirted", "semi-skirted", or "non-skirted". A
         more complete description is still pending.
+      stacking_z_height: The vertical pitch in mm between two identical plates stacked directly on
+        top of each other (i.e. the height a plate adds to a stack, equal to ``size_z`` minus the
+        overlap with the plate below). Required by some stacker devices (e.g. the Agilent BenchCel)
+        and left as ``None`` when unknown.
     """
 
     super().__init__(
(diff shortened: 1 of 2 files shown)


rickwierenga:
thanks!

BioCam:
Nice!

Should the same be applied to `PetriDish`?

PR #1112, opened 2026-06-24
`ResourceStack`: nest stacked plates by `Plate.stacking_z_height`

c-reiter:
## Summary

Follow-up to #1110, which added `Plate.stacking_z_height`. This PR makes `ResourceStack` actually use it, and brings in the documentation that belonged with that work.

**Why a separate PR:** these changes were force-pushed to the #1110 branch a few minutes *after* it had already been squash-merged, so they never reached `main`. #1110 landed only the `Plate` attribute (param, `serialize()`, `__eq__`/`__hash__`, tests); the `ResourceStack` nesting and the docs were left behind on the merged branch. This PR re-applies that delta on top of current `main`.

## Changes

- **`ResourceStack`**: bare plates stacked in the z direction now **nest** by their stacking pitch. A plate placed directly on another bare plate sinks in by `size_z - stacking_z_height`, so a stack of `N` identical plates is `size_z + (N - 1) * stacking_z_height` tall — applied consistently to both `get_size_z()` and child placement via a shared `_nesting_overlap()` helper. Gated on the value being set, and skipped when the lower plate wears a lid. This parallels the lid-nesting `ResourceStack` already performed.
- **Docs**: documented `stacking_z_height` (what it is, how to measure it, the `ResourceStack` nesting behavior) on the Plate resource page and the ResourceStack notebook, and added it to the example plate definition in the contributor guide.
- **Tests**: two/three plates nest to the correct height *and* positions, no-nesting without the value (back-compat), and no nesting onto a lidded plate.

## Notes

- Backwards compatible: no library plate sets `stacking_z_height`, so `ResourceStack` behavior is unchanged for all existing labware.

### DIFF
--- a/docs/contributor_guide/contributing-new-resources.md
+++ b/docs/contributor_guide/contributing-new-resources.md
@@ -59,6 +59,10 @@ def AGenBio_96_wellplate_Ub_2200ul(name: str, lid: Optional[Lid] = None) -> Plat
     size_x=127.76,  # from spec
     size_y=85.48,  # from spec
     size_z=42.5,   # from spec
+    # Optional: the vertical pitch one plate adds to a stack of identical plates
+    # (size_z minus how far two identical plates nest). Only needed by plate
+    # stackers (e.g. the Agilent BenchCel); leave unset (None) otherwise.
+    stacking_z_height=39.0,  # measured
     ...
   )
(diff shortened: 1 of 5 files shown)


rickwierenga:
thanks!

PR #1114, opened 2026-06-24
fix(keyence): read full barcode reply and release the read beam (LOFF)

### LINKED ISSUE #1101: `KeyenceBarcodeScannerBackend.scan_barcode()` truncates barcodes to 1 character and never turns the  beam off (no `LOFF`)
`scan_barcode()` returns only the **first character** of a barcode. `send_command()` calls
  `self.io.read()` with no `num_bytes`, and `Serial.read` defaults to 1 byte — no loop, no
  terminator scan — so the reader's `\r`-terminated reply is cut to one byte. It also sends
  `LON` but never `LOFF`, leaving the beam latched on after the call.

  Verified on hardware (Keyence BL-600HA, Code 128 label `Test`):
  - PLR `scan_barcode()` → `data='T'`
  - Same reader driven directly (`LON` → read to `\r` → `LOFF`) → `Test`

  **Source** (`barcode_scanners/keyence/keyence_backend.py`, also on `master`):
  ```python
  async def send_command(self, command: str) -> str:
      await self.io.write((command + "\r").encode(...))
      response = await self.io.read()   # no num_bytes → 1 byte (Serial.read default)
      return response.decode(...).strip()

  async def scan_barcode(self) -> Barcode:
      data = await self.send_command("LON")   # never followed by LOFF
      ...

  Fix: accumulate bytes until \r instead of a single-byte read, and send LOFF after
  the reply (in a finally). PR #1028's generic SerialBarcodeScanner already reads
  CR/LF-terminated lines correctly — good prior art for the pattern.

dinukajayalath:
Fixes #1101

**Problem:**

KeyenceBarcodeScannerBackend.scan_barcode() returns only the first character of a barcode, and leaves the reader's beam latched on:

- send_command() calls self.io.read() with no num_bytes, and Serial.read defaults to 1 byte — no loop, no terminator scan so
The reader's \r-terminated reply is truncated to one character.
- scan_barcode() sends LON (which latches the read beam on) but never the matching LOFF, so the beam stays on after the call.

Verified on hardware (Keyence BL-600HA, Code 128 label Test): PLR returned data='T'; the same reader driven directly (LON →read to \r → LOFF) returned Test.

**Fix:**

- send_command accumulates bytes until \r, stopping at the terminator or at the first byte-less read (the port timeout=1
elapsed with no reply — e.g. no barcode present, a valid empty result). Replies now come back whole, and a present barcode returns the instant the terminator lands instead of blocking on a fixed byte count.
- scan_barcode sends LOFF in a finally block to release the beam, whether the read succeeded or raised. LOFF runs through a try/except that logs rather than propagates, so a LOFF failure can't mask the scan's result or error.

Tests

keyence_backend_tests.py. Each fails against pre-fix code and passes after:
- send_command: accumulates to \r, stops at the terminator, treats a byte-less read as an empty reply, writes \r-terminated
commands.
- scan_barcode: sends LOFF after LON on success / NG / ERR99; a failing LOFF masks neither a successful result nor a reader error.

Glad to make any change upon reviewing.

### DIFF
--- a/pylabrobot/barcode_scanners/keyence/keyence_backend.py
+++ b/pylabrobot/barcode_scanners/keyence/keyence_backend.py
@@ -73,20 +73,37 @@ class KeyenceBarcodeScannerBackend(BarcodeScannerBackend):
       )
 
   async def send_command(self, command: str) -> str:
-    """Send a command to the barcode scanner and return the response.
-    Keyence uses carriage return \r as the line ending by default."""
+    """Send a command and return its reply, accumulated byte-by-byte up to the \r terminator.
+
+    Replies are bare \r-terminated and silence is a valid empty reply (e.g. no barcode), so
+    stop at \r or the first byte-less read rather than reading a fixed byte count."""
 
     await self.io.write((command + "\r").encode(self.serial_messaging_encoding))
-    response = await self.io.read()
-    return response.decode(self.serial_messaging_encoding).strip()
+    buf = bytearray()
+    while True:
+      chunk = await self.io.read()
+      if not chunk:
+        break  # port timeout elapsed with no byte: reply done, or none coming
+      buf.extend(chunk)
+      if chunk == b"\r":
+        break
+    return buf.decode(self.serial_messaging_encoding).strip()
 
   async def stop(self):
     await self.io.stop()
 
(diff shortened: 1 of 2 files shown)


rickwierenga:
thanks!

PR #1116, opened 2026-06-24
fix(liconic): wait for the z-lift to reach the read position before scanning a barcode

### LINKED ISSUE #1115: Liconic scan_barcode() triggers scan before z-lift finishes moving — reads wrong level's barcode
**Summary**

`ExperimentalLiconicBackend.scan_barcode()` sends the ST 1910 move command and then immediately calls `self.barcode_scanner.scan()` without waiting for the move to finish. The laser latches on while the lift is still travelling, so it decodes whatever plate it passes, not the requested level (plate room).

On an STX220 with plates at levels 22 and 10, scanning level 5 returns level 10's barcode (read as the lift descends past it).

The sibling method read_barcode_inline() and every other motion method (open_door, take_in_plate, move_position_to_position,…) already pair their ST command with `await self._wait_ready()`. `scan_barcode()` is the only one missing it.

https://github.com/PyLabRobot/pylabrobot/blob/00841b817dbfece5e189ea6e11506af28b13f85b/pylabrobot/storage/liconic/liconic_backend.py#L563-L579

**Fix:**
await self._send_command("ST 1910")
await self._wait_ready()              # <-- missing
barcode = await self.barcode_scanner.scan()

 Will open a PR with the fix

dinukajayalath:
Summary:
## What

`ExperimentalLiconicBackend.scan_barcode()` sent the `ST 1910` move command to position the z-lift at the read height and then immediately triggered the scan, without waiting for the move to finish. The beam latched on while the lift was
still travelling, so it decoded whichever plate it passed en route instead of the requested level — e.g. on an STX220 with plates at levels 22 and 10, Scanning level 5 returned level 10's barcode.

This adds `await self._wait_ready()` between `ST 1910` and the scan, matching every other motion-issuing method in the backend (`open_door`, `close_door`, `initialize`, `take_in_plate`, `move_position_to_position`) and the sibling `read_barcode_inline()`, which already pair their `ST` command with `_wait_ready()`. `scan_barcode()` was the only one missing it. 

Fixes #1115

## Test

Added `TestScanBarcode` to `liconic_backend_tests.py`: it asserts `ST 1910` is sent and `_wait_ready` is awaited. Verified it fails against the pre-fix code. 

Also verified on hardware (STX220): plates loaded at two levels, scanning a lower third level now returns the requested level's barcode rather than one the lift passes on the way down.

Happy to make any changes upon reviewing

### DIFF
--- a/pylabrobot/storage/liconic/liconic_backend.py
+++ b/pylabrobot/storage/liconic/liconic_backend.py
@@ -573,6 +573,7 @@ class ExperimentalLiconicBackend(IncubatorBackend):
     await self._send_command(f"WR DM25 {pos_num}")  # plate
     await self._send_command(f"WR DM5 {n}")  # plate position in carousel
     await self._send_command("ST 1910")  # move shovel to barcode reading position
+    await self._wait_ready()  # block until the lift reaches the read position
 
     barcode = await self.barcode_scanner.scan()
     logger.info(f"Scanned barcode: {barcode.data}")
--- a/pylabrobot/storage/liconic/liconic_backend_tests.py
+++ b/pylabrobot/storage/liconic/liconic_backend_tests.py
@@ -371,6 +371,24 @@ class TestInitialize(unittest.IsolatedAsyncioTestCase):
     self.backend._wait_ready.assert_awaited()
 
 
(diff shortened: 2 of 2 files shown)


rickwierenga:
thanks!

Issue #1113, opened 2026-06-24
Sequential/stacking ('Stacking Access') storage abstraction backed by ResourceStack


c-reiter:
## Context

The storage user guide already defines two retrieval patterns — **Stacking Access (Sequential)** vs **Random Access** — and explicitly puts the Agilent BenchCel 4R, Labware MiniHub, and Lab Services PlateCarousel in the *sequential* column (`docs/user_guide/01_material-handling/storage/storage.rst`).

In code, though, that distinction doesn't exist yet: `Incubator` is the only storage frontend, and it is fundamentally **random-access** — it takes `racks: List[PlateCarrier]` of addressable `PlateHolder` sites and exposes `find_random_site` / `find_smallest_site_for_plate` / `take_in_plate(site)`. Every backend (`CytomatBackend`, LiCONiC, Heraeus, and now `BenchCel4RBackend`) subclasses `IncubatorBackend`.

## The mismatch (surfaced while adding the BenchCel, #1109)

A BenchCel stacker is a single-ended **LIFO stack of nesting plates** — the firmware only addresses whole stackers (downstack/upstack at one end); there is no slot addressing. To fit the `Incubator` model, #1109 currently fakes each stacker as a 20-site `PlateCarrier` where only the stacker identity is ever used and the sites are cosmetic. The frontend therefore advertises random-access operations (`find_random_site`, placing into an arbitrary site) that the hardware cannot honor.

## Opportunity

`ResourceStack(direction="z")` already models exactly this: single-ended LIFO access (it only lets you unassign the top item), and — since #1110 (`Plate.stacking_z_height`) and #1112 — it nests plates by their stacking pitch, which is precisely a stacker's geometry. #1109 also now maps `Plate.stacking_z_height` → BenchCel `StackingThickness`.

## Proposal / question for discussion

Would you be open to a **sequential/stacking storage abstraction** (e.g. a `Stacker` frontend, or a shared storage base) backed by `ResourceStack`, exposing downstack/upstack (LIFO) semantics instead of random-access site selection? It would model the documented "Stacking Access" category directly and would also fit the MiniHub / PlateCarousel.

Open design questions I'd love your steer on, @rickwierenga:
- A new parallel frontend (`Stacker`) vs. generalizing the existing storage frontend to support both access patterns.
- Whether to introduce a `StorageBackend` base above `IncubatorBackend`, or keep backends as-is.
[...]

PR #1119, opened 2026-06-25
`Stacker`: sequential ("stacking access") storage capability

c-reiter:
## Summary

Adds a `Stacker` capability for **sequential / LIFO plate storage** - the code counterpart to the "Stacking Access (Sequential)" category that's already described in the storage user guide but had no implementation (today every storage device is forced through the random-access `Incubator`).

Design per the discussion in #1113. @rickwierenga preferred a **composable capability** over a split-frontend hierarchy, with the **transfer position as part of the stacker**, so this:
- is a standalone capability wrapping a `StackerBackend` (no shared `StorageBackend` base, no changes to `Incubator`);
- owns the transfer position (the "loading tray", borrowing the incubator's term).

## What's here

- **`Stacker(Machine, Resource)`** - holds one or more single-ended LIFO stacks (each a `ResourceStack(direction="z")`) plus a loading tray.
  - `downstack(stack)` → moves the accessible (top) plate onto the loading tray and returns it.
  - `upstack(stack, plate=None)` → moves a plate from the loading tray onto a stack (defaults to whatever is on the tray).
  - LIFO is enforced by `ResourceStack` (only the top plate is accessible); stack height/positions come from `Plate.stacking_z_height` via the nesting added in #1112.
  - Helpers: `get_accessible_plate`, `get_stack_by_plate_name`, `summary`.
- **`StackerBackend`** — minimal device interface: `set_stacks`, `downstack`, `upstack`. Deliberately **no** door/temperature/shaking (a stacker is not an incubator).
- **`StackerChatterboxBackend`** — no-op backend for tests/demos.
- 10 tests (LIFO order, nesting height, empty-stack / occupied-tray / no-plate errors, foreign-stack rejection, etc.).

## Intended consumers

The Agilent BenchCel (#1109) and the HighRes MicroServe. Migrating the BenchCel from its current faked-`PlateCarrier` `Incubator` model onto this capability is a planned follow-up.

## Notes / follow-ups

- Serialization round-trip is intentionally not implemented yet — it needs `ResourceStack` serialization support (`ResourceStack.__init__` takes `direction`, not `size_*`, so it doesn't round-trip through the generic path). `Stacker.serialize` works; `Stacker.deserialize` raises `NotImplementedError` for now (and resolves the `Machine`/`Resource` MRO ambiguity).
- Open design questions from #1113 still worth your call: a generic "transfer position" abstraction shared with incubators, and whether carousels fold in later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - In-process `MicroSpinMockServer` (`pylabrobot.centrifuge.highres.mock_server`) that faithfully emulates the MicroSpin's wire protocol -- including the firmware's "`status` blocks until the spindle has stopped" semantics and the low-G spin-down-detection hang -- usable as a Python async context manager or runnable as a script (`python -m pylabrobot.centrifuge.highres.mock_server`) for `nc`/`telnet` debugging.
 - `MicroSpinBackend.reset()` recovery helper that issues `abort` -> `clearbuttonabort` -> `status`, using the last as the gate that genuinely confirms the rotor has stopped.
 - User guide notebook for the MicroSpin (`docs/user_guide/01_material-handling/centrifuge/highres_microspin.ipynb`).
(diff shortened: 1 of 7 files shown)


rickwierenga:
high level:

interface looks good overall

it would be good to target v1 with this and make Stacker a `Capability` (v1) rather than a `Machine` (v0)

c-reiter:
Superseded by #1142, which implements this as a v1 `Capability` (per @rickwierenga's feedback here and on #1113) sharing a `LoadingTrayRetrieval` base with `AutomatedRetrieval`, rather than a v0 `Machine`. Closing in favour of the v1 version.

PR #1128, opened 2026-06-28
feat(STAR): parallelize firmware commands across modules

rickwierenga:
## Summary

Moves the STAR firmware command lock out of `STARPIPBackend` and into `STARDriver`, so it coordinates **every** module instead of only the pipetting channels. Previously the 96-head (`H0`) and X-arm (`X0`) called `driver.send_command` directly and bypassed the lock entirely.

The lock now models the firmware's master/slave structure:

- **`C0` master commands run exclusively** — no other command is in flight.
- **Slave-module commands (`Px` / `H0` / `X0`) overlap across modules** but serialize within a module (no two `H0`, two `X0`, or two `Px` at once). Implemented as a readers/writer gate (`slave_module()` vs `c0()`) plus a per-module mutex.
- **Read-only request commands (`R*`) are lock-free on every module** — they take no lock and run fully in parallel, even alongside a same-module move or an exclusive `C0` master command.

The practical payoff: an X-arm (`X0:XP`) move and a 96-head Y (`H0:YA`) move can now run concurrently, so the 96-head can travel a **diagonal** — something the exclusive `C0:EM` combined move can never be parallelised into.

## Changes

- `star/driver.py`: `_FirmwareLock` (per-module mutexes + master/slave gate) and a `send_command` override that applies it.
- `star/pip_backend.py`: dropped the old PIP-only lock; `send_command` now just delegates to the driver.
- `docs/cookbook/star_parallel_head96_xy.py`: runnable sample that overlaps `X0` + `H0` to sweep the 96-head along a diagonal (with acceleration-level matching for a clean line).
- `CHANGELOG.md`: entries under Unreleased.

## Testing

- Full STAR suite green (`pylabrobot/hamilton/liquid_handlers/star/`, 117 passed).
- `ruff check` / `ruff format --check` clean on changed files; no new `mypy` errors.
- Verified on hardware (STAR + CoRe 96-head): concurrent `X0 || H0` moves the head ~400 mm in X and ~350 mm in Y together in ~2.3 s vs ~4.4 s sequential — a clean diagonal.

### DIFF
--- a/pylabrobot/hamilton/liquid_handlers/star/driver.py
+++ b/pylabrobot/hamilton/liquid_handlers/star/driver.py
@@ -6,6 +6,7 @@ import enum
 import logging
 import math
 import re
+from contextlib import AsyncExitStack, asynccontextmanager
 from dataclasses import dataclass, field
 from typing import Any, List, Optional, Tuple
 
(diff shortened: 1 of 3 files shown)


rickwierenga:
Sample script for driving the 96-head along a diagonal by overlapping an `X0` (X-arm) move with an `H0` (head Y) move — kept out of the repo, pasting here for reference.

```python
"""Move the STAR 96-head diagonally by overlapping two firmware modules.

The STAR firmware exposes independent module controllers. The 96-head's X is driven by the
X-arm (``X0:XP``, :meth:`STARDriver.experimental_x_arm_move`) and its Y by the head module
(``H0:YA``, :meth:`STARHead96Backend.move_y`). Because they live on different modules, the
driver's firmware lock lets them run concurrently: each is a *slave-module* command, so they
overlap as long as they target different modules (and no exclusive ``C0`` master command is
in flight). Issuing both in a single ``asyncio.gather`` therefore moves the head along a
diagonal -- something the exclusive ``C0:EM`` combined move can never be parallelised into.

``move_y`` is acceleration-ramp-limited (raising its speed barely shortens the move) and
``experimental_x_arm_move`` exposes only an ``acceleration_level`` (1-5), not a cruise speed.
So to get a clean diagonal we time each leg once and pick the X ``acceleration_level`` whose
duration best matches the Y leg, then overlap them.

Run against a real STAR with an X-arm and a CoRe 96-head installed:

    python star_parallel_head96_xy.py
"""

import asyncio
import time

from pylabrobot.hamilton.liquid_handlers.star.star import STAR

# Endpoints in each axis' own frame. X is the X0:XP frame; Y is the head Y frame.
X_A, X_B = 250.0, 650.0  # mm, X-arm range [90, 1350]
Y_A, Y_B = 150.0, 500.0  # mm, 96-head Y range [93.75, 562.5]


async def _timed(coro) -> float:
  start = time.perf_counter()
  await coro
  return time.perf_counter() - start


async def main() -> None:
  star = STAR()
  try:
    await star.setup()
    driver = star.driver
    head96 = driver.head96
    if head96 is None:
      raise RuntimeError("This sample needs a CoRe 96-head installed.")

    async def move_x(x: float, accel: int = 3) -> None:
      await driver.experimental_x_arm_move(x=x, acceleration_level=accel)  # X0:XP

    async def move_y(y: float) -> None:
      await head96.move_y(y=y)  # H0:YA

    # Always travel at a safe Z before any lateral motion.
    await head96.move_to_z_safety()

    # Time the (acceleration-limited) Y leg -- our reference duration.
    await move_y(Y_A)
    t_y = await _timed(move_y(Y_B))
    print(f"H0 Y leg: {t_y:.2f} s")

    # Time the X leg at each acceleration level and pick the closest match to the Y leg.
    leg_times = {}
    for accel in range(1, 6):
      await move_x(X_A, accel=accel)
      leg_times[accel] = await _timed(move_x(X_B, accel=accel))
      print(f"X0 leg @ acceleration_level={accel}: {leg_times[accel]:.2f} s")
    best_accel = min(leg_times, key=lambda a: abs(leg_times[a] - t_y))
    print(f"=> matching X0 acceleration_level={best_accel} ({leg_times[best_accel]:.2f} s)")

    # Overlap the two slave-module commands: the head sweeps a diagonal.
    await move_x(X_A, accel=best_accel)
    await move_y(Y_A)
    before = await head96.request_position()
    t_diag = await _timed(asyncio.gather(move_x(X_B, accel=best_accel), move_y(Y_B)))
    after = await head96.request_position()

    print(f"\ndiagonal (X0 || H0): {t_diag:.2f} s")
    print(f"head moved: x {before.x:.1f} -> {after.x:.1f}   y {before.y:.1f} -> {after.y:.1f}")
  finally:
    await star.stop()


if __name__ == "__main__":
  asyncio.run(main())
```

PR #1130, opened 2026-06-29
`OT2RobotGeometry`: add static OT-2 robot geometry with per-mount reach checks

BioCam:
Adds `OT2RobotGeometry`, a frozen dataclass capturing the OT-2's fixed geometry in the robot frame: the gantry working extents, per-mount nozzle offsets, the head8 per-channel y-offsets, and the partial-tip body paddings. Because every OT-2 Standard shares this geometry it is a constant rather than something probed from the device - the key difference from Hamilton STAR, whose installed configuration must be read from firmware.

It also provides a centre-based `can_reach_position(mount, position, channel_offset)` predicate (mirroring `STARBackend.can_reach_position`) that works for single and multi-channel alike: a single pipette uses `channel_offset=0`, a head8 passes each nozzle's offset from `channel_y_offsets()`. Near the front/back limits only a subset of a multi's nozzles reach, which is why an edge column is picked up partially rather than all 8.

Pure additive: a new module, its tests, and one export. No behaviour change. This is the foundation for wiring reach checks into the OT-2 backend operations, which lands in a follow-up.

### DIFF
--- a/pylabrobot/resources/opentrons/ot2_geometry.py
+++ b/pylabrobot/resources/opentrons/ot2_geometry.py
@@ -1,0 +1,87 @@
+import dataclasses
+from typing import List, Tuple
+
+from pylabrobot.resources.coordinate import Coordinate
+
+
+@dataclasses.dataclass(frozen=True)
+class OT2RobotGeometry:
+  """Static geometry of an Opentrons OT-2, expressed in the robot frame.
+
+  Every OT-2 Standard shares this geometry, so it is a constant rather than something probed from
+  the device. This is the key difference from Hamilton STAR hardware, whose installed configuration
+  varies per machine and must be read from firmware; the only OT-2 data worth discovering at runtime
+  is which pipettes are mounted, not where the deck is.
+
+  The robot frame has its origin at the front-left corner of slot 1, with +x to the right, +y to
+  the back, and +z up. All distances are in mm.
+  """
+
+  # Gantry working area reachable by the reference (right) mount.
+  extents: Coordinate = dataclasses.field(default_factory=lambda: Coordinate(446.75, 347.5, 0.0))
+
+  # Partial-tip body clearance: how far a pipette's bounding box may sit beyond the deck extents at
+  # each edge. These bound the pipette body in partial nozzle configurations only; they do not
+  # constrain a fully-configured single or multi-channel pipette and are not used by the
+  # single-channel reach check below.
+  padding_front: float = 31.89
+  padding_rear: float = -35.91
+  padding_left: float = 0.0
+  padding_right: float = 0.0
+
+  # Nozzle offset of each mount from the gantry carriage; the right mount is the reference.
+  left_mount_offset: Coordinate = dataclasses.field(
+    default_factory=lambda: Coordinate(-34.0, 0.0, 0.0)
+  )
+  right_mount_offset: Coordinate = dataclasses.field(
+    default_factory=lambda: Coordinate(0.0, 0.0, 0.0)
+  )
+
+  def mount_offset(self, mount: str) -> Coordinate:
+    """Nozzle offset of ``mount`` ("left" or "right") from the gantry carriage."""
+    if mount == "left":
+      return self.left_mount_offset
+    if mount == "right":
+      return self.right_mount_offset
+    raise ValueError(f"mount must be 'left' or 'right', got {mount!r}")
+
+  def single_channel_reach(self, mount: str) -> Tuple[float, float, float, float]:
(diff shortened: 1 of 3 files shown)


pylabrobot/resources/opentrons/ot2_geometry.py
```diff
@@ -0,0 +1,87 @@
...
+
+@dataclasses.dataclass(frozen=True)
+class OT2RobotGeometry:
+  """Static geometry of an Opentrons OT-2, expressed in the robot frame.
+
+  Every OT-2 Standard shares this geometry, so it is a constant rather than something probed from
+  the device. This is the key difference from Hamilton STAR hardware, whose installed configuration
+  varies per machine and must be read from firmware; the only OT-2 data worth discovering at runtime
+  is which pipettes are mounted, not where the deck is.
+
```
rickwierenga, line 14: the code should only describe itself, not be a story comparing it to other machines (claude will often do this)


```suggestion
  Every OT-2 Standard shares this geometry, so it is a constant rather than something probed from
  the device. The only OT-2 data worth discovering at runtime is which pipettes are mounted, not
  where the deck is.
```

pylabrobot/resources/opentrons/ot2_geometry.py
```diff
@@ -0,0 +1,87 @@
...
+  @staticmethod
+  def channel_y_offsets(num_channels: int = 8, channel_pitch: float = 9.0) -> List[float]:
+    """Y offset of each channel from the head center, in mm, for a linear multi-channel head.
+
+    Indexed back-to-front to match the Opentrons nozzle map: index 0 is the back-most nozzle (row
+    A, the primary/critical point) at the largest +y, descending to the front-most nozzle at -y.
+    Defaults describe the head8: 8 channels at 9 mm pitch spanning the head center by +-31.5 mm.
+    """
+    half = (num_channels - 1) / 2
+    return [(half - i) * channel_pitch for i in range(num_channels)]
```
rickwierenga: given that the 8 channels are fixed, how to interpret `num_channels`?

rickwierenga: is this for partial tip pickup scenarios?

BioCam: as a simple start, I recommend here to index based on coordinates and usual channel indexing:
(1) left to right in x
(2) back to front in y

Each channel is still an independent unit that _can_ act with others but doesn't have to.
That way a tip in tip_spot "C1" e.g. can still be picked up by an 8-channel head's channel_2 and act like a standalone channel, giving maximum power and flexibility.

In v1, with a fully developed `head8` capability I would like to build the a "per-head channel indexing" system.

But this needs to be tested in steps and in v0 we can use the Opentrons OT-2 as a testing bed for it.

rickwierenga: > That way a tip in tip_spot "C1" e.g. can still be picked up by an 8-channel head's channel_2 and act like a standalone channel, giving maximum power and flexibility.

so the answer to the question is yes?

> But this needs to be tested in steps and in v0 we can use the Opentrons OT-2 as a testing bed for it.

this is better done in v1

BioCam: Yes this is for partial tip pickup

I agree getting this straight into v1 would be best but since v1 is not ready yet for production and we need this urgently im happy to develop in v0 and v1 in parallel, I.e. merge this and learn from it how to submit updates to v1 in parallel

BioCam: In regards to 8 channels: yes i agree that doesn't make sense to make mutable, will update

pylabrobot/resources/opentrons/ot2_geometry.py
```diff
@@ -0,0 +1,86 @@
...
+    """Whether a nozzle displaced ``channel_offset`` mm in y from the head center can reach
+    ``position`` in x and y on ``mount``.
+
+    Mirrors ``STARBackend.can_reach_position``: a pure predicate over the reachable region, here the
+    single-channel deck extents for the mount. For a single channel use the default
+    ``channel_offset=0``. For a head8, pass each channel's offset from :meth:`channel_y_offsets`; the
+    center must land within the extents, so near the front/back limits only a subset of channels can
+    reach, which is why an edge column is picked up partially rather than all 8. Z is not bounded by
+    deck extents and is checked separately by the motion layer.
+    """
```
rickwierenga: still has star comparison

see https://github.com/PyLabRobot/pylabrobot/pull/1130#discussion_r3507911694

PR #1131, opened 2026-06-29
`OTDeck`: model slots as resource holders and render them in the visualizer

BioCam:
Models the OT-2 deck's 12 slots as `ResourceHolder` children inset onto the deck plate corner, and updates the visualizer to draw those holders.

Modelling each slot as a holder gives it a real location and a single child (its labware), so placement, serialization, and slot lookup all go through the holder instead of ad-hoc deck math. The visualizer then renders each slot from its holder - an occupied slot shows its labware, an empty one shows the slot number on top - and finds the deck by type (`instanceof Deck`) rather than a hardcoded name.

Foundation of the OT-2 series; depends on nothing else in it.

### DIFF
--- a/pylabrobot/visualizer/lib.js
+++ b/pylabrobot/visualizer/lib.js
@@ -713,27 +713,29 @@ function getSnappingResourceAndLocationAndSnappingBox(resourceToSnap, x, y) {
     }
   }
 
-  // Check if the resource is in the OT Deck.
+  // Check if the resource is in the OT Deck. The slots are ResourceHolder children of the deck, so
+  // their positions and sizes come from the serialized resource rather than being duplicated here.
   if (deck.constructor.name === "OTDeck") {
-    const siteWidth = 128.0;
-    const siteHeight = 86.0;
-
-    for (let i = 0; i < otDeckSiteLocations.length; i++) {
-      let siteLocation = otDeckSiteLocations[i];
+    for (const holder of deck.children) {
+      if (!holder.location || !(holder.name && /slot_\d+$/.test(holder.name))) {
+        continue;
+      }
+      const siteX = deck.location.x + holder.location.x;
+      const siteY = deck.location.y + holder.location.y;
       if (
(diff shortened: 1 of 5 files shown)


pylabrobot/visualizer/lib.js
```diff
@@ -1350,72 +1355,78 @@ class VantageDeck extends Deck {
...
+  { x: 115.65, y: 158.53 },
+  { x: 248.15, y: 158.53 },
+  { x: 380.65, y: 158.53 },
+  { x: 115.65, y: 249.03 },
+  { x: 248.15, y: 249.03 },
+  { x: 380.65, y: 249.03 },
+  { x: 115.65, y: 339.53 },
+  { x: 248.15, y: 339.53 },
+  { x: 380.65, y: 339.53 },
 ];
```
rickwierenga: why does the visualizer still need this? can we not load it from the resource definition?

pylabrobot/visualizer/lib.js
```diff
@@ -1350,72 +1355,78 @@ class VantageDeck extends Deck {
...
+  }
 
-      // Add a text label in the site
+  draw(layer) {
+    super.draw(layer);
+    // Holder children are drawn after the main shape, so a slot number placed in the main shape is
+    // hidden. Add the number to the deck group last (on top) but only for empty slots, so the
+    // labware in an occupied slot is shown instead of being covered by the number.
+    const width = 128.0;
+    const height = 86.0;
```
rickwierenga: this is also duplicated from python. can we not load it?

pylabrobot/visualizer/lib.js
BioCam: You're right, it doesn't - remnant of the fake slots that were here before this PR

pylabrobot/visualizer/lib.js
BioCam: good point, implemented it now

rickwierenga:
this has conflicts but otherwise good to merge

PR #1132, opened 2026-06-29
`OpentronsOT2ChatterboxBackend`: a dry-run OT-2 backend with I/O logging, via a transport seam

BioCam:
Adds `OpentronsOT2ChatterboxBackend`, a dry-run backend for the OT-2 that runs the real backend logic with no hardware and no `ot_api` library, mirroring how `STARChatterboxBackend` dry-runs `STARBackend`. Unlike `OpentronsOT2Simulator` (which overrides the high-level methods with single-channel stubs), the chatterbox replaces only the transport, so pipette selection, tip/volume tracking, and the per-operation wire calls all run unchanged.

To enable that, the backend routes every `ot_api` call - including the run-cancel calls in `stop()` - through a single `self._ot` transport handle. The real handle wraps `ot_api` in a transparent `_IOLogger` proxy that logs every HTTP call at `LOG_LEVEL_IO`, like other backends get from their io object; the chatterbox swaps the handle for a recorder that logs the same way, so a dry run captures the same wire trace as a real run.

Characterization tests pin the existing wire calls (home / stop-cancel / list_connected_modules / addressable-area trash drop) so the seam refactor cannot change them silently; chatterbox tests run a full protocol with no `ot_api` installed and assert the recorded calls.

Foundation of the OT-2 series; depends on nothing else in it.

### DIFF
--- a/pylabrobot/liquid_handling/backends/opentrons_backend.py
+++ b/pylabrobot/liquid_handling/backends/opentrons_backend.py
@@ -31,9 +31,6 @@ from pylabrobot.resources.tip_rack import TipRack
 try:
   import ot_api
 
-  # for run cancellation
-  import ot_api.requestor as _req
-
   USE_OT = True
 except ImportError as e:
   USE_OT = False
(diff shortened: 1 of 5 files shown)


pylabrobot/liquid_handling/backends/opentrons_backend.py
```diff
@@ -44,6 +44,38 @@
...
+  The OT-2 talks HTTP through ``ot_api`` rather than a pylabrobot.io transport, so
+  this wrapper gives it the same wire-level logging every other backend gets from
+  its io object. Submodules (``lh``, ``health``, ...) are wrapped recursively;
+  plain attributes (e.g. ``run_id``) pass through untouched.
+  """
+
+  def __init__(self, target: Any, prefix: str = ""):
+    object.__setattr__(self, "_target", target)
+    object.__setattr__(self, "_prefix", prefix)
+
```
rickwierenga, line 63: why is it like that and not:

```
self._target = target
self._prefix = prefix
```

?

BioCam: overcomplicated nonsense, thank you for catching - simplified now

PR #1135, opened 2026-06-30
`OpentronsOT2Backend`: support the 8-channel pipette

BioCam:
The backend modeled every mount as a single channel, so the OT-2 8-channel pipette could not be driven correctly. This models a multi mount as 8 channels - anchored at channel 0 (back, row A) and filling toward row H at the 9 mm nozzle pitch - and issues one wire command per column for multi-channel pickup, aspirate, and dispense, matching the hardware, which acts on a whole column at once rather than channel by channel. Single-channel pipettes are unchanged: they stay one channel and one command per operation.

Tests cover the channel model (mount -> channel count, channel-0 anchoring) and the per-column wire grouping, run as a dry run against the chatterbox so no hardware or `ot_api` is needed.

Stacked on #1132, whose `OpentronsOT2ChatterboxBackend` is the dry-run harness these tests use. The self-contained change here is the three commits above that base.

### DIFF
--- a/pylabrobot/liquid_handling/backends/opentrons_backend.py
+++ b/pylabrobot/liquid_handling/backends/opentrons_backend.py
@@ -149,9 +149,31 @@ class OpentronsOT2Backend(LiquidHandlerBackend):
     if not skip_home:
       await self.home()
 
+  @staticmethod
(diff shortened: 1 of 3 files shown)


rickwierenga:
does not really apply anymore after v1 rewrite, so closing this

PR #1137, opened 2026-06-30
`Container`: forward volume-tracker changes to state updates; format Visualizer volume readouts

BioCam:
A `Container`'s `VolumeTracker` changes were applied internally but never broadcast. `Container.__init__` creates the tracker but never calls `tracker.register_callback(self._state_updated)`, so a bare container such as a `Trough` updates its own volume yet notifies no subscribers.

`Well` and `Tube` each register that callback in their own `__init__`, so plates and tube racks already update live; every other direct `Container` subclass was missed. The most visible symptom is in the Visualizer: a `Trough`'s liquid level stays frozen at its seeded volume even as it drains, because the tracker's `_callback` is never set. The tracking math was always correct - only the notification was missing.

- `Container.__init__` now calls `self.tracker.register_callback(self._state_updated)` right after creating the tracker. `super().__init__()` runs first, so `_state_updated` is already bound, and this wires every `Container` subclass in one place.
- `Well` and `Tube` drop their now-redundant `register_callback` calls and inherit the wiring through `super().__init__()`. `register_callback` is a plain setter, so the registered callback is identical and behaviour is unchanged.
- The Visualizer formats volume readouts through a shared `formatVolume` helper: at most 2 decimals with no trailing zeros (so a draining `Trough` reads `327.2`, not `327.20000000004626`) and thousands separators on `volume` and `max_volume`, so values above 1000 read as e.g. `60,000`.

Behaviour: no change for `Well`, `Tube`, or plates (the same callback, now set one level up); `Trough` and other bare containers emit state updates on volume change for the first time, and the Visualizer renders their level live and cleanly formatted.

Tests: adds a regression test asserting a `Container` forwards tracker changes to its state-update callbacks; the container, well, and visualizer suites pass.

### DIFF
--- a/pylabrobot/resources/container.py
+++ b/pylabrobot/resources/container.py
@@ -79,6 +79,9 @@ class Container(Resource):
 
     self.max_volume = max_volume or (size_x * size_y * size_z)
     self.tracker = VolumeTracker(thing=f"{self.name}_volume_tracker", max_volume=self.max_volume)
+    # Notify state-update subscribers (e.g. the Visualizer) on volume changes; without this
+    # a bare Container like a Trough updates its volume internally but never broadcasts it.
+    self.tracker.register_callback(self._state_updated)
     self._compute_volume_from_height = compute_volume_from_height
     self._compute_height_from_volume = compute_height_from_volume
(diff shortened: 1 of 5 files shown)


rickwierenga:
thanks!

PR #1140, opened 2026-07-01
feat(vspin): grip steps

rkaptur:
Add grip_steps parameter to Access2Backend.load and unload

load() and unload() were only tested with a 1 cm plate. For other plate types the gripper may not seat firmly enough with a single close command. This PR adds a grip_steps: int = 1 parameter (range 1–4) to both methods that controls how many times the close gripper command (1105000a00420700010000803f02008c64) is sent after the arm descends to the plate. The default of 1 is fully backwards-compatible.

### DIFF
--- a/pylabrobot/centrifuge/vspin_backend.py
+++ b/pylabrobot/centrifuge/vspin_backend.py
@@ -120,14 +120,14 @@ class Access2Backend(LoaderBackend):
     await self.send_command(bytes.fromhex("1105000a004607000100000000020015fd"))
     await self.send_command(bytes.fromhex("1105000e00440b0000000040400000204102007d82"))
 
-  async def unload(self, grip_steps: int = 1):
+  async def unload(self, grip_steps: Literal[1, 2, 3, 4] = 1):
     """Unload a plate from the centrifuge.
 
     Args:
       grip_steps:  Number of steps taken to tighten the grip.
         Higher values may improve grip for certain plate types.
     """
-    if not 1 <= grip_steps <= 4:
+    if grip_steps not in (1, 2, 3, 4):
       raise ValueError("grip_steps must be between 1 and 4")
     logger.debug("[loader] unload")
 
--- a/pylabrobot/centrifuge/vspin_backend.py
+++ b/pylabrobot/centrifuge/vspin_backend.py
@@ -94,14 +94,14 @@ class Access2Backend(LoaderBackend):
     logger.debug("[loader] open")
     await self.send_command(bytes.fromhex("1105000a0042070001000080bf0200b73e"))
 
-  async def load(self, grip_steps: int = 1):
+  async def load(self, grip_steps: Literal[1,2,3,4] = 1):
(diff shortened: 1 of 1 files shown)


pylabrobot/centrifuge/vspin_backend.py
```diff
@@ -94,8 +94,15 @@ async def open(self):
...
-    """only tested for 1cm plate, 3mm pickup height"""
+  async def load(self, grip_steps: int = 1):
+    """Load a plate into the centrifuge.
+
+    Args:
+      grip_steps: Number of steps taken to tighten the grip.
+        Higher values may improve grip for certain plate types.
+    """
+    if not 1 <= grip_steps <= 4:
+      raise ValueError("grip_steps must be between 1 and 4")
```
rickwierenga:
```suggestion
  async def load(self, grip_steps: Literal[1,2,3,4] = 1):
    """Load a plate into the centrifuge.

    Args:
      grip_steps: Number of steps taken to tighten the grip.
        Higher values may improve grip for certain plate types.
    """
    if not grip_steps not in (1, 2, 3, 4):
      raise ValueError("grip_steps must be between 1 and 4")
```

pylabrobot/centrifuge/vspin_backend.py
```diff
@@ -107,12 +114,21 @@ async def load(self):
       raise LoaderNoPlateError("no plate found on stage")
 
     await self.send_command(bytes.fromhex("1105000a00460700018fc2b540020023dc"))
+    for _ in range(grip_steps):
+      await self.send_command(bytes.fromhex("1105000a00420700010000803f02008c64"))
```
rickwierenga, line 118: lol this firmware api continues to surprise

pylabrobot/centrifuge/vspin_backend.py
```diff
@@ -107,12 +114,21 @@ async def load(self):
...
-    """only tested for 1cm plate, 3mm pickup height"""
+  async def unload(self, grip_steps: int = 1):
+    """Unload a plate from the centrifuge.
+
+    Args:
+      grip_steps:  Number of steps taken to tighten the grip.
+        Higher values may improve grip for certain plate types.
+    """
+    if not 1 <= grip_steps <= 4:
+      raise ValueError("grip_steps must be between 1 and 4")
```
rickwierenga:
```suggestion
  async def unload(self, grip_steps: Literal[1, 2, 3, 4] = 1):
    """Unload a plate from the centrifuge.

    Args:
      grip_steps:  Number of steps taken to tighten the grip.
        Higher values may improve grip for certain plate types.
    """
    if grip_steps not in (1, 2, 3, 4):
      raise ValueError("grip_steps must be between 1 and 4")
```

rickwierenga:
thanks! 1 small suggestion applied in two locations, what do you think

rkaptur:
Good idea, should be fixed now

rickwierenga:
thanks!

PR #1141, opened 2026-07-01
Vspin driver improvements

TheRealDarkLuke:
Add a driver for the old V11 Vspin


rickwierenga:
thanks for the PR!

interestingly, we have vspins with the Agilent and v11 label and both of them work through the existing backend. I wonder if this is not an agilent/v11 difference but rather a firmware version difference which would correlate with agilent/v11 (because Agilent acquired them of course)

the structure of the code looks very similar between the two. Rather than having a completely new backend, why dont we have one backend with a structure like this?

```python
def something():
  if self._model == "agilent":
    cmd = "A"
  elif self._model == "v11":
    cmd = "V"
  else:
    raise
  self.send(cmd)
```

this way we have less duplication and the "Agilent" version can benefit from your improvements as well. It might also give us some more clarity on what exactly changed between versions and if there's an even smaller abstraction

TheRealDarkLuke:
Good catch! I didn't think of that, but it makes perfect sense. 

Fair point about the duplicates—I'll fix that right away.

pylabrobot/centrifuge/vspin_backend_tests.py
```diff
@@ -0,0 +1,105 @@
...
+    self.writes = []
+
+  async def read(self, num_bytes: int) -> bytes:
+    if self.read_chunks:
+      return self.read_chunks.pop(0)
+    return b""
+
+  async def write(self, data: bytes) -> int:
+    self.writes.append(data)
+    return len(data)
```
rickwierenga: why not use unit test mock for this?

pylabrobot/centrifuge/vspin_backend.py
```diff
@@ -302,6 +352,9 @@ async def stop(self):
     await self.configure_and_initialize()
     await self.io.stop()
 
+  def _command(self, name: str) -> bytes:
+    return _VSPIN_COMMANDS[self._command_set][name]
+
```
rickwierenga: this name is a bit confusing. maybe `_get_command_bytes` or sth?

rickwierenga (commented):
thanks for the refactor!

looks like we actually did have this "old" firmware at some point, and then I changed it to this new format to make it work on some other vspins and since it worked on others must have assumed this new one was more universal. That was clearly a mistake, we have too many vspins 😅

still the question stands what is the difference between them? why do you think it's Agilent vs v11?

pylabrobot/centrifuge/vspin_backend_tests.py
```diff
@@ -0,0 +1,105 @@
...
+    packet = _status_packet()
+
+    parsed = VSpinBackend._find_status_packet(b"\x00\xff" + packet + b"\x00")
+
+    assert parsed is not None
+    self.assertEqual(parsed.status, 0x11)
+    self.assertEqual(parsed.current_position, 12070)
+    self.assertEqual(parsed.tachometer, -10)
+    self.assertEqual(parsed.home_position, 6733)
+
```
rickwierenga: this depends on the default parameter in `_status_packet`. ideally we would just test parsing

pylabrobot/centrifuge/vspin_backend_tests.py
TheRealDarkLuke: This is just an educated guess. 
Assuming there is no hardware difference and the distinction is purely firmware-based: 
The "old" firmware is typically found on devices labeled "V11". It is highly unlikely that Agilent devices are running this outdated firmware, as they should all be on a newer version. Therefore, Agilent-labeled devices should work perfectly fine, whereas V11 devices might vary from one another.

The question is: how do we differentiate them? My suggested naming convention wasn't accurate enough. 
To be honest, I'm not very familiar with that device, so I'm open to any good suggestions.

pylabrobot/centrifuge/vspin_backend.py
TheRealDarkLuke: You're right - changed

pylabrobot/centrifuge/vspin_backend.py
```diff
@@ -332,6 +385,17 @@ class _StatusPositionTachometer(ctypes.LittleEndianStructure):
...
+  @staticmethod
+  def _find_status_packet(resp: bytes) -> Optional[_StatusPositionTachometer]:
+    for start in range(max(0, len(resp) - 13)):
+      packet = resp[start : start + 14]
+      if len(packet) < 14 or packet[0] not in _KNOWN_VSPIN_STATUSES:
+        continue
+      if (sum(packet[:-1]) & 0xFF) != packet[-1]:
+        continue
+      return VSpinBackend._StatusPositionTachometer.from_buffer_copy(packet)
+    return None
```
rickwierenga, line 397: this seems a bit fragile as it silently skips "unknown" vspin status responses. How do we know we have all? There has to be some better way of reading status responses. I think they arrive in individual ftdi packets

pylabrobot/centrifuge/vspin_backend.py
```diff
@@ -180,6 +180,46 @@ def _save_vspin_calibrations(device_id, remainder: int):
...
+    raise ValueError("command_set must be 'agilent' or 'old_firmware'")
+  return _VSPIN_COMMAND_SET_ALIASES[normalized]
+
+
+def _with_vspin_checksum(cmd: bytes) -> bytes:
+  """Return ``cmd`` with the final VSpin checksum byte recomputed."""
+  if len(cmd) <= 2 or cmd[0] != 0xAA:
+    return cmd
+  payload = cmd[1:-1]
+  return b"\xaa" + payload + bytes([sum(payload) & 0xFF])
```
rickwierenga, line 222: this strips the first byte and the checksum from the given command, and then adds it back. In the case of the first byte, it is always 0xAA per previous if statement, so removing that is always unnecessary. But I think what we should really do is store the "stem" of each command and have the checksum and 0xAA added automatically

rickwierenga:
closing in favor of #1227 

please let me know if that works for you. it should be pretty complete ;)

TheBirdAttack:
Hi Rick

Yes, absolutely. I was planning to continue working on this, but my tests on the real hardware didn't work out as expected.
Thanks for jumping in and getting this done!

I'll gladly test [#1227](https://github.com/PyLabRobot/pylabrobot/pull/1227) on our device.

Cheers,
Lucas

PR #1149, opened 2026-07-05
Fix README documentation links

TheRealDarkLuke:
On `main` , the link to the centrifuge documentation was broken.
### Centrifuges ([docs](https://docs.pylabrobot.org/stable/user_guide/01_material-handling/centrifuge/_centrifuge.html))

grafik

##
While already fixing that Link, this PR updates **ALL** README documentation links to use `stable` instead of a specific release version. That should keep the links pointing to the current stable docs when new releases are published.

`stable` seems to be a own page, but identical to `0.2.1` 
(Kind of confusing as 0.2.1 is labeled with stable)


rickwierenga:
thanks!

PR #1160, opened 2026-07-08
`Container`: accept lids via a `Liddable` mixin, centred on the top face

BioCam:
`Plate` was the only resource that could hold a lid, and placement was origin-aligned - so a lid that isn't the plate's exact footprint (a trough lid, or a slightly-undersized plate lid) sat in a corner.

Proposes a `Liddable` mixin shared by `Plate` and `Container`, with the lid centred on the parent's top face. (Where the mixin lives is open - it's small and easy to relocate.)

- `lid.py`: `Lid` (moved from `plate.py`, re-exported for back-compat) + `Liddable` (`.lid`, `has_lid()`, centred `get_lid_location`, double-lid and minimum-size guards).
- `Plate(Liddable, ItemizedResource["Well"])` and `Container(Liddable, Resource)` - troughs, tubes, petri dishes and wells inherit lids; `drop_resource` / `move_lid` gate on `Liddable`.

Footprint-matched lids are byte-identical (verified across every `with_lid` plate); off-footprint lids now centre - sub-mm (the one rotated STAR lid test is re-pinned). Rotation alignment is left to the move/arm layer, so the resource stays geometry-only.

Tests: `lid_tests.py` (centred placement, container lids, guards, serialize round-trip); full suite, ruff and mypy clean.

### DIFF
--- a/pylabrobot/resources/lid.py
+++ b/pylabrobot/resources/lid.py
@@ -9,7 +9,7 @@ from .resource import Coordinate, Resource
 # A lid may be modelled up to this much smaller than the resource it covers - its rim sits just
 # inside the outer edge, so real plate lids run a few tenths of a mm under. A larger shortfall means
 # the wrong lid. Used by Liddable.assign_child_resource.
-LID_UNDERSIZE_TOLERANCE_MM = 1.0
+LID_UNDERSIZE_TOLERANCE = 1.0
 
 
 class Lid(Resource):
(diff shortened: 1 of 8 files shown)


pylabrobot/resources/lid.py
rickwierenga, line 1: making `lid` a lazily computed property would really simplify this class

pylabrobot/resources/lid.py
```diff
@@ -0,0 +1,129 @@
...
+        or resource.get_size_y() < self.get_size_y() - LID_UNDERSIZE_TOLERANCE_MM
+      ):
+        raise ValueError(
+          f"Lid '{resource.name}' ({resource.get_size_x()} x {resource.get_size_y()} mm) is smaller "
+          f"than '{self.name}' ({self.get_size_x()} x {self.get_size_y()} mm) and cannot cover it."
+        )
+      self._lid = resource
+      location = location or self.get_lid_location(resource)
+    else:
+      assert location is not None, "Location must be specified if resource is not a lid."
```
rickwierenga: use value error not assert

Issue #1159, opened 2026-07-08
Difference between Hamilton Method Editor initialization and PLR initialization


HaamsRee:
## `STARBackend.setup()` discards mounted tips through `C0DI`, unlike Hamilton Method Editor, causing crash at waste

Hello, it's me again with the weird half-sized 4-channel setup. Finally back to testing. This seems to be a separate problem.

### Summary

On my Hamilton STAR, calling `STARBackend.setup()` while tips are mounted can crash during the pipetting-channel initialization/discard step. The robot moves to the firmware waste X position, then appears to shift further in X during the discard motion. The tips hit the slope of the waste/trash block instead of the flat part, Z touch does not trigger normally, and the channels continue down, bending the tips.

This seems to happen because PyLabRobot uses `C0DI` to discard mounted tips during setup, while Hamilton Method Editor uses `C0DI` only for initialization and then uses `C0TR` for the actual tip discard.

### Hardware/configuration

- Robot: Hamilton STAR
- Pipetting channels: 4
- Channel minimum Y spacing: 18 mm, not the usual 9 mm
- Firmware extended configuration reports:
  - `xw13400` → tip waste X = `1340.0 mm`
  - `ys180` → minimum raster pitch = `18.0 mm`
- The crash happens during initialization when tips are already mounted.

### Observed behavior in PyLabRobot

During `STARBackend.setup()`, PyLabRobot does:

```python
tip_presences = await self.request_tip_presence()

if not initialized or any(tip_presences):
  await self.initialize_pip()
```

`initialize_pip()` then sends `C0DI` with active channels and a deep discard range:

```python
    await self.initialize_pipetting_channels(
      x_positions=[
        int(self.extended_conf.tip_waste_x_position * 10)
      ],  # Tip eject waste X position.
      y_positions=y_positions,
      begin_of_tip_deposit_process=int(self._channel_traversal_height * 10),
      end_of_tip_deposit_process=1220,
      z_position_at_end_of_a_command=3600,
      tip_pattern=[True] * self.num_channels,
      tip_type=4,  # TODO: get from tip types
      discarding_method=0,
    )
```

This means setup uses DI as the mounted-tip discard path.

### Comparison with Hamilton Method Editor

I ran this sequence in Hamilton Method Editor:

1. Initialize instrument
2. Pick up four tips
3. Initialize instrument again

That worked correctly and discarded the tips to waste without crashing.

From the Hamilton USB trace, before the second initialization:

```
[...]

PR #1161, opened 2026-07-08
Fix and test STAR setup discard of mounted tips

HaamsRee:
## Summary

Fixes STAR backend setup behavior when tips are already mounted on single channels. #1159 

Previously, `STARBackend.setup()` called `initialize_pip()` when tip presence was detected. That made the `C0DI` initialization command run with active channels (`tm1...`), so firmware could try to discard mounted tips as part of `DI`.

This PR changes the mounted-tip setup path to:

1. Run pip channel initialization with inactive channels (`tm0...`) so `C0DI` does not discard tips.
2. Discard the mounted tips separately using the existing `C0TR` tip-discard path.
3. Re-query tip presence and raise if tips remain.

The normal `initialize_pip()` behavior is unchanged.

## Motivation

On a STAR with 4 channels spaced at 18 mm, setup-time discard through `C0DI` caused the tips to hit the waste slope and continue downward, bending tips. A Hamilton Method Editor trace showed that Hamilton initializes with inactive channels and then performs the actual discard with `C0TR`.

## Tests

Added STAR backend tests for:

- existing `initialize_pip()` active `DI` command behavior,
- setup-only inactive `DI` command behavior,
- setup-time `TR` discard command using the detected tip-presence pattern.

Ran:

```bash
uv run pytest pylabrobot/liquid_handling/backends/hamilton/STAR_tests.py -q --tb=short --disable-warnings --color=no
```

Result:

```bash
110 passed, 13 warnings, 81 subtests passed
```


pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
```diff
@@ -2095,8 +2095,19 @@ async def set_up_pip():
         # Skip pip-channel I/O; the __init__ defaults stand in.
         # TODO: does not yet gate request_tip_presence or instrument-init moves.
         return
-      if not initialized or any(tip_presences):
+      has_tips = any(tip_presences)
+      if has_tips:
+        await self._initialize_pip_without_tip_discard()
+      elif not initialized:
```
rickwierenga, line 2101: should this initialize call not be gated by `if initialized`?

HaamsRee: The intention was to keep somewhat consistent with the old behavior.

The old behavior, if I'm understanding correctly, is:

If not initialized or there's any tip left on the channels, do `initialize_pip()`. `initialize_pip()` then does the C0DI with tm1&. That works fine when there are no tips, but crashes, at least in my case, with tips.

So I split the two conditions. If there are tips present, call the new `_initialize_pip_without_tip_discard()` which sends C0DI but with tm0& instead. This does the initialization but the tips stay afterward. If there are no tips, just do the old `initialize_pip()`.

But since `_initialize_pip_without_tip_discard()` doesn't discard tips, we need to actually discard the tips to complete the initialization as Method Editor does. This is checked and done below. And the check won't run if it went the `initialize_pip()` route at the start since there were no tips to begin with.

PR #1162, opened 2026-07-10
Update arm resource state after pickup

j0nch:
## Summary

This updates arm resource moves so the resource model reflects pickup/drop boundaries in real time instead of waiting for the full move to complete.

For both the new arm capability stack and the legacy `LiquidHandler` resource-move path:

- after a successful physical pickup, the resource is unassigned from its source location
- after a successful physical drop, the resource is assigned to the destination
- drop placement math uses resource rotation captured at pickup time, so it does not depend on the source parent still being present

This allows downstream orchestration to safely react as soon as a source location is actually free. For example, a workflow can begin staging another resource to a handoff location once an arm has picked up the current resource, rather than waiting for the full pickup + travel + drop sequence to complete.

## Legacy STAR/iSWAP release-width fix

While testing the legacy STAR/iSWAP path, we found one important detail: after early unassignment, legacy `STARBackend.drop_resource()` could no longer reliably recompute the release width from `resource.get_absolute_size_x/y()`, because the resource no longer had its source/carrier rotation context.

This branch now captures `resource_width_at_pickup` in the legacy `ResourcePickup` / `ResourceDrop` flow and uses that captured width for the iSWAP drop release width. This matches the new capability path, which already carries held-resource width state.

Without this, some rotated/source-context-dependent moves could release at a width derived from the wrong resource axis instead of the actual gripped width.

## Tests

- Added capability arm tests covering source unassignment after pickup and destination assignment after drop.
- Added legacy `LiquidHandler` tests covering source unassignment after pickup, destination assignment after drop, and preservation of pickup-time release width after source unassignment.

Locally run:

```bash
python3 -m py_compile \
  pylabrobot/capabilities/arms/arm.py \
  pylabrobot/capabilities/arms/articulated_arm.py \
  pylabrobot/capabilities/arms/orientable_arm.py \
  pylabrobot/legacy/liquid_handling/standard.py \
  pylabrobot/legacy/liquid_handling/liquid_handler.py \
  pylabrobot/legacy/liquid_handling/backends/hamilton/STAR_backend.py \
  pylabrobot/legacy/liquid_handling/liquid_handler_tests.py

python3 -m unittest pylabrobot.capabilities.arms.arm_tests
python3 -m unittest pylabrobot.legacy.liquid_handling.backends.tecan.EVO_tests
```

Also ran the targeted legacy `TestLiquidHandlerSerializeState` suite locally; all targeted tests passed.

## Hardware validation

Validated on a STAR/iSWAP workcell using the legacy `LiquidHandler` path:

- source locations became available at pickup as expected
- destination assignment/orientation remained correct after drop
- iSWAP release width remained correct after the `resource_width_at_pickup` fix
- an orchestration test was able to safely overlap downstream work using the earlier source-free signal while preserving single-arm-operation and single-handoff-location occupancy invariants

### DIFF
--- a/pylabrobot/capabilities/arms/arm.py
+++ b/pylabrobot/capabilities/arms/arm.py
@@ -30,6 +30,10 @@ class _PickedUpState:
   offset: Coordinate
   pickup_distance_from_bottom: float
   resource_width: float
+  # Preserve the resource pose at pickup time. Once pickup succeeds, the
+  # resource is unassigned from its source so the source is immediately free in
+  # the model; later drop math must not depend on the old parent tree.
+  resource_absolute_rotation_at_pickup: Rotation
   rotation: Rotation = Rotation()
 
 
@@ -152,11 +156,9 @@ class _BaseArm(Capability):
     to_location: Coordinate,
     offset: Coordinate,
     pickup_distance_from_bottom: float,
-    rotation_applied_by_move: float,
+    resource_absolute_rotation_after_move: Rotation,
   ) -> Coordinate:
-    center = resource.center().rotated(
-      Rotation(z=resource.get_absolute_rotation().z + rotation_applied_by_move)
-    )
+    center = resource.center().rotated(resource_absolute_rotation_after_move)
     loc = to_location + center + offset
     return Coordinate(
       loc.x,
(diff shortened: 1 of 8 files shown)


j0nch:
Superseded by #1194, which ports the active legacy resource-movement behavior onto current main. This PR targets the now-integrated v1b1 branch and includes the capabilities-arm package that current main no longer contains.

PR #1168, opened 2026-07-20
feat(sartorius): add Sartorius Entris II balance driver

rickwierenga:
Adds a driver for the Sartorius Entris II balance (pylabrobot/sartorius/entris.py).

- SartoriusEntris2: plain class over pylabrobot.io.serial.Serial (9600 8-O-1).
  API: setup, stop, read_weight, tare, zero, get_model, get_serial_number,
  send_command.
- SBI print protocol: <ESC>P weigh, <ESC>T tare, <ESC>x1_ model, <ESC>x2_
  serial. ESC byte 0x1B by default; esc_literal=True sends literal "ESC ".
- Device error codes parsed from Stat… responses (DEVICE_ERRORS).
- Docs API entry added.

Not tested against hardware; setup() logs a warning asking for a PR to remove
it once verified.

### DIFF
--- a/pylabrobot/sartorius/entris.py
+++ b/pylabrobot/sartorius/entris.py
@@ -51,7 +51,8 @@ class SartoriusEntris2:
 
   Commands (SBI control commands, format "<Esc> <char> CR LF"):
     <Esc>P     print current weight value
-    <Esc>T     zero/tare
+    <Esc>T     tare (Zero/Tara command)
+    <Esc>V     zero (Key ZERO)
     <Esc>x1_   query balance model
     <Esc>x2_   query serial number
     The trailing "_" in x1_/x2_ is a literal underscore. <Esc> is the escape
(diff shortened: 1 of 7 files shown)


pylabrobot/sartorius/entris.py
```diff
@@ -0,0 +1,197 @@
...
+    if filtered.count(".") > 1 or filtered in ("", "+", "-"):
+      raise SartoriusError(title="Invalid weight value", message=repr(response))
+    try:
+      return float(filtered)
+    except ValueError as exc:
+      raise SartoriusError(title="Invalid weight value", message=repr(response)) from exc
+
+  # === Public API ===
+
+  async def read_weight(self) -> float:
```
BioCam, line 173: I propose naming it `measure_weight` for consistency with measurement name classification - "reading" comes from scales with a display but many scales don't have a display, devoiding the meaning of the word.
"Measure" is accurate in all instances

PR #1176, opened 2026-07-21
feat(kbiosystems): add Ultraseal PRO (formerly WASP) heat sealer driver

rickwierenga:
## What

Adds `KBiosystemsUltrasealPRO`, a PyLabRobot driver for the KBiosystems **Ultraseal PRO** inline heat sealer, alongside the existing Ultraseal ePRO / XT Pro and KUBE sealers.

**Naming:** the Ultraseal PRO was sold for years as the **WASP** ([product page](https://kbiosystems.com/product/ultraseal-pro), "formerly the Wasp"). Following the ePRO/eSeal precedent in this package, the class, module, and docs are named after the current product (`KBiosystemsUltrasealPRO`, `ultraseal_pro.py`, `ultraseal-pro/`), with "(formerly the WASP)" noted in the docstring and tutorial so people with older units can still find it.

The Ultraseal PRO speaks the same shared ASCII/RS-232 protocol as the other KBiosystems sealers (`?`/`E`/`S`/`A`/`B`/`C`/`D`/`F`, 9600 8-N-1, command-echo, CR-terminated), so `KBiosystemsUltrasealPRO` subclasses `KBiosystemsSealer` and reuses the shared transport and status/error state machine. It is closest to the Ultraseal XT Pro (inline, pneumatically driven, self-advancing/cutting film, compressed-air supply), with two deliberate differences taken from the protocol manual:

- **Status byte layout** — `ParkMode` is bit 7 (`0x80`) with bit 6 (`0x40`) a documented spare, whereas the XT Pro reports `ParkMode` at `0x40`. `UltrasealPROStatus` keeps `Spare = 0x40` as a member so a byte with that bit set decodes instead of raising.
- **No shuttle commands** — the Ultraseal PRO draws the shuttle in, seals, and returns it automatically during the seal cycle, so it exposes no `park`/`unpark`/`reset` (`P`/`U`/`R`) over serial. It also self-initializes on power up (no `I` command).

Its error table is the model-specific subset (1, 2, 3, 4, 7, 9, 10) and the sealing-temperature ceiling is 200 °C per the spec sheet.

## Hardware-verified

The Ultraseal PRO driver has been confirmed against a real device, so it ships without the not-tested warning. As part of this, the warning was moved out of the shared `_open()` and into each unverified model's `setup()` — so verifying a model is just deleting that one line. The ePRO and XT Pro (still unverified) keep theirs; the Ultraseal PRO has none.

## Included

- `pylabrobot/kbiosystems/ultraseal_pro.py` — `KBiosystemsUltrasealPRO`, `UltrasealPROStatus`
- `__init__.py` exports
- `sealer.py` — `_open()` no longer emits the maturity warning (relocated to the unverified subclasses' `setup()`)
- Tests in `sealer_tests.py` (setup, seal wire-bytes, bit-7 Park Mode vs bit-6 spare, low-air decode, overheating error; extended the shared-low-bits check to include the Ultraseal PRO)
- Docs: API autosummary entry + an `ultraseal-pro/hello-world.ipynb` tutorial

## Checks

- `pytest pylabrobot/kbiosystems/sealer_tests.py` — 19 passed
- `ruff check` clean, `ruff format` applied, `mypy` clean
- Verified the not-tested warning fires for the ePRO and XT Pro and is absent for the Ultraseal PRO

### DIFF
--- a/docs/api/pylabrobot.kbiosystems.rst
+++ b/docs/api/pylabrobot.kbiosystems.rst
@@ -23,6 +23,16 @@ pylabrobot.kbiosystems package
     KBiosystemsUltrasealEPRO
     UltrasealEPROStatus
 
+.. currentmodule:: pylabrobot.kbiosystems.ultraseal_pro
+
+.. autosummary::
+  :toctree: _autosummary
+  :nosignatures:
+  :recursive:
+
+    KBiosystemsUltrasealPRO
+    UltrasealPROStatus
+
 .. currentmodule:: pylabrobot.kbiosystems.ultraseal_xt_pro
 
 .. autosummary::
--- a/docs/user_guide/kbiosystems/index.md
+++ b/docs/user_guide/kbiosystems/index.md
@@ -4,5 +4,6 @@
 :maxdepth: 1
 
 ultraseal-epro/hello-world
+ultraseal-pro/hello-world
 ultraseal-xt-pro/hello-world
 ```
--- a/docs/user_guide/kbiosystems/ultraseal-pro/hello-world.ipynb
+++ b/docs/user_guide/kbiosystems/ultraseal-pro/hello-world.ipynb
@@ -1,0 +1,98 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "ultraseal-pro-intro",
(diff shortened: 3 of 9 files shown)


ben-ray:
I tested this on my kbiosystems wasp (also known as Thermo ALPS3000), and it works. Note that you need a straight rs232 cable to an rs232 port to control these instruments. I tried many different chipset rs232-usb cables, none worked besides an ol' rs232 port.

PR #1182, opened 2026-07-22
`STARBackend`: track the X-arm carriage via a deck-owned `XArm` resource

BioCam:
The STAR moves its X-arm carriage but PyLabRobot keeps no model of where it is, so the Visualizer cannot draw the arm and nothing downstream can reason about its position.

This is the second PR of the epic "Safety Systems: Tracking STAR Capabilities", building on #1181. It adds carriage-position tracking via a deck-owned `XArm` resource that owns an `XArmTracker` - the same ownership pattern as a `TipSpot` owning a `TipTracker`, so the backend drives the position from firmware and the Visualizer reads it through the standard state channel.

- Adds an `XArm` resource (a full-deck-height frame at the channel stop-disk safe Z) that owns an `XArmTracker` whose `x` is exposed as resource state. The deck creates one `XArm` per installed arm at setup via `get_or_create_x_arm`, sized and modelled from the resolved X-drive configuration (`extended_conf.left_x_drive` / `right_x_drive`).
- The lowest-level X moves and position queries keep the tracker current: `experimental_x_arm_move`, `position_left_x_arm_`, and `position_right_x_arm_` commit the commanded target (invalidating it if the move fails), and `request_left_x_arm_position` / `request_right_x_arm_position` record the read-back value. `get_x_arm_position` reports the tracked `(left, right)` reference points, `None` while unknown.
- Seats each arm so its reference point sits at the arm's current x, read from the machine at setup (the left parks to a known home; the right reads its own), rather than at a shared nominal origin. Because the two arms then occupy distinct real positions, no collision override is needed to place them.
- Makes `HamiltonDeck`'s placement-overlap check 3D: two resources collide only when their bounding boxes intersect on all three axes. The arm sits above the deck plane (at the stop-disk safe Z), so it no longer counts as occupying the footprint of the carriers beneath it. This is a strictly-looser change (it only adds a z condition), so no layout that placed cleanly before can newly fail; it also applies to `VantageDeck` / `NimbusDeck`, which inherit the base deck.
- The Visualizer draws each arm as a translucent full-deck-height frame with a reference line at the tracked `x`, gliding on update, and floats it above the deck it spans (re-applied after resources are added, since a resource assigned after the arm would otherwise paint over it).

Behaviour: additive. A new deck-owned resource plus tracker, and the existing X moves and queries now record state - no wire behaviour changes. The arm is state and view only: it does not gate motion (reachability is the #1181 bound; collision reasoning comes later in the epic). The deck overlap check is only loosened (z added), so existing layouts are unaffected. `STARChatterboxBackend.request_right_x_arm_position` now reports "not installed" based on the configured right drive rather than on whether the tracker exists, so setup can read the home before the tracker is created.

Caveat: only the left arm's resting home is validated against hardware (the machine parks it to a known x at every init). The right arm's home is deck-derived in simulation (`_simulated_right_x_arm_home`) and read from the machine on hardware, but has not been checked against a physical dual-rail STAR (#822); dual-rail is uncommon and the code path is identical to the left, so it self-corrects on real hardware.

Tests: `TestXArmPositionTracking` (each low-level move and query updates the tracker, and a failed move invalidates it), `TestXArmPresence` (setup creates one deck `XArm` per installed arm, and re-setup reuses rather than duplicates it), `TestXArmVisualizerXArms` (the arm renders with the expected geometry at the tracked position), and new `HamiltonDeck` cases for the 3D overlap (a resource above the deck plane no longer blocks placement beneath it, same-footprint same-z still collides, and flush faces do not). `ruff format`, `ruff check --select I`, `ruff check`, and `mypy pylabrobot --check-untyped-defs` are clean; the STAR, resources, and visualizer suites pass. The Visualizer paint-order (arm drawn above the deck) is verified in a headless browser rather than the unit suite.

### DIFF
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -2359,8 +2359,11 @@ class STARBackend(HamiltonLiquidHandler, HamiltonHeaterShakerInterface):
         continue
       if arm.width is None:
         raise RuntimeError(f"{position} X-arm geometry not resolved")
-      deck.get_or_create_x_arm(f"{position}_x_arm", arm.width, arm.model, arm.reference_point)
-      await queries[position]()
+      x = await queries[position]()  # read the arm's current x before seating the resource
+      x_arm = deck.get_or_create_x_arm(
+        f"{position}_x_arm", x, arm.width, arm.model, arm.reference_point
+      )
+      x_arm.tracker.set_x(x)  # seed the tracker now that the arm (and its tracker) exists
 
   def _check_x_arm_reachable(self, x: float, position: Literal["left", "right"] = "left") -> None:
     """Raise if x (mm) is outside the X-drive's travel range (``x_range``).
--- a/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
+++ b/pylabrobot/liquid_handling/backends/hamilton/STAR_backend.py
@@ -99,12 +99,15 @@ from pylabrobot.resources.hamilton import (
 )
 from pylabrobot.resources.hamilton.hamilton_decks import (
   HamiltonCoreGrippers,
+  HamiltonSTARDeck,
   rails_for_x_coordinate,
 )
 from pylabrobot.resources.liquid import Liquid
 from pylabrobot.resources.rotation import Rotation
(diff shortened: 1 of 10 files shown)


BioCam:
superseded by upcoming v1 STAR integration

PR #1184, opened 2026-07-23
feat(opentrons): Opentrons Flex liquid handler (plain-class, mount-addressed heads)

vcjdeboer:
## Opentrons Flex liquid handler (plain-class, mount-addressed heads)

Adds the Opentrons Flex as a plain-class device, following the post-capability architecture. It drives the on-robot **robot-server HTTP API** (Protocol Engine: `/runs`, `/commands`, `/instruments`) directly.

### Architecture

- **`OpentronsRobot(abc.ABC)`** — shared base owning the transport, the run/command lifecycle, and instrument discovery. The transport sits behind a small **`OpentronsTransport` Protocol** with two implementations: an `httpx` transport for real hardware and an offline recording transport for dry runs (so the whole device is testable without a robot).
- **`OpentronsFlex(OpentronsRobot)`** — the device. `setup()` discovers the mounted pipette(s) and composes a head sub-object per gantry mount (`flex.left` / `flex.right`), or `flex.head96` for the 96-channel head. `stop()` drops any mounted tips to the trash, homes the gantry, then cancels the run and disconnects.
- **`FlexHead1` / `FlexHead8` / `FlexHead96`** — mount-addressed fixed heads. Each operation sends **one** robot-server command anchored at the reference well; the fixed head fans it out to its N nozzles (a column for Head8, the whole plate for Head96, a single well for Head1). Head8 also supports single/partial pickup via `configureNozzleLayout`.

### Behaviour

- **Resource-tree state.** Tip and volume state commit to `TipSpot.tracker` / `Well.tracker`, only for the channels actually actuated (None-skip), and only through a transactional **stage → wire → verify → commit/rollback** flow — so an infeasible operation raises before any hardware motion, and a failed wire command leaves no partial state. Gated by the global `does_tip_tracking()` / `does_volume_tracking()` switches.
- **Hardware tip-presence authority.** The Flex's per-pipette `tipDetected` sensor is the ground truth for tip presence: pickups are verified against it and rolled back on a missed pickup, and `get_mounted_tips()` (public per-channel telemetry) is reconciled against it.
- **Positioning.** Aspirate/dispense default to 1 mm above the well bottom (the Opentrons default) and auto-issue `prepareToAspirate` before the first aspirate after a pickup.
- **Name-based labware.** The robot owns the authoritative labware geometry, resolved from the Opentrons load name (`ot_load_name`); the PLR resources carry a nominal SBS grid for tracking/addressing only. `FlexDeck` models slots as `ResourceHolder`s; name-based tip-rack and plate factories are included.

### Hardware status

`FlexHead8` is **verified on real Opentrons Flex hardware** (robot-server API 8.8: setup, homing, and column tip pickup confirmed against the tip-presence sensor). `FlexHead1` and `FlexHead96` are implemented but **not yet hardware-verified** (they need 1-channel / 96-channel pipettes) and emit a one-time warning on first use.

### Docs & tests

- A Head8 hello-world notebook (`docs/user_guide/opentrons/flex/hello-world.ipynb`) and an API reference page, wired into the docs toctrees.
- Offline unit tests for the transport, device composition, and all three heads (via the recording transport).

### Follow-ups (not in this PR)

- `FlexRobotGeometry` native reach checks (per-mount / nozzle-envelope), folding a grounded operating-envelope model.
- `FlexDeck` deserialization guard (reject labware assigned directly to the deck).
- OT-2 on the shared `OpentronsRobot` base (pending verification that the two share enough at the wire level).


Silousr:
**Downstream report: bridged this driver into an agent-facing protocol (simulation only)**

I maintain the labwire bridge that wraps PyLabRobot for AI-agent control (discussed in the earlier thread about cancellation on the Flex). Since the plain-class redesign will eventually reach us, I ran an experiment against this PR to see how a protocol adapter fares in the new architecture. Everything below is observation from that exercise, in case any of it is useful mid-redesign. Nothing here needs action on my account.

Setup: pinned this PR's head commit (6ee378e), drove OpentronsFlex through our protocol layer, and exercised it against a simulation of the robot-server command layer (mock HTTP for /health, /runs, /runs/{id}/commands, /instruments), which I understood to be how the PR itself was validated. No hardware involved on my side either.

What ported cleanly:

- FlexDeck behaves as a first-class Deck throughout. Resource traversal, name lookup, per-well tip and volume trackers, and our reference validation all worked over it with zero changes to our resource code. The slot-as-child-resource design is what made that true.
- The list-shaped method signatures (pick_up_tips, drop_tips, aspirate, dispense) are close enough to the LiquidHandler shapes that our existing handlers called the driver without modification.
- One command per HTTP POST maps well to honest cancellation semantics. After the earlier Flex discussion we ended up declaring every atomic operation non-cancellable and only bridge-sequenced compounds stoppable between steps; this driver's command granularity made that mapping exact.

Things I hit that may be worth knowing:

- Installing the PR head as a package (pip install from the commit) cannot import pylabrobot.liquid_handling: the chain fails at pylabrobot.molecular_devices.imageXpress, which is missing from the installed package at that commit. Looks like a packaging artifact of the older v1b1 base rather than anything in the driver files; current v1b1 does not have the problem, so a rebase presumably clears it. Only matters to someone consuming the PR as an install rather than a checkout.
- Separately, on current v1b1 (1ae9dc6), our shipped LiquidHandler-based test suite passes 110 of 113 through the legacy shims. The three failures, in case the data is useful: pylabrobot.liquid_handling.errors has no shim module (old-path import of ChannelizedError breaks), and the Cor_96_wellplate_360ul_Fb factory now produces model string cor_96_wellplate_360uL_Fb, which breaks anything keyed on the old model string. In our case that key was a safety annotation, so the mismatch failed silent rather than loud, which was a good lesson about string-keyed config on our side.
- Multi-element calls: the driver builds each robot command from element [0] of its list parameters but commits tracker state for every zipped element, so a two-well aspirate updates both wells' volume trackers while the robot aspirates only the first. The PR already describes itself as single-channel-first, so this is expected territory; the part that bit us as a consumer is that the tracker side is silent about it. Our adapter refuses multi-element calls until batching lands, which was an easy guard once we saw it.
- Channel state: which physical channel holds which tip lives in _channel_tips, which is private. LiquidHandler.head was public, and our telemetry read it. In the plain-class model there was nothing public for an adapter to read for that, so we read the private attribute and documented that we did. Just noting where the public surface ended for us; the resource-tree trackers covered everything else.

The driver held up well under a fairly adversarial consumer. Happy to share the simulation harness or the full compatibility notes if either is ever useful.

rickwierenga:
thank you @vcjdeboer !

could you please rebase this onto main?

vcjdeboer:
@Silousr thanks for the test. The PR has been reworked since the commit you pinned. It is now a mount-addressed head model (flex.left / flex.right / flex.head96, with FlexHead1 / FlexHead8 / FlexHead96), so two of your points are addressed:

- Multi-element tracker divergence: the fixed-head model has no arbitrary use_channels path. Each op sends one command anchored at the reference well and the hardware fans it to the head's nozzles; trackers
- commit only for actuated channels (None-skip), via stage → wire → verify → commit/rollback. Arbitrary multi-well isn't offered rather than silently mishandled.
- Channel→tip state: get_mounted_tips() is public now, and has_tip_on_hardware() reads the Flex tipDetected sensor, which is the authority. Pickups verify against it and roll back on a miss.

The import artifact was a v1b1-base issue and clears on the current main rebase. FlexHead8 is verified on a real Flex; Head1/Head96 are coded but not yet hardware-tested.

Also, the bigger reason was the abstraction: the Flex only has fixed heads (1/8/96-channel across two gantry mounts), so a flat single-channel class with a use_channels list was modelling independently-addressable channels the machine doesn't have (I think use_channels is a Hamilton STAR concept). The head model matches the real hardware, and it lets the robot stay the authority for the things it owns anyway (labware geometry via load names, tip presence via the tipDetected sensor).

So this build doesn't include a reach-geometry model like the OT2RobotGeometry that @BioCam added for the OT-2 (in resources/opentrons/, though nothing wires it into a device yet I think). Worth deciding whether the Flex should use the same approach.

miikee:
@vcjdeboer We built on this branch and opened a PR against it: vcjdeboer/pylabrobot#1 (raising it here since people watch this PR rather than incoming PRs on a fork).

Adds the gripper, containers/reservoirs, motion, fine pipetting, custom labware upload, and per-pipette flow rates, and runs it on real hardware for the first time: a Flex with a p50 single channel, 35 checks so far. Nothing removed, 25 of your public methods untouched.

Two things worth flagging even if you don't want the PR:

- The driver was sending its own `prepareToAspirate` before a well-addressed aspirate. The robot already primes those itself, at the well top in open air, then descends. Sending our own first set the ready flag and skipped that, so the prime happened wherever the tip already was: submerged, that draws roughly 3.9 uL on a p50 or 79.5 uL on a p1000 of the wrong well into the tip. It passed unit tests because the chatterbox accepted whatever the driver sent; it now models the engine's ready-to-aspirate rule.
- Dropping `opentrons-shared-data` matters more than it looks. It pins `numpy~=1.26.4`, which fights whatever numpy the rest of an environment wants. It was only read for two static tables, so those now live in the tree with drift tests against Opentrons' own definitions.

Happy to split any of it up or drop parts if you'd rather keep this PR lean for the rebase onto main.

vcjdeboer:
Merged, thanks @miikee. I went through it closely and ran the new surface on the robot first. A few notes, plus a follow-up commit I layered on top.

Verified against the robot's own 8.8.1 engine and kept as-is:
- Priming fix is right. A well-addressed aspirate self-primes at the well top (commands/aspirate.py) and a pickup primes the plunger (execution/tip_handler.py:342), so my old client-side prepareToAspirate was priming submerged and drawing the well in. Good catch.
- io.HTTP was empty for the OT2, copied that pattern; filling it with capture/replay is the right call.
- Per-model flow rates confirmed (p1000 on a 50 µL tip really is 6 µL/s at v3.3 vs 478 at v3.4), and the drift test is nice.
- getTipPresence: agreed it's simulator-only (legacy_simulator/legacy_instrument_core.py:411), not a hardware bug.

Two things to reconcile:
1. Aspirate surface. My follow-up unifies (plate, column) + aspirate_single + aspirate_container into one aspirate(target, volume, *, use_channels) per head (same for pick_up_tips, which is where the layout is emitted since the engine won't reconfigure with tips on). Your container tracker-math and flow rates stay underneath, the old calls stay as bridges, and it scales to partial columns (QUADRANT). Happy to iterate.
2. numpy / shared-data. I think this got tangled between our parts. opentrons-shared-data doesn't pin numpy (8.3.0 and the robot's 8.8.1 both declare only jsonschema/pydantic/typing-extensions); the numpy<2 constraint is the opentrons package's, which shared-data doesn't require, and our extra is opentrons-http-api-client==0.2.1, which has no deps at all. My base never imported shared-data at runtime either (only mypy.ini). The vendoring plus drift test is still good; I just want the rationale to match.

What I added on top (hardware-tested on the 8-ch head, full column plus single H1 plus a front-4 partial): the unified surface, partial columns, and a reach/travel layer. Reach caps are grounded to 8.8.1, and travel always arcs above a tip rack (cross-slot moves and the trash-drop lift to a computed floor, so an idle rack the robot wasn't told about is still cleared). That trash-drop was travelling low.

vcjdeboer:
**Single-tip pickup.** The 8 nozzles move together, so a pickup is only "single" when the other 7 sit over empty spots. The clearance check for that stays.

I changed two things in the interface:

- **Default `primary_nozzle` is now `H1`, not `A1`.** With `H1` (front nozzle) the 7 idle nozzles hang over the back edge. So picking the back tip (well `A1`) always works, also on a full rack. `H1` also reaches front deck slots the back nozzle cannot.
- **`well` is now optional.** Leave it out and the next tip is taken from tip tracking, in the order the pickup goes through the rack (H1 back-to-front, A1 front-to-back) — the way the Opentrons engine does it. So "give me one tip" can never pick a well and anchor that collide. Give a well and it takes that exact tip. Tip tracking off and no well: it raises, it does not guess.

vcjdeboer:
@rickwierenga this is rebased onto main and green again. Could you do a review pass when you have time?

vcjdeboer:
Rebased again, now onto your new OT-2 code (#1193) @rickwierenga — the Flex driver adopts the same architecture (shared `OpentronsAPI`/run, driver under `opentrons/flex/` like `ot2/`). Tests green, ready for review. @miikee — heads up, we're adopting the OT-2 architecture here.

rickwierenga:
thank @vcjdeboer and @miikee !

PR #1195, opened 2026-08-09
feat(hamilton): complete TCP transport client on transport/tcp

cmoscy:
## Summary
Completes the Hamilton TCP/HOI client under `pylabrobot/hamilton/transport/tcp`: protocol, wire types, introspection, error tables, and unit tests.

Instrument-agnostic shared layer only — no Prep/Nimbus device peers. Prep (follow-up PR) builds on this.


rickwierenga:
`_send_raw` will write again on a second attempt which is unsafe if the first write succeeded. is it necessary to retry or is that just defensive? I would rather defer this stuff to the user/caller. also I think many errors are actual protocol errors, not just connections errors so in those cases retrying will not help and might even be dangerous

also it seems like we should have a background reading thread like the star has that reads all incoming commands and then matches them to the specific command that was sent

rickwierenga:
until we figure out concurrency on the tcp protocol (see the locking mechanism for the star), I think we should have a lock to force one command at a time

rickwierenga:
also is it possible to get rid of hasattr etc.?

PR #1197, opened 2026-08-10
fix(hamilton): resolve trash_core96 from the deck instead of a stale pointer

WillVT84:
## Problem

`HamiltonSTARDeck._trash96` is assigned only in `__init__`, and `get_trash_area96()` returns it directly. That cached reference does not survive the deck being rebuilt or the trash being unassigned, which produces two distinct failures.

**1. Deserialized decks raise even though the trash is present.** `serialize()` encodes the 96 trash as a child resource and emits `with_trash96=False`. A deck restored through `Deck.deserialize()` therefore runs `__init__` with the flag off and leaves `_trash96` as `None`, while `trash_core96` is re-attached from the serialized children. The resource is on the deck and correctly positioned, but the getter raises:

```python
deck2 = Deck.deserialize(STARLetDeck().serialize())
"trash_core96" in [c.name for c in deck2.children]  # True
deck2.get_trash_area96()
# RuntimeError: Trash area for 96-well plates was not created. Initialize with `with_trash96=True`.
```

Anyone loading a saved layout has to re-establish the pointer by hand.

**2. `clear(include_trash=True)` leaves a stale pointer.** `clear()` unassigns the trash but never clears the attribute, so the getter hands back an orphaned `Trash` with `parent = None`. That fails later and further from the cause than an outright error would.

## Fix

Resolve the resource by name against the child tree, mirroring how the base class's `Deck.get_trash_area()` already works, and only trust the cached reference while it is still assigned to this deck. The `parent is self` guard is what fixes the second case: an unassigned trash falls through to the lookup and then raises honestly.

Behaviour when the trash genuinely was not created (`with_trash96=False`) is unchanged.

## Tests

Two regression tests in `hamilton_deck_tests.py`, one per failure mode. Both were confirmed to fail against the unfixed source — the round-trip test errors with the original `RuntimeError`, and the `include_trash` test fails with "RuntimeError not raised" — and to pass with the fix.

## Verification

- `pytest pylabrobot/resources` — 225 passed. The 5 failures are pre-existing Opentrons deck tests that fetch labware over the network and hit `SSL: CERTIFICATE_VERIFY_FAILED` locally; identical on baseline.
- `ruff check` / `ruff format --check` (0.15.4, per the pin) — clean on both changed files.
- `mypy` (1.18.2, per the pin) — `Success: no issues found in 2 source files`.

## Note

Not addressed here, but the underlying design smell is that `serialize()` writes `with_trash`/`with_trash96`/`core_grippers` as `False`/`None` because the data lives in the children — the "not very pretty to have this key though" comments. That constructor-flags-vs-children mismatch is what produced this bug, and may affect other cached references the same way. Happy to follow up separately if that's of interest.

### DIFF
--- a/pylabrobot/resources/hamilton/hamilton_decks.py
+++ b/pylabrobot/resources/hamilton/hamilton_decks.py
@@ -458,12 +458,11 @@ class HamiltonSTARDeck(HamiltonDeck):
       origin=origin,
     )
 
-    self._trash96: Optional[Trash] = None
     if with_trash96:
       # got this location from a .lay file, but will probably need to be adjusted by the user.
-      self._trash96 = Trash("trash_core96", size_x=122.4, size_y=82.6, size_z=0)  # size of tiprack
+      trash96 = Trash("trash_core96", size_x=122.4, size_y=82.6, size_z=0)  # size of tiprack
       self.assign_child_resource(
-        resource=self._trash96,
+        resource=trash96,
         location=Coordinate(x=-42.0 - 16.2, y=120.3 - 14.3, z=216.4),
       )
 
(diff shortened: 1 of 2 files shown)


WillVT84:
@rickwierenga This may be an artifact of how we manage decks in our implementation, but I've been having to evoke a helper function to re-establish this pointer whenever we load our deck in a protocol. Figured I'd submit a bug fix here.

pylabrobot/resources/hamilton/hamilton_decks.py
```diff
@@ -548,11 +548,20 @@ def rails_to_location(self, rails: int) -> Coordinate:
...
+    # `_trash96` is only set in `__init__`, so it is stale whenever the deck was rebuilt from a
+    # serialized layout (which encodes the trash as a child and passes `with_trash96=False`) or the
+    # trash was unassigned. Resolve against the child tree instead, and only trust the cached
+    # reference while it is still assigned to this deck.
+    if self._trash96 is not None and self._trash96.parent is self:
+      return self._trash96
+
+    if self.has_resource("trash_core96"):
+      self._trash96 = cast(Trash, self.get_resource("trash_core96"))
+      return self._trash96
```
rickwierenga: why not always use `if self.has_resource("trash_core96"):`, that would be more robust. we can also delete the `_trash96` attribute in that case and simplify it.

rickwierenga:
thanks for the PR! I think we can simplify further per comment above

pylabrobot/resources/hamilton/hamilton_decks.py
WillVT84: Yes, this is cleaner. Changes have been made.

rickwierenga:
thanks!

PR #1198, opened 2026-08-10
feat: add opt-in contextual execution event bus

j0nch:
## Summary

Introduces an opt-in, in-process structured EventBus for contextual PLR execution events.

- Adds `pylabrobot.events` with a synchronous fan-out `EventBus`, scoped/default bus helpers, inherited operation context, correlated `started` / `completed` / `failed` operation events, and typed JSON-ready resource references.
- Uses PLR's canonical serialized `Coordinate` representation for geometric targets, including its `type: "Coordinate"` discriminator.
- Adds optional diagnostic transport events for serial, USB, and FTDI I/O, plus Hamilton USB and PreciseFlex firmware commands. These are distinct from semantic frontend operations.
- Preserves direct operated-resource identity while including structural ancestry for consumers needing a higher-level view.
- Keeps normal execution unchanged when no EventBus subscriber is registered; observer exceptions are isolated from hardware control flow.
- Adds focused tests for the EventBus and instrumented frontends.

This branch is based directly on current `main` and no longer includes or depends on the real-time liquid-handler resource-state changes from #1194.

## Current implementation coverage

This is deliberately an initial, opt-in implementation. It does **not** automatically instrument every PLR device or public method.

Currently instrumented frontends:

- `legacy.machines.Machine`: `machine.setup`, `machine.stop`
- `legacy.storage.Incubator`: `incubator.fetch_plate`, `incubator.take_in_plate`
- `legacy.liquid_handling.LiquidHandler`: resource pickup/move/drop; tip pickup/drop; 96-head tip pickup/drop; aspirate; dispense
  - Resource pickup records the direct moved resource and its direct source holder when assigned at invocation.
  - Resource drop records the destination holder.
  - Liquid and tip operations retain direct well/tip identities and channel-level operation details.
- `legacy.shaking.Shaker`: `shaker.shake`, `shaker.stop_shaking`
  - When a direct resource is loaded in the `ResourceHolder`, it is included in `resources`.
- `legacy.temperature_controlling.TemperatureController`: `set_temperature`, `wait_for_temperature`, `hold_temperature`, `deactivate`
  - When a direct resource is loaded in the `ResourceHolder`, it is included in `resources`.
  - `temperature_controller.hold_temperature` records a protocol-requested `duration` dwell and the configured target temperature when known. It does not reissue `set_temperature` or claim that the loaded resource reached the target.
- `brooks.precise_flex.PreciseFlex`: lifecycle, fault/home/freedrive, joint/cartesian/rail/gripper motion, pick/drop, and park
- `agilent.vspin.VSpin`: `centrifuge.spin`, including loaded bucket resources and requested cycle parameters
- `agilent.vspin.Access2`: `centrifuge_loader.load`, `centrifuge_loader.unload`

Diagnostic transport coverage, not semantic frontend-operation coverage:

- `io.Serial`, `io.USB`, and `io.FTDI`: `io.read` and `io.write`
- `hamilton.transport.usb.HamiltonUSBDriver`: `firmware.command.started`, `.completed`, and `.failed`
- `brooks.precise_flex.PreciseFlex`: `precise_flex.firmware_command.*`

Both semantic and diagnostic events are supported: semantic records describe public PLR operations, while diagnostic records preserve the transport and controller activity used to execute them.

## Documentation and driver-adoption contract

The PR adds two levels of documentation:

- A [user EventBus guide](docs/user_guide/machine-agnostic-features/event-bus.md) covering subscription, event shape, application context, semantic versus diagnostic records, and the current coverage matrix.
- Per-frontend user references for machine lifecycle, incubators, liquid handling, thermal/shaking, PreciseFlex, VSpin/Access2, and diagnostic transports.
- A [contributor guide](docs/contributor_guide/event-bus.md) defining the implementation contract for incremental driver adoption.

The contributor guide defines:

- lifecycle naming and operation-ID correlation;
- direct resource and physical source/destination semantics;
- resource-transfer, liquid-handling, tip-handling, thermal/shaking, centrifuge, loader, and controller-motion templates;
- failure-event behavior and use of PLR's canonical units without redundant field-name suffixes; and
- a contributor test checklist.

The contract deliberately does not require irrelevant fields for every device. Drivers should describe what their PLR frontend actually did, while downstream consumers derive their own display or protocol views from the structured references.

## Event model

Semantic operations emit correlated lifecycle records such as:

```text
liquid_handler.resource_pickup.started
liquid_handler.resource_pickup.completed
liquid_handler.resource_pickup.failed
```

The event context carries a stable operation ID and can be augmented by callers with execution context. Event data identifies the issuing device and direct PLR resources, without requiring consumers to infer resource identity from controller-level commands.


rickwierenga:
this is sick! very nice

- I dont think it should apply to legacy since we want to keep that code as similar as possible
- can you write a simple user guide / move some contributor things out of that tutorial and move it into one for users? basically most is user information right now.
- individual device pages in the user guide would benefit from a page on which events are emitted with which lifecycle options. ideally we include that for the devices that are also getting the life cycle events in this PR (many devices will be added later)

j0nch:
> this is sick! very nice
> 
> * I dont think it should apply to legacy since we want to keep that code as similar as possible
> * can you write a simple user guide / move some contributor things out of that tutorial and move it into one for users? basically most is user information right now.
> * individual device pages in the user guide would benefit from a page on which events are emitted with which lifecycle options. ideally we include that for the devices that are also getting the life cycle events in this PR (many devices will be added later)

Due to the lightweight, opt-in nature, I think it's okay to add events to legacy with very low risk of actually changing any legacy dependencies/behavior, though we could scrutinize it a bit more carefully if there's any concerns.

I've updated the contributor guide and split off additional user-facing documentation and guidance for how each device's events should look.

I've also updated the "coordinate" target implementation to use PLR's native serializable Coordinate, as you pointed out.

Issue #934, opened 2026-03-13
STARChatterboxBackend: configurable mock responses for query methods


BioCam:
## Context

`STARChatterboxBackend` overrides query methods (like `head96_request_tip_presence`, `core_check_resource_exists_at_location_center`, `request_presence_of_carriers_on_deck`, etc.) with hardcoded return values. This works for basic simulation but doesn't let users customize simulation behavior for different scenarios without subclassing.

## Proposal

Add a standardised pattern where query methods accept a `mock_response` parameter with a sensible default, allowing users to configure simulation parameters per-instance. For example:

```python
async def head96_request_tip_presence(self, mock_response: int = 0) -> int:
    return mock_response
```

### Design questions to resolve

1. **Per-call vs per-instance**: Should `mock_response` be a method parameter (configurable per call) or an instance attribute set at init/setup time (configure once, return consistently)?
2. **Signature compatibility**: The base class signatures don't have `mock_response`, so this parameter is only usable when calling directly on `STARChatterboxBackend`. Is that acceptable, or do we need a different approach?
3. **Scope**: Which methods should get this treatment? All query methods, or only the ones that return parsed data?
4. **Naming convention**: `mock_response`, `simulated_value` (already used by `channel_dispensing_drive_request_position`), or something else?

## Affected methods (current hardcoded returns)

- `head96_request_tip_presence()` → `0`
- `core_check_resource_exists_at_location_center()` → `True`
- `request_presence_of_carriers_on_loading_tray()` → `[]`
- `request_presence_of_carriers_on_deck()` → `[]`
- `request_tip_presence()` → tip tracker state
- `request_z_pos_channel_n()` → `285.0`
- `channel_dispensing_drive_request_position()` → `0.0` (already has `simulated_value` param)
- `request_iswap_initialization_status()` → `True`
- `request_pip_height_last_lld()` → `list(range(12))`

Note: `channel_dispensing_drive_request_position` already implements this pattern with `simulated_value` — that could serve as the starting point for a consistent convention.

VivienP:
Current `main` no longer matches the original method list.

Already owned elsewhere:
- tip presence, 96-head tip presence, and tip length: `TipTracker`
- last LLD heights: latched by simulated LLD
- parked flags and machine/factory configuration: existing backend/config objects
- carrier presence / CoRe resource-existence: firmware command path, not chatterbox hardcoded replies

[...]

PR #1200, opened 2026-08-13
Fix simulated STAR last LLD heights

VivienP:
## Summary

Update `STARChatterboxBackend.request_pip_height_last_lld()` to report the
heights produced by simulated liquid-level detection instead of returning the
hardcoded `list(range(12))`.

The chatterbox now keeps a per-channel latch of the latest simulated absolute
LLD height. `_run_lld_on_channel_batch()` updates that state using the
measurement it already computes, and `request_pip_height_last_lld()` returns it.

This keeps the simulated query consistent with the existing state-backed LLD
path without changing the public API.

Related to #934.

## Tests

Added regression coverage for:

- one returned value per configured channel before any LLD
- initial values of `0.0`
- updating the correct physical channel after `probe_liquid_heights()`

Validation:

- `STAR_tests.py`: 122 passed, 81 subtests
- `legacy/liquid_handling`: 479 passed, 1 skipped, 101 subtests
- Ruff: clean
- mypy: no new errors; 6 existing errors remain elsewhere in the repository

### DIFF
--- a/pylabrobot/legacy/liquid_handling/backends/hamilton/STAR_chatterbox.py
+++ b/pylabrobot/legacy/liquid_handling/backends/hamilton/STAR_chatterbox.py
@@ -105,6 +105,9 @@ class STARChatterboxBackend(STARBackend):
     self._num_channels = num_channels
     self._iswap_parked = True
     self._sim_iswap_information = iswap_information  # None means use default at setup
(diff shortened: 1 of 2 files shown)


rickwierenga:
thanks for the PR!

@BioCam can you review/merge this?

BioCam:
Thank you @VivienP !

Great addition to the simulation engine

PR #1201, opened 2026-08-13
`io.ftdi`: keep the bytes of a cancelled read, and record what a capture needs to replay

BioCam:
Follows #1199, now rebased onto `main`. The `io.usb` half of this PR is dropped: #1199 disposes the claimed device itself, on the worker thread rather than the loop thread, and covers the cancelled case as well.

Moving the blocking calls onto the executor makes them cancellable for the first time. `asyncio.wait_for` around a read on `io.ftdi` used to be a no-op, because the call never suspended and so always ran to completion; now the timeout fires. Cancelling the task does not stop the worker thread though: the transfer finishes, its bytes land in a future nobody is waiting on, and they are dropped. Those bytes are already out of the chip's buffer, so the next `read` picks up after them, part-way through the response the driver was assembling. Nothing raises where the loss happens, so it surfaces later as a checksum failure or a reply matched to the wrong command. `stop_shaking()` in `AgilentBioTekPlateReaderBase` already has this shape: it cancels `_shaking_task`, which may be inside `send_command`.

- `FTDI.read` shields the executor call and keeps what it produced in a small buffer that the next `read` is served from. A `read` issued while one is in flight waits for that one instead of starting a second.
- `FTDI.write` shields too, so a cancelled write reaches the device in full rather than leaving it holding a partial command, or none, with no way for the caller to tell which.
- `FTDI.usb_purge_rx_buffer`, `FTDI.usb_reset` and `FTDI.stop` drop the buffer and anything in flight, since all three mean the caller has declared that data stale.
- `FTDICommand` assigns `data` and takes `module`, like `SerialCommand` and `USBCommand` already do. It declared `data` without assigning it, so every `FTDICommand` was written out without its bytes, and rebuilding one from a capture file raised on `module`. `FTDIValidator.read` also compared the recorded hex string's length against a byte count, so no recorded read could match. `FTDIValidator` can now replay a recorded capture; wiring the `validate()` entry point back up, and widening `FTDIValidator.__init__` to accept what `FTDI.serialize()` returns, are separate and not touched here.
- `FTDI.read` and `FTDI.write` keep the `io.read` / `io.write` events added in #1209; both call sites moved, so `FTDIEventTests` locks them. `FTDI.readline` no longer emits an event of its own, because the single-byte reads it is now built from each emit one.
- `FTDI.readline` is assembled from single-byte reads, with a `terminator` and a `timeout`. `pylibftdi`'s own raises `TypeError` unless the device was opened in text mode and `FTDI.setup` opens it in byte mode, so the inherited one could only ever raise. No backend on `io.ftdi` calls it; the fifteen `readline` callers in the tree are all on `io.serial` or `io.socket`.

Behaviour: no change to a `read` or `write` that runs to completion, which is every caller in the tree today. A cancelled `read` still raises, but no longer costs the bytes already in flight; a cancelled `write` now always completes. `FTDI.readline` gains two optional arguments and returns an assembled line instead of raising. Captures from `io.ftdi` gain the `data` field they always declared; no existing `io.ftdi` capture file was replayable, so nothing regresses. `io.serial`, `io.hid` and `io.usb` reads have the same cancellation shape, pre-dating #1199 since they were already offloaded, and are left alone here.

Tests: `FTDICancellationTests` (a cancelled `read`'s bytes reach the next `read` and are captured once, a cancelled `write` still reaches the device, a purge discards held bytes, a text-mode device still reads, a second concurrent reader backs off), `FTDICaptureTests` (a capture replays through `FTDIValidator`), `FTDIEventTests` (a read and a write still emit their `io.*` events), and `FTDIReadlineTests` (`terminator`, `timeout`, empty `terminator`). All eleven of the pre-existing ones fail against `main` alone. `ruff format`, `ruff check --select I`, `ruff check` and `mypy pylabrobot/io --check-untyped-defs` are clean; the full suite passes (2259 passed, 2 skipped, 557 subtests).

### DIFF
--- a/pylabrobot/io/ftdi.py
+++ b/pylabrobot/io/ftdi.py
@@ -35,8 +35,9 @@ logger = logging.getLogger(__name__)
 class FTDICommand(Command):
   data: str
 
-  def __init__(self, device_id: str, action: str, data: str):
-    super().__init__(module="ftdi", device_id=device_id, action=action)
+  def __init__(self, device_id: str, action: str, data: str, module: str = "ftdi"):
+    super().__init__(module=module, device_id=device_id, action=action)
+    self.data = data
 
 
 class FTDI(IOBase):
@@ -85,6 +86,9 @@ class FTDI(IOBase):
     # Will be resolved in setup()
     self._dev: Optional[Device] = None
     self._executor: Optional[ThreadPoolExecutor] = None
+    # Bytes off the wire that no read has taken yet, and the read still in flight, if any.
+    self._unread = bytearray()
+    self._pending_read: Optional["asyncio.Future"] = None
 
     if get_capture_or_validation_active():
       raise RuntimeError(
@@ -272,6 +276,7 @@ class FTDI(IOBase):
   async def usb_reset(self):
     loop = asyncio.get_running_loop()
     await loop.run_in_executor(self._executor, lambda: self.dev.ftdi_fn.ftdi_usb_reset())
+    self._discard_reads()
     logger.log(LOG_LEVEL_IO, "[%s] usb_reset", self._device_id)
     capturer.record(FTDICommand(device_id=self.device_id, action="usb_reset", data=""))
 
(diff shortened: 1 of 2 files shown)


rickwierenga:
based

PR #885, opened 2026-02-10
RFC 7386 JSON Merge Patch serialization with compact format

rickwierenga:
## Summary
- Add `apply_merge_patch()`, `create_merge_patch()`, and `compact()` to `serializer.py` implementing RFC 7386 JSON Merge Patch
- `compact()` strips fields matching constructor defaults from serialized resource trees, producing smaller JSON
- `Resource.save()` now emits compact format automatically
- All `deserialize()` methods (12 files) updated to tolerate missing optional fields, so both old full-format and new compact-format JSON load correctly
- `deserialize()` in the core serializer now fills missing params from `__init__` defaults before calling constructors

## Test plan
- [x] 34 new/updated tests in `serializer_tests.py` covering `apply_merge_patch`, `create_merge_patch`, `compact`, and round-trip deserialization
- [x] All 563 existing tests pass unchanged
- [x] Verified old full-format JSON still deserializes correctly (backward compat)
- [x] Verified compact JSON with missing optional fields deserializes correctly
- [x] Verified `Resource.save()` → `Resource.load_from_json_file()` round-trip

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1205, opened 2026-08-16
Add resource metadata

burnpanck:
We found it useful to store arbitrary metadata with resources. For example, this enables us to label wells and containers with a metadata key `source_for`, indicating that these are where certain reagents can be found. Or it could be used to store barcode information with samples. Information regarding which direction a specific plate should be gripped by an iSWAP. Having that kind of information attached to the resources themselves is important because a) the metadata gets together with the deck, and b) it is cleaned up automatically when the corresponding resources are removed again from the deck.

### DIFF
--- a/pylabrobot/resources/resource.py
+++ b/pylabrobot/resources/resource.py
@@ -96,8 +96,8 @@ def _match_type(resource: Resource, matcher: TypeMatcher) -> bool:
   raise TypeError(f"Unexpected type matcher of type {type(matcher).__qualname__}")
 
 
-def _match_top_level_attribute(resource: Resource, attr_name: str, matcher: StrAttrMatcher) -> bool:
-  """Return whether ``resource``'s top-level attribute ``attr_name`` matches ``matcher``.
+def _match_attribute_value(attr_val: Any, matcher: StrAttrMatcher) -> bool:
+  """Return whether a top-level resource attribute value matches ``matcher``.
 
   The interpretation of ``matcher`` depends on its type:
 
@@ -107,7 +107,6 @@ def _match_top_level_attribute(resource: Resource, attr_name: str, matcher: StrA
 
   Note the callable receives different arguments than for `_match_type`.
   """
-  attr_val = getattr(resource, attr_name, None)
   if isinstance(matcher, re.Pattern):
     return attr_val is not None and bool(matcher.search(str(attr_val)))
   if callable(matcher):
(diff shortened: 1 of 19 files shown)


pylabrobot/resources/resource.py
```diff
@@ -126,30 +201,22 @@ def get_size_z(self) -> float:
...
       "size_x": self._size_x,
       "size_y": self._size_y,
       "size_z": self._size_z,
+      "location": serialize(self.location),
+      "rotation": serialize(self.rotation),
+      "category": self.category,
+      "model": self.model,
+      "barcode": self.barcode.serialize() if self.barcode is not None else None,
+      "preferred_pickup_location": serialize(self.preferred_pickup_location),
+      "metadata": self.metadata.copy(),
```
rickwierenga: maybe we can use `serialize` on the metadata so it recursively serializes data into the plr standard?

also since we are using JSON Merge Patch (RFC 7386) in v1 (see #885) , we should exclude None fields from metadata

rickwierenga:
thanks for the PR!

just one small comment above

could you also fix the merge conflicts?

after that this is good to merge

j0nch:
Fascinating feature, I'm already thinking about refactoring all of the protocol orchestration context dictionaries into Resource.metadata once this is merged!

pylabrobot/resources/resource.py
burnpanck: We made a conscious choice not to touch `metadata` at all, for the following reasons: In our case, we like to store non-JSON stuff (like `pint.Quantity` or some enums and dataclasses) in metadata that may not play well with `serialize`. More generally, `serialize` has a few quirky behaviours that can be perceived as footguns in unconstrained settings like `metadata`, such as silently converting strings like `"NaN"` into floats. In a similar way, omitting `None` may have surprising interactions with `has_metadata` queries. Technically, you could argue that there are no `None` default constructor arguments where RFC 7386 applies to. (also, if the goal is to make the files smaller, IMHO [removing structural duplication](https://discuss.pylabrobot.org/t/thoughts-on-having-a-compact-save-format-for-a-deck-a-resource-that-isnt-just-the-built-in-serialize/534/11) is much more valuable). I have no strong feelings towards `None` though.

rickwierenga: fair point

rickwierenga:
rebased it onto head

PR #1208, opened 2026-08-17
docs: add EventBus Slack notification example

j0nch:
## Summary

- extend the Slack notifications cookbook with a lightweight EventBus subscriber
- demonstrate forwarding completed and failed semantic operations without blocking instrument execution on webhook requests
- document the filtering policy and link the example from the EventBus user guide
- update the cookbook index to make the EventBus example discoverable

## Validation

- executed `docs/cookbook/slack_notifications.ipynb` with the chatterbox backend and no configured webhook
- `make docs-check`
- `git diff --check`


rickwierenga:
based

PR #1211, opened 2026-08-18
legacy(precise_flex): expose the lifecycle steps behind setup and stop

miikee:
`setup()` on the PreciseFlex does four things at once: opens the link, agrees the response protocol, raises high power, and sweeps the arm through its whole envelope to home. A caller who only wants to read a position has to accept the sweep, and one who wants the link without high power has no way to ask.

Each step is now its own public method, so the arm can be driven atomically:

- `connect` opens the link and agrees the response protocol. Powers nothing, moves nothing.
- `initialize` raises high power and attaches, so the arm accepts commands. Moves nothing.
- `home` is the only step that moves.
- `disconnect` detaches, drops high power, and releases the link. `initialize` is what raised the power, so this is what drops it.

`setup()` and `stop()` are unchanged in behavior and now just compose these, so existing code keeps working and the PyLabRobot convention still holds: `MachineBackend` declares both abstract, and the tree has 139 `setup` and 163 `stop` definitions. They stay the entry point; the split adds a finer grain underneath rather than replacing it.

Four tests added, covering that `connect` and `initialize` each send their own commands and neither homes, that `disconnect` drops power and releases the link, and that `setup(skip_home=True)` still does everything but home. Existing tests are untouched and pass.

**Tested on PreciseFlex PF400**

### DIFF
--- a/pylabrobot/legacy/arms/precise_flex/precise_flex_backend.py
+++ b/pylabrobot/legacy/arms/precise_flex/precise_flex_backend.py
@@ -91,16 +91,37 @@ class PreciseFlexBackend(SCARABackend, ABC):
     return arr
 
   async def setup(self, skip_home: bool = False):
-    """Initialize the PreciseFlex backend."""
-    await self.io.setup()
-    await self.set_response_mode("pc")
-    await self.power_on_robot()
-    await self.attach(1)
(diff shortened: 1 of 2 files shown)


rickwierenga:
thanks

rickwierenga:
over time we should have granular lifecycle control like this for all integrations

PR #1212, opened 2026-08-18
Fix legacy VSpin grip validation and add centrifuge events

j0nch:
## Summary

Fix the legacy Access2 loader's `grip_steps` validation and add canonical EventBus coverage to the resource-aware legacy centrifuge and loader frontends.

## Legacy Access2 bug fix

`Access2Backend.load()` currently uses:

```python
if not grip_steps not in (1, 2, 3, 4):
```

This rejects every valid value, including the default `grip_steps=1`, before the loader sends any hardware command. The condition was introduced with the configurable grip-step feature in #1140; `unload()` already uses the intended predicate.

This PR changes `load()` to the matching validation:

```python
if grip_steps not in (1, 2, 3, 4):
```

Regression coverage verifies that the default reaches the existing loader command sequence and that an invalid value fails before any command is sent.

## Legacy centrifuge EventBus coverage

The modern Agilent VSpin and Access2 frontends already emit:

- `centrifuge.spin`
- `centrifuge_loader.load`
- `centrifuge_loader.unload`

This PR adds the same canonical semantic operations to the resource-aware legacy `Centrifuge` and `Loader` frontends. Instrumenting the frontends rather than the FTDI backends allows events to report the actual PLR resources and physical transfer endpoints.

`centrifuge.spin` reports:

- the centrifuge resource in `device`;
- directly loaded resources in `resources`;
- holder/resource associations in `bucket_resources`;
- `relative_centrifugal_force` and `duration`;
- the requested or actual backend-default `acceleration_fraction` and `deceleration_fraction` when the backend exposes them.

Loader operations report the directly moved plate and actual holders:

- `centrifuge_loader.load`: loader to selected bucket;
- `centrifuge_loader.unload`: selected bucket to loader.

The context factories exactly match their decorated public method calling semantics, including positional `g` and `duration` arguments and forwarded backend keyword arguments. Their fields follow the EventBus contracts formalized in #1210.

## Compatibility

- No VSpin or Access2 command sequence was changed.
- No public frontend or backend signature was changed.
- With no active EventBus listener, the decorators remain no-ops.
- Existing resource assignment behavior is unchanged; assignment events are correlated beneath the semantic loader operation when a listener is active.
- The EventBus implementation is frontend-generic, so other legacy `Centrifuge` and `Loader` backends receive the same semantic operation family using their own resource model and backend defaults.

## Verification

- Modern VSpin/Access2 and legacy centrifuge suites: **23 passed**
- Tests cover successful spin and load lifecycles, unload failure lifecycle, resource/end-point semantics, positional spin arguments, VSpin backend defaults, valid default grip steps, and invalid grip-step rejection.
- `git diff --check`: clean

### DIFF
--- a/docs/user_guide/agilent/vspin/events.md
+++ b/docs/user_guide/agilent/vspin/events.md
@@ -1,6 +1,8 @@
 # VSpin and Access2 events
 
-Each listed semantic operation emits `started`, `completed`, or `failed` lifecycle records.
+Each listed semantic operation emits `started`, `completed`, or `failed` lifecycle records. The
+modern Agilent frontends and the resource-aware legacy `Centrifuge` and `Loader` frontends use the
+same canonical operation names and payload semantics.
 
 ## VSpin centrifuge
 
--- a/docs/user_guide/machine-agnostic-features/event-bus.md
+++ b/docs/user_guide/machine-agnostic-features/event-bus.md
@@ -121,6 +121,8 @@ events.
 | `legacy.liquid_handling.LiquidHandler` | resource pickup/move/drop; tip pickup/drop; 96-head tip pickup/drop; aspirate; dispense |
 | `legacy.shaking.Shaker` | `shaker.shake`, `shaker.stop_shaking` |
 | `legacy.temperature_controlling.TemperatureController` | set temperature, wait for temperature, deactivate |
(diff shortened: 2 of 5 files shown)


rickwierenga:
thanks good find

PR #1214, opened 2026-08-22
Single-nozzle pipetting height, and gripper targets that stay inside the axis

miikee:
Three bench findings from the on-site PF400/Flex session on 2026-08-21, all reproduced as unit tests.

**A single-nozzle aspirate or dispense can name its height.** `aspirate_single` / `dispense_single` hardcoded `None` for `liquid_height`, so a caller could describe where in the well to pipette on every other path but this one. It now passes through.

**No commanded gripper target reaches the end of the axis.** The servo overshoots: an open to the advertised maximum of 134.0 landed the axis at 134.062, and the controller then refused every commanded move, gripper or not, until the arm was homed. That happened twice in one session. Targets are now held half a unit inside whichever soft limits are known, on the millimetre path (`move_gripper`) and on the joint path (`move_gripper_joint_position`, which took no bounds at all). The default jaw opening moves 10 -> 14 to match the fitted gripper measured on the bench, and `default_open_position` names the target an "open the gripper" command should reach for instead of a stop.

**Two error messages pointed the reader the wrong way.** The out-of-range text said homing would not recover the axis. True of the rotary axes, which are absolute; false of the gripper, which has no brake, falls out of range whenever the motors lose power, and is re-seated by exactly the homing the message ruled out. It now answers per axis. The empty-jaws report gave only the threshold it used, so when a plate IS in the jaws nobody could tell whether to change the margin or the grip position; it names both.

Full suite: 2681 passed, 7 skipped.

This is the bottom of a four-repo wave. cheshire-drivers pins `pylabrobot@main`, so its PR is red until this one merges.


miikee:
Opened against the wrong repository by mistake; this belongs on the Cheshire-Labs fork. Closing.

PR #1216, opened 2026-08-24
v1: `hamilton`: complete the text protocol with command assembly and reply routing

BioCam:
Hamilton machines speak a text protocol: a command names a module, a command, an optional identifier and its parameters; the reply repeats them and carries an error field. `protocol/text/framing.py` currently only parses replies. This adds the other half, which the v1 Hamilton driver needs - building a command, reading the error fields back out, and matching a reply to the command that asked for it.

Most of that already exists, duplicated in `transport/usb/usb.py` and `legacy/liquid_handling/backends/hamilton/base.py`. The routing is private to those two classes, and one of them extends `LiquidHandlerBackend` - so a Hamilton heater shaker or tilt module, which speaks the same protocol but pipettes nothing, would have to inherit a liquid handler to reuse it.

- `assemble_command` builds a command string from a module, a command, an optional id and named parameters. It produces byte-identical output to `HamiltonLiquidHandler._assemble_command` for a bare command, mixed scalar types, a one-hot expanded list, and the reserved-word workaround.
- `to_list` is the one-hot channel encoding, where a value is given per involved channel and expanded to one per channel. Both `_to_list` methods now delegate to it and drop their copies.
- `find_error_fields` and `read_id` read a reply; `ReplyRouter` owns the reading thread and hands each reply to whichever command is waiting on it.

Behaviour: `to_list` rejects an empty list, or one longer than the channel pattern, with `ValueError` where one copy asserted. Both cases were already rejected further down the same function, so this only makes the rejection earlier and the message specific. Nothing sent to a machine changes.

`assemble_command` is added alongside `_assemble_command` rather than replacing it: that method reads `self.num_channels`, which raises before setup on `STARBackend`, so commands that carry no list parameter can be assembled before setup today and moving it would break that.

Tests: fourteen cases, each covering a branch with no coverage today.

Note on something carried over unchanged: the `datetime` branch formats with `%h`, the abbreviated month name, where it means `%H`, the hour - so a datetime parameter assembles as `2026-08-24 Aug:30`. It is dead in both copies, which is how it survived. Fixing it is a behaviour change to an uncovered branch and wants its own PR.

### DIFF
--- a/pylabrobot/hamilton/protocol/text/framing.py
+++ b/pylabrobot/hamilton/protocol/text/framing.py
@@ -11,7 +11,7 @@ either. What a given value *means* belongs with the machine that reports it.
 
 import datetime
 import re
-from typing import Dict, List, Optional, Sequence, TypeVar
+from typing import Any, Dict, List, Optional, Sequence, TypeVar
 
 T = TypeVar("T")
 
@@ -218,8 +218,6 @@ def assemble_command(
   module: str,
   command: str,
   id_: Optional[int] = None,
-  tip_pattern: Optional[List[bool]] = None,
-  num_channels: Optional[int] = None,
   **kwargs,
 ) -> str:
   """Assemble a firmware command.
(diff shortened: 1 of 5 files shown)


pylabrobot/hamilton/protocol/text/framing.py
```diff
@@ -168,3 +174,170 @@ def parse_firmware_version_date(fw_version: str) -> datetime.date:
...
+  if arg_index < len(val):
+    raise ValueError(f"Too many values for tip pattern {tip_pattern}: {val}")
+  return result
+
+
+def assemble_command(
+  module: str,
+  command: str,
+  id_: Optional[int] = None,
+  tip_pattern: Optional[List[bool]] = None,
```
rickwierenga: since this has `tip_pattern`, maybe we can split this into a separate channel aware version and keep `assemble_command` as the name/function for a (future) command that is agnostic, since other devices will also use this

rickwierenga (approved):

pylabrobot/hamilton/protocol/text/framing.py
BioCam: Good catch - done in bbd35da5d.

`assemble_command` now takes a module, a command, an optional id and named parameters, and knows nothing about channels. `assemble_channel_command` encodes each list parameter against the channel pattern and delegates to it, with `encode_channel_list` holding that encoding on its own.

The `RuntimeError` for a missing `num_channels` went with it - it only existed because the argument was optional on a function that sometimes needed it. On the channel version it is required, so the case cannot arise.

Verified unchanged: every command the test suite assembles is byte-identical to before, and 50000 randomised commands through the original method and the new pair agree exactly.

PR #1225, opened 2026-08-28
Fix visualizer server startup failures

### LINKED ISSUE #1224: Visualizer setup can hang indefinitely when server socket creation fails
Visualizer.setup() treats every OSError while starting its WebSocket or HTTP server as a port collision. It increments the port and retries forever.

Meanwhile, the caller waits using a blocking loop:

while lock.locked():
  time.sleep(0.001)

If socket creation fails for another reason—such as PermissionError—the exception is swallowed, the startup thread retries indefinitely, and the calling event loop hangs. The visualizer test currently relies on pytest-timeout to terminate
this condition.

Expected behavior:

- Retry only address-in-use errors.
- Propagate other server-thread exceptions to the caller.
- Bound the number or duration of port retries.
- Replace blocking lock polling with awaitable, timeout-aware synchronization.

Fixing this would also allow removal of the project’s sole pytest-timeout dependency.

Aming9303:
## What changed

- retry server startup only for `EADDRINUSE`, with a 100-port limit
- propagate other WebSocket and HTTP startup exceptions to `setup()`
- replace blocking lock polling with timeout-aware awaitable futures
- stop the WebSocket server even before a browser connects and clean up test servers reliably
- support port `0` so tests can request OS-assigned ports
- remove the no-longer-needed `pytest-timeout` dependency

## Tests

- `python -m pytest pylabrobot/visualizer/visualizer_tests.py -q` (26 passed)
- `python -m ruff check pylabrobot/visualizer/visualizer.py pylabrobot/visualizer/visualizer_tests.py`
- `python -m ruff format --check pylabrobot/visualizer/visualizer.py pylabrobot/visualizer/visualizer_tests.py`
- `python -m mypy pylabrobot/visualizer/visualizer.py`

Resolves #1224

### DIFF
--- a/pylabrobot/visualizer/visualizer.py
+++ b/pylabrobot/visualizer/visualizer.py
@@ -511,7 +511,7 @@ class Visualizer:
             self._socket_handler, self.host, self.ws_port
           ) as server:
             if self.ws_port == 0:
-              self.ws_port = server.sockets[0].getsockname()[1]
+              self.ws_port = next(iter(server.sockets)).getsockname()[1]
             print(f"Websocket server started at http://{self.host}:{self.ws_port}")
             startup.set_result(None)
             await self.stop_
--- a/pylabrobot/visualizer/visualizer.py
+++ b/pylabrobot/visualizer/visualizer.py
@@ -1,4 +1,6 @@
 import asyncio
+import concurrent.futures
+import errno
 import functools
 import http.server
 import inspect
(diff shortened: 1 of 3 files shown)


rickwierenga:
thanks for the pr!

please green the ci

Aming9303:
Fixed the Python 3.9 mypy failure by treating server.sockets as an iterable rather than indexing it. The focused type check now passes, along with all 26 visualizer tests, Ruff, and format checks.

rickwierenga:
thanks!

PR #1226, opened 2026-08-28
Fix well index decoding in Imager.capture for Well objects

hinderling:
Plate items are ordered column-major (A1, B1, ..., H1, A2, ...), but `capture()` decoded a well's index with `divmod(idx, num_items_x)`, which mis-addresses wells: `capture(well=plate.get_well("B2"))` images `A10` on a 96-well plate, while the `equivalent capture(well=(1, 1))` images B2.

Fix: Use the `get_child_row`/`get_child_column` accessors instead of hand-decoding the index, so the resolution no longer depends on item ordering at all. Adds a regression test.


rickwierenga:
thanks

PR #1228, opened 2026-08-31
`Resource`: fix a name when the resource is created, and refuse to change it later

BioCam:
A resource's name is its identifier: it is unique across the tree, it is what `get_resource` looks a resource up by, and it is the key `serialize_all_state` writes state under. Until now it could still be reassigned after the resource existed, and anything already named after it was left behind.

## The problem

A well is named three times before it is usable - `well_0_0` when it is constructed, `well_A1` by `create_ordered_items_2d`, and finally `plate_well_A1` by `ItemizedResource.__init__`, which prefixes every item it is handed with its own name. The last two are renames, and the last one writes the plate's name into all 96 of its wells.

Because the plate itself can still be renamed, those wells can be left describing a plate that no longer exists:

```python
plate = Cor_96_wellplate_360ul_Fb(name="plate")   # wells: plate_well_A1 ...
plate.name = "sample_plate"                       # wells: plate_well_A1 ...
```

Nothing raises, but `get_resource("sample_plate_well_A1")` finds nothing and the plate's state is serialized under keys that no longer correspond to anything in the tree.

## This PR

- the `name` setter now refuses, raising `AttributeError` rather than `RuntimeError` - that is what Python raises for a read-only property, so `setattr` and anything catching attribute errors behave as they would for any other immutable attribute.
- `create_equally_spaced_2d` takes a `name_for` callable, so each item is constructed with the name it will keep rather than renamed afterwards.
- `create_ordered_items_2d` takes `name_prefix`, the name of the resource that will hold the items, and passes it down. This matches `create_homogeneous_resources`, which already takes a `name_prefix` and already constructs each site with its final name. Carriers have named their children this way all along; this brings the plate side onto the same convention. The two differ only in separator - carriers join with `-`, items with `_`, which is what `ItemizedResource` used and what keeps every name unchanged.
- `ItemizedResource.__init__` no longer renames what it is given.
- the 167 call sites pass `name_prefix`: 155 pass `name_prefix=name`, the name their holder is being created with, and 12 in tests pass a literal. Four sites that build items by hand - two Opentrons loaders, the STAR teaching rack, and one test - name them after their holder directly.
- `named()` writes the field rather than going through the setter, since the copy it returns did not exist a moment ago and belongs to no tree; it is being built, not renamed.

Behaviour: names are unchanged, verified byte for byte across 342 resources on a loaded STARlet deck, and `named()` returns what it always did. The only change a caller can see is that assigning to `name` now raises.

### DIFF
--- a/pylabrobot/resources/resource.py
+++ b/pylabrobot/resources/resource.py
@@ -833,7 +833,8 @@ class Resource(SerializableMixin):
   def named(self, name: str) -> Self:
     """Return a copy of this resource with the given name."""
     new_resource = self.copy()
-    new_resource.name = name
+    # The copy is new and in no tree, so this finishes building it rather than renaming anything.
+    new_resource._name = name
     return new_resource
 
   def center(self, x: bool = True, y: bool = True, z: bool = False) -> Coordinate:
--- a/pylabrobot/resources/resource.py
+++ b/pylabrobot/resources/resource.py
@@ -235,14 +235,18 @@ class Resource(SerializableMixin):
 
   @name.setter
   def name(self, name: str):
-    """Set the name of this resource.
+    """Refuse the change: a resource's name is how it is identified.
 
-    Will raise a `RuntimeError` if the resource is assigned to another resource.
+    Raises:
+      AttributeError: Always. The name is the resource's identifier - it is unique across the tree,
+        it is what `get_resource` finds a resource by, and it is the key state is serialized under.
+        Anything built from it, such as the wells of a plate, takes it at construction, so a name
+        that changed afterwards would leave those disagreeing with it.
     """
-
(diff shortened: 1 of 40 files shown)


pylabrobot/resources/resource.py
```diff
@@ -829,7 +833,8 @@ def at(self, location: Coordinate) -> Self:
   def named(self, name: str) -> Self:
     """Return a copy of this resource with the given name."""
     new_resource = self.copy()
-    new_resource.name = name
+    # The copy is new and in no tree, so this finishes building it rather than renaming anything.
+    new_resource._name = name
```
rickwierenga, line 837: this is also good

rickwierenga (approved):

PR #1232, opened 2026-09-01
Add model grouping to device registry

rickwierenga:
## Summary

- add validated model objects with per-model support statuses to device registry entries
- render searchable model names across the first three table columns, with status badges aligned beneath device support
- let each model-bearing device row reveal its models through a compact chevron, with a bulk Show models toggle
- automatically reveal matching models when search narrows to 10 or fewer device families and count searchable models in the prompt
- consolidate the Masterflex model note and the Thermoshake, ALPS, Cytomat, PreciseFlex, STAR/STARlet, Ultraseal, and BioShake families
- limit the Cytation 1 capability listing to microscopy
- update affected guide cards and document the registry schema and model browsing behavior

## Testing

- `.venv/bin/python -m pytest docs/_exts/plr_devices/registry_tests.py` (29 passed)
- `.venv/bin/python -m ruff check docs/_exts/plr_devices/data.py docs/_exts/plr_devices/html.py docs/_exts/plr_devices/registry_tests.py`
- warning-as-error Sphinx builds of the machine table and all affected guide pages

### DIFF
--- a/docs/_exts/plr_devices/data.py
+++ b/docs/_exts/plr_devices/data.py
@@ -191,8 +191,24 @@ def _validate(device: Any, index: int, seen_ids: Dict[str, int], path: Path) -> 
     )
 
   models = device.get("models", [])
-  if not isinstance(models, list) or not all(isinstance(model, str) for model in models):
-    raise DeviceRegistryError(f"{where} ({device_id}): models must be a list of strings")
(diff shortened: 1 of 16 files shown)


rickwierenga:
scaling the device registry since we have too many different models now ...

PR #1237, opened 2026-09-03
PreciseFlex: split setup() into connect(), initialize() and disconnect()

miikee:
`setup()` did four things behind one name: open the socket, agree the response
mode, raise high power and attach, then home. A caller that wants to read a
position, or to reconnect after a controller restart, had no way to ask for
part of that.

- `connect()` opens the link and sets the response mode.
- `initialize()` raises power, attaches, leaves freedrive, and reads the
  controller's configuration.
- `disconnect()` detaches, drops power and closes the link.

Neither `connect` nor `initialize` moves the arm; `home()` is still the only
verb that sweeps it. `setup()` calls all of them in order, so existing callers
are unaffected.

Configuration discovery moves into `_discover_configuration()` and stays
best-effort. Because it can fail, `has_configuration` says whether the arm
actually read its own limits, which a caller that would rather adapt than be
raised at can now check.

**One ordering change worth flagging:** `stop_freedrive_mode` and configuration
discovery now run before `home()` rather than after. Freedrive off before homing
is the safer order, but it is a behaviour change, not a pure refactor.
`_handle_out_of_range_axes` still runs after `home()`, as before.

11 new tests cover the split, including that bring-up sends no motion command.

Draft: first of five on this file. The rest are ready and will follow as this
lands, since a PR base has to be a branch in this repo.

### DIFF
--- a/pylabrobot/brooks/precise_flex/precise_flex.py
+++ b/pylabrobot/brooks/precise_flex/precise_flex.py
@@ -277,25 +277,39 @@ class PreciseFlex:
     },
   )
   async def setup(self, skip_home: bool = False):
-    """Initialize the PreciseFlex driver.
-
-    Opens the socket connection, sets response mode to PC, powers on the
-    robot, attaches it, and (optionally) homes it.
+    """Bring the arm fully up: link, control, and (unless skipped) home.
 
     Args:
       skip_home: If True, skip the homing step during setup.
     """
-    await self.io.setup()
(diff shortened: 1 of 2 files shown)


miikee:
Superseded by #1239, which carries this commit plus the other four PreciseFlex changes as one PR. Verified: this branch's tip is an ancestor of #1239's head, so nothing is dropped. Closing to keep one PR per subsystem rather than a chain.

PR #1239, opened 2026-09-03
PreciseFlex: lifecycle verbs, a motion barrier, gripper axis limits, single-axis moves and a scoped speed

miikee:
Five changes to the PF400 driver, found running one on a bench. They are five
commits on one file, in order, and each stands on its own if you would rather
take them separately.

**1. `setup()` splits into `connect()` / `initialize()` / `disconnect()`.** It
did four things behind one name. A caller that wants to read a position, or to
reconnect after a controller restart, had no way to ask for part of it. Neither
`connect` nor `initialize` moves the arm; `home()` is still the only verb that
sweeps it, and `setup()` calls all of them in order so existing callers are
unaffected. Configuration discovery moves into `_discover_configuration()` and
stays best-effort, so `has_configuration` now says whether the arm actually read
its own limits.

**2. The gripper and the rail wait for the arm to stop.** `moveJ` returns when
the controller accepts it, not when the arm arrives, and the connection stays
free during travel. Joint and Cartesian moves already wait behind
`_wait_for_eom` via `_guarded_move_j`; `move_gripper`,
`move_gripper_joint_position` and `move_rail` did not. A grip issued after an
approach closes the jaws on the way down onto the plate. This is the one I would
take first.

**3. Gripper targets stay inside the axis, and widths anchor on the calibration
pair.** Commanding past the end of the gripper axis leaves a standing position
error the controller reports as an overheating motor (-3104), after which it
refuses every move until the arm is homed. It happened twice in one session:
opening to the advertised maximum of 134.0 landed the axis at 134.062, because
the servo overshoots. Separately, `setup()` copied the discovered axis limits
straight into `min_gripper_width` / `max_gripper_width`, which are millimetres,
so every width silently changed meaning after discovery.

**4. `move_one_axis`, `move_one_axis_relative`, `move_to_safe`.** Moving one
axis meant building a whole joint pose. The relative form offsets from the pose
read inside the guarded move, so a retry after recovery offsets from where the
axis actually ended up. `move_to_safe` runs the controller's own `movetosafe`,
which was buried in `park()`'s fallback branch and unreachable on its own.
`_set_grasp_data` becomes public, since a caller composing its own pick has to
set the grip before it.

**5. `at_speed`.** `move_to_location`'s `speed_pct` writes the profile speed and
never puts it back, so a slow approach left the arm slow for everything after
it. `at_speed` reads the current speed, sets the new one and restores in a
`finally`, so a fault mid-move cannot strand the arm slow.

Tests: 82 in `precise_flex_tests.py`, all passing, plus the rest of the suite.

Two things I would rather flag than bury. `_GRIPPER_LIMIT_HEADROOM = 0.5` is a
number measured on our gripper; happy to make it a constructor argument if you
would rather not have a bare module constant. And the settle barrier in (2) has
no opt-out, so pre-positioning the jaws during travel is no longer possible --
say the word and I will add a `wait: bool = True`.

### DIFF
--- a/pylabrobot/brooks/precise_flex/config.py
+++ b/pylabrobot/brooks/precise_flex/config.py
@@ -82,11 +82,6 @@ class PreciseFlexConfiguration:
 
   @property
   def gripper_width_range(self) -> tuple[float, float]:
-    """Travel limits of the gripper axis, in the controller's own units.
-
-    Not millimetres: a jaw width reaches these through
-    ``PreciseFlex.closed_gripper_position``, so callers must convert.
-    """
     return self.soft_limits[Axis.GRIPPER]
 
   @property
--- a/pylabrobot/brooks/precise_flex/config.py
+++ b/pylabrobot/brooks/precise_flex/config.py
@@ -81,7 +81,12 @@ class PreciseFlexConfiguration:
   reach_class: Literal["standard", "extended", "unknown"] = "extended"
 
   @property
-  def gripper_width_range(self) -> tuple:
+  def gripper_width_range(self) -> tuple[float, float]:
+    """Travel limits of the gripper axis, in the controller's own units.
+
+    Not millimetres: a jaw width reaches these through
+    ``PreciseFlex.closed_gripper_position``, so callers must convert.
+    """
     return self.soft_limits[Axis.GRIPPER]
 
   @property
--- a/pylabrobot/brooks/precise_flex/precise_flex.py
+++ b/pylabrobot/brooks/precise_flex/precise_flex.py
@@ -5,7 +5,18 @@ import dataclasses
 import logging
 import time
 import warnings
-from typing import Callable, ClassVar, Dict, List, Literal, NamedTuple, Optional, Sequence
+from contextlib import asynccontextmanager
+from typing import (
+  AsyncIterator,
+  Callable,
+  ClassVar,
+  Dict,
+  List,
+  Literal,
+  NamedTuple,
+  Optional,
+  Sequence,
+)
 
 from pylabrobot.brooks.precise_flex import kinematics
 from pylabrobot.brooks.precise_flex.config import Axis, PreciseFlexConfiguration
@@ -1211,6 +1222,34 @@ class PreciseFlex:
     """Get the current speed percentage of the arm's movement."""
     return await self.request_profile_speed(self.profile_index)
 
+  @asynccontextmanager
+  async def at_speed(self, speed_pct: Optional[float]) -> AsyncIterator[None]:
+    """Run a move at its own speed, then put the profile speed back.
+
+    The restore belongs here rather than with the caller: a fault between the move
+    and the restore would otherwise leave the arm slow for everything after it.
+    """
+    if speed_pct is None:
+      yield
+      return
+    prior = await self._request_speed()
+    await self._set_speed(speed_pct)
+    try:
+      yield
+    finally:
+      try:
+        await self._set_speed(prior)
+      except Exception:
+        # Raising here replaces whatever the move was already failing with, so name the
+        # state plainly: the arm is still at the move's speed and nothing else will reset it.
+        logger.error(
+          "[PreciseFlex %s] could not restore profile speed to %s; the arm is still at %s",
+          self.io._host,
+          prior,
+          speed_pct,
+        )
+        raise
+
   # -- brakes, torque & freedrive -----------------------------------------------------------
 
   async def release_brake(self, axis: int) -> None:
@@ -2107,7 +2146,8 @@ class PreciseFlex:
     Args:
       position: Target joint pose. Omitted axes keep their live values.
       speed_pct: Movement speed override as a percentage (0-100). If None, uses the current
-        speed setting.
+        speed setting. This STICKS: it is not restored afterwards. Wrap a call in
+        ``at_speed`` instead to scope a speed to one move.
     """
     if speed_pct is not None:
       await self._set_speed(speed_pct)
@@ -2697,9 +2737,7 @@ class PreciseFlex:
       direction,
       resource_width,
     )
-    if rail_position is not None:
-      await self.move_rail(rail_position)
-    elif self._has_rail:
+    if rail_position is None and self._has_rail:
       raise ValueError(
         "rail_position must be specified for pick_up_at_location when using a rail-equipped arm."
       )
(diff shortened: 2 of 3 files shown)


pylabrobot/brooks/precise_flex/config.py
```diff
@@ -81,7 +81,12 @@ class PreciseFlexConfiguration:
   reach_class: Literal["standard", "extended", "unknown"] = "extended"
 
   @property
-  def gripper_width_range(self) -> tuple:
+  def gripper_width_range(self) -> tuple[float, float]:
+    """Travel limits of the gripper axis, in the controller's own units.
+
+    Not millimetres: a jaw width reaches these through
+    ``PreciseFlex.closed_gripper_position``, so callers must convert.
+    """
```
rickwierenga: why do we not convert it in here so every python function is in standard PLR units? we should only convert to/from machine units in the functions that directly interact with hardware, users should not have to care

pylabrobot/brooks/precise_flex/precise_flex.py
```diff
@@ -742,7 +789,7 @@ async def _move_j(self, profile_index: int, joint_coords: JointPose) -> None:
       )
     await self.send_command(f"moveJ {profile_index} {angles_str}")
 
-  async def _move_one_axis(self, axis: Axis, position: float) -> None:
+  async def _recover_axis(self, axis: Axis, position: float) -> None:
```
rickwierenga, line 786: why is this changed?

pylabrobot/brooks/precise_flex/precise_flex.py
```diff
@@ -1338,12 +1413,15 @@ async def _request_grasp_data(self) -> tuple[float, float, float]:
       raise PreciseFlexError(-1, "Unexpected response format from GraspData command.")
     return (float(parts[0]), float(parts[1]), float(parts[2]))
 
-  async def _set_grasp_data(
+  async def set_grasp_data(
     self, plate_width: float, finger_speed_pct: float, grasp_force: float
   ) -> None:
```
rickwierenga, line 1412: this should stay private because it's always used in a `pick` command. we should minimize the state changes on the robot

pylabrobot/brooks/precise_flex/precise_flex.py
```diff
@@ -1621,6 +1699,15 @@ def configuration(self) -> "PreciseFlexConfiguration":
...
 
+  @property
+  def has_configuration(self) -> bool:
+    """Whether the controller's configuration was actually read.
+
+    Discovery is best-effort, so an arm can finish setup and still not know its own
+    limits. A caller that would rather adapt than be raised at asks this first.
+    """
+    return self._configuration is not None
+
```
rickwierenga: isnt this required for operation? ie if we cant load the config, we should not allow initialization to complete

pylabrobot/brooks/precise_flex/precise_flex.py
```diff
@@ -2056,12 +2146,55 @@ async def move_to_joint_position(
...
+    recovered position, which is what a relative move should mean.
+
+    Args:
+      axis: The axis to move.
+      distance: Signed offset to apply to that axis, in the axis's own units.
+      speed_pct: Movement speed override as a percentage (0-100). If None, uses the
+        current speed setting.
+    """
+    if speed_pct is not None:
+      await self._set_speed(speed_pct)
```
rickwierenga, line 2180: what is the justification for having relative commands? they are almost always annoying

pylabrobot/brooks/precise_flex/precise_flex.py
```diff
@@ -2355,16 +2491,21 @@ async def move_gripper(
...
-      and self._gripper_soft_max is not None
-      and not (self._gripper_soft_min <= units <= self._gripper_soft_max)
-    ):
-      raise ValueError(
-        f"gripper width {width} mm maps to firmware units {units:.1f}, outside the gripper "
-        f"axis range [{self._gripper_soft_min}, {self._gripper_soft_max}] - check "
-        f"closed_gripper_position (currently {self.closed_gripper_position})."
-      )
+    if self._gripper_soft_min is not None and self._gripper_soft_max is not None:
+      # An advertised end converts back through a subtract and an add, so it can land a
```
rickwierenga: I think if we dont have the soft min and max, we should not allow the code below to proceed

pylabrobot/brooks/precise_flex/precise_flex.py
```diff
@@ -2389,14 +2530,52 @@ async def move_gripper_joint_position(
...
       await self.send_command("gripper 1")
 
+  def _within_gripper_limits(self, units: float) -> float:
+    """A gripper target held a little inside the axis' soft limits.
+
+    Both ends are adopted together off one discovered tuple, so the arm has either
+    read its limits or read neither, and before setup there is nothing to hold to.
+    """
+    if self._gripper_soft_min is None or self._gripper_soft_max is None:
+      return units
```
rickwierenga: why is this valid?

pylabrobot/brooks/precise_flex/config.py
miikee: Agreed.  If I can find the conversion, I'll add it in here.

miikee: I think the comment is wrong.  

I think the name is confusing though.  This isn't a width, its 2 positions for soft limits.

pylabrobot/brooks/precise_flex/precise_flex.py
miikee: I moved away from using the `pick` command to maintain more control over the approach.  So I compose a sequence of motions to perform the pick and use `set_grasp_data` to set the gripper grasp.

To avoid state change, we could have something do the motion such as: `grasp(position, grasp_force, finger_speed_pct)`  

but then this is duplicating the `move_gripper(self, width: float, force_sensing: bool = False)` function

pylabrobot/brooks/precise_flex/precise_flex.py
miikee: This was actually one of my most used commands lol. I often need to lift the base ~10mm to clear something before I can move it back to a safe position.

pylabrobot/brooks/precise_flex/precise_flex.py
miikee: Sounds good - I'll fix this

pylabrobot/brooks/precise_flex/precise_flex.py
miikee: This should probably just refuse

PR #1240, opened 2026-09-03
OT-2 (legacy): fix the position read, and bound every request instead of blocking forever

miikee:
Two commits on the legacy OT-2 backend. The second builds on the first, so they
are one PR.

**1. `_current_channel_position` is broken today.** It calls
`ot_api.lh.save_position`, which `ot_api` does not define, so every caller
raises. `move_channel_x`, `move_channel_y` and `move_channel_z` on this backend
cannot work at all right now: they always fail with `RuntimeError: Failed to
query current pipette position`. It now enqueues `savePosition` itself and polls
the command the way `ot_api`'s own wrapper does, awaiting rather than sleeping.

With the read working, `move_channel_to` moves a channel to an absolute position
and holds whichever axes are left out. Chaining the per-axis calls was the only
way to reach an arbitrary point before, and each of those descends separately,
so a three-axis move could clip labware between the steps.

**2. Every request is bounded and off the event loop.** `ot_api` reaches the
robot through `urlopen` and passes it no timeout, so a request the robot never
answers blocks the process for good. All 24 `ot_api` calls are issued from an
`async def` and ran inline, so that block also froze every other device
coordinating behind it.

Every call now goes through one `_request` helper: a daemon thread keeps the
loop free, `asyncio.wait_for` ends the wait at a real deadline, and a lock keeps
commands one at a time the way the blocking calls used to for free. The thread
is deliberately not the loop's shared executor, because `asyncio.run()` joins
that executor on the way out.

Three budgets on the constructor, named to match the Flex backend:
`request_timeout`, `command_timeout`, `status_poll_interval`.

Robot commands are enqueued and polled here rather than through `ot_api`'s
`@command` decorator, whose 30s ceiling no caller can raise and whose poll loop
has no sleep in it at all. A command that times out is contained rather than
abandoned: the run is stopped so what is still queued cannot execute. Retrying
an aspirate used to make the pipette aspirate twice from one well.

**Not verified on an OT-2** -- I do not have one. The `savePosition` command and
its response shape are the same ones a Flex uses, where they are hardware
verified. I diffed all seven command payloads against the installed `ot_api`
and they are byte-identical, and the tests fake the robot at its command queue
so they pin the params a real robot would receive.

Worth your call: this moves knowledge of the Opentrons HTTP command payloads
into PyLabRobot rather than leaving it in `ot_api`. I think that is right, since
the payloads are protocol facts and PLR was already pinned to `ot_api`'s bugs,
but it is a coupling decision rather than a detail, so say if you would rather
it stayed out.

413 tests passing.

### DIFF
--- a/pylabrobot/legacy/liquid_handling/backends/opentrons_backend.py
+++ b/pylabrobot/legacy/liquid_handling/backends/opentrons_backend.py
@@ -1,9 +1,10 @@
 import asyncio
 import inspect
 import logging
+import threading
 import time
 import uuid
-from typing import Any, Dict, List, Optional, Tuple, Union, cast
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union, cast
 
 from pylabrobot import utils
 from pylabrobot.io import LOG_LEVEL_IO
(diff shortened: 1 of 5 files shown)


rickwierenga:
thanks for the PR!

does this need to change legacy or can we do it in the current implementation of OT2 on the main branch? that version has many benefits and improvements. I recommend using it if possible. If you would rather keep this on legacy lmk and we can also do that

PR #1247, opened 2026-09-09
Fix 3D rotation composition with quaternions

### LINKED ISSUE #1165: `Resource.rotate` implementation is wrong in general
Currently, `Resource.rotate` reads
```python
  def rotate(self, x: float = 0, y: float = 0, z: float = 0):
    """Rotate counter-clockwise by the given number of degrees."""

    self.rotation.x = (self.rotation.x + x) % 360
    self.rotation.y = (self.rotation.y + y) % 360
    self.rotation.z = (self.rotation.z + z) % 360

    # Rotation is part of the resource's state; notify subscribers (e.g. the
    # Visualizer) so they can re-render.
    self._state_updated()
```

However, `rotation` represents rotations as Euler angles (as can be inferred from `Rotation.get_rotation_matrix`). The above implementation **siltently gives wrong results**. A trivial example is rotating a resource which currently is rotated 90 degrees around z about the x axis. Because Euler angles in the roll-pitch-yaw convention rotate about x first, simply adding to the roll value affectively acts as a rotation around y instead of the requested x.

The correct composition behaviour of Euler angles unfortunately is very nontrivial; requiring forward and inverse trigonometrics. That also leads to certain rotations have an ill-defined Euler angle representation ("gimbal lockup") where the inverse trigonometrics diverge. The standard solution in robotics is to use quaternions to represent rotations, where composition is compartively simple, and are free of divergences. Thanks to the `Rotation` abstraction in PLR, such a refactor could be fairly local. The biggest issue are the places where the abstractions are leaked, and code directly looks into `Rotation`. The one place I have in mind (`LiquidHandler.drop_resource`) also silently ignores all but yaw, so that one benefits from a refactor anyway (should probably raise `ValueError` on significantly off-axis rotations).

fly1d:
Fixes #1165.

## Summary

- Store a normalized quaternion inside `Rotation` while keeping the public `x`, `y`, and `z` accessors and serialized shape.
- Compose `Rotation` values with the Hamilton product instead of adding Euler components.
- Apply `Resource.rotate()` increments in the parent-coordinate frame while preserving the existing `Rotation` instance and `[0, 360)` component normalization.
- Cover non-commuting rotations, parent-child composition, component assignment, gimbal-lock representations, and serialization round trips.

## Why

Adding Euler components does not compose general 3D rotations. For example, applying an x-axis increment after a 90-degree z rotation produces a different matrix when the angles are added. Quaternion multiplication preserves the requested order without introducing a matrix-to-Euler conversion into each composition.

Euler angles remain part of the public and serialized API. Equivalent Euler representations are therefore selected close to the previous component-wise values, with an explicit singular case at `pitch = +/-90` degrees.

## Verification

- `.venv/bin/python -m pytest -q`: 2983 passed, 36 skipped, 759 subtests passed
- `BIN=.venv/bin/ make lint`
- `BIN=.venv/bin/ make format-check`
- `BIN=.venv/bin/ make typecheck`: 808 source files checked
- 10,000 deterministic composition and serialization checks, including near-singular angles: maximum matrix errors `1.110e-15` and `2.753e-12`

No hardware was connected or driven.

AI assistance: OpenAI Codex assisted with implementation and test validation.

### DIFF
--- a/pylabrobot/resources/rotation.py
+++ b/pylabrobot/resources/rotation.py
@@ -1,70 +1,192 @@
 import math
 
 from pylabrobot.serializer import SerializableMixin
-from pylabrobot.utils.linalg import matrix_multiply_3x3
+
+# Quaternions use (w, x, y, z); Euler angles retain the Rz * Ry * Rx convention.
+_Quaternion = tuple[float, float, float, float]
+
+
+def _normalize_quaternion(quaternion: _Quaternion) -> _Quaternion:
+  """Return a unit-length quaternion."""
+  norm = math.sqrt(sum(component * component for component in quaternion))
+  w, x, y, z = quaternion
+  return w / norm, x / norm, y / norm, z / norm
+
+
+def _multiply_quaternions(left: _Quaternion, right: _Quaternion) -> _Quaternion:
+  """Compose two quaternions using the Hamilton product."""
+  left_w, left_x, left_y, left_z = left
+  right_w, right_x, right_y, right_z = right
+  return _normalize_quaternion(
+    (
+      left_w * right_w - left_x * right_x - left_y * right_y - left_z * right_z,
+      left_w * right_x + left_x * right_w + left_y * right_z - left_z * right_y,
+      left_w * right_y - left_x * right_z + left_y * right_w + left_z * right_x,
+      left_w * right_z + left_x * right_y - left_y * right_x + left_z * right_w,
+    )
+  )
+
+
+def _quaternion_from_euler(x: float, y: float, z: float) -> _Quaternion:
+  """Convert roll, pitch, and yaw in degrees to a quaternion."""
+  half_roll = math.radians(x) / 2
+  half_pitch = math.radians(y) / 2
+  half_yaw = math.radians(z) / 2
+  cos_roll, sin_roll = math.cos(half_roll), math.sin(half_roll)
+  cos_pitch, sin_pitch = math.cos(half_pitch), math.sin(half_pitch)
+  cos_yaw, sin_yaw = math.cos(half_yaw), math.sin(half_yaw)
+  return _normalize_quaternion(
+    (
+      cos_roll * cos_pitch * cos_yaw + sin_roll * sin_pitch * sin_yaw,
+      sin_roll * cos_pitch * cos_yaw - cos_roll * sin_pitch * sin_yaw,
+      cos_roll * sin_pitch * cos_yaw + sin_roll * cos_pitch * sin_yaw,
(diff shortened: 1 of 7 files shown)


BioCam:
@fly1d nice fix - this is the right way to do it. A heads-up on how it interacts with #1248, which I just merged.

This makes `Rotation.get_rotation_matrix` about **17x faster** (5.0 → 0.30 µs) but `Rotation.__add__` about **24x slower** (0.16 → 3.88 µs).
Before #1248 the slowdown outweighed the speed-up, because `get_absolute_location` called `__add__` a quadratic number of times, so the whole thing got slower.
Those calls are gone now, so against current main it comes out roughly neutral: 3-4% slower on an unrotated deck, no measurable difference on rotated chains.

The small slowdown is `.x`/`.y`/`.z` becoming properties (27 → 49 ns per read) - `get_absolute_location` reads all three once per chain level to decide whether a matrix is needed at all.
Comparing the stored quaternion against the identity instead would claw it back.

pylabrobot/resources/rotation.py
```diff
@@ -1,71 +1,186 @@
...
+    """Return the rotation as a 3x3 matrix."""
+    w, x, y, z = self._quaternion
+    return [
+      [1 - 2 * (y * y + z * z), 2 * (x * y - w * z), 2 * (x * z + w * y)],
+      [2 * (x * y + w * z), 1 - 2 * (x * x + z * z), 2 * (y * z - w * x)],
+      [2 * (x * z - w * y), 2 * (y * z + w * x), 1 - 2 * (x * x + y * y)],
     ]
-    # Combine rotations: The order of multiplication matters and defines the behavior significantly.
-    # This is a common order: Rz * Ry * Rx
-    return matrix_multiply_3x3(matrix_multiply_3x3(Rz, Ry), Rx)
```
BioCam, line 58: This is `matrix_multiply_3x3`'s last production caller - afterwards it is referenced only by its own test and the two test files in this PR.

Worth dropping it, or noting that it is kept as a test reference.

BioCam (changes requested):

pylabrobot/resources/rotation.py
fly1d: Kept `matrix_multiply_3x3` as an independent test reference and clarified that purpose in its docstring.

fly1d:
Thanks. I rebased onto current `main` and applied the suggested hot-path fix in `6330eb50`: `get_absolute_location` now compares the stored quaternion directly with the identity quaternion. The 88 focused tests, Ruff, formatting, and mypy on the changed modules pass.

PR #1249, opened 2026-09-09
`Resource`: turn about a joint, and the links and end-effectors that need it

BioCam:
An articulated device has joints that are not at a resource's own corner, and nothing in PyLabRobot can express one. A carousel, a rotor, a hinge, every link of an arm - all pivot on a point somewhere on the part, and `Resource.rotate` has always turned about the left front bottom corner, so a link modelled today swings off its joint rather than on it.

- `rotate`, `rotate_to` and `rotated` take an optional `pivot_coordinate`: the point to turn about, in the resource's own frame. The resource is carried by however far the turn moved that point, so the point stays put and the resource swings on it. The offset is taken back through the parent's rotation, since `pivot_coordinate` is measured in the resource's frame and `location` in the parent's. A pivot on a resource with no location raises `NoLocationError`, because a pivot is held by moving `location` and there is nothing to move.
- `rotate_to` is new: it sets the rotation about each axis where `rotate` turns by an amount. Axes left as `None` keep the angle they had, and each is normalised to `[0, 360)`.
- `LinkBody` is one rigid member of a chain. It is an ordinary resource, so its origin is a corner, and it carries `proximal_joint` and `distal_joint` as coordinates within it. The link is the line between those two joints and nothing stores it; `length` is the distance, and is `None` on a member that ends the chain. Because the origin is a corner rather than a joint, a member turns about `proximal_joint`, which is what the pivot above is for. Material hangs off it as children with their own origins, which keeps a member's frame apart from the shape around it.
- `MechanicalGripper` is a `LinkBody` that ends the chain: nothing attaches past a tool, so it has no distal joint, and what sits at the far end of its span is its tool centre point. Body, fingers and pads are bolted to it, and how far apart the fingers stand is state rather than shape. It is sized to its body alone, because `jaw_width` moves the fingers and a box drawn around them would resize with the jaws. The jaws straddle the grip centre.

Behaviour: unchanged without a pivot. `pivot_coordinate` defaults to `None`, and both the matrix build and the offset are skipped without one, so `rotate` costs what it always did at any depth.

Scope against `main`: this is no longer additions-only to `resource.py`. `rotate` and `rotated` each gained the optional argument, so both signatures and both one-line docstrings changed and `rotated` forwards it - five upstream lines, none deleted outright. `rotate_to` and the two new classes are additions.

Tests: the new primitives carry their own - a chain folding on its joints, the link measured between its joints, the jaws standing symmetrically about the grip centre and refusing a width they cannot reach, and a pad sitting the same way on both fingers. `resource_tests` covers the pivot about every axis, under a rotated parent, and under an unlocated parent that a rotated ancestor carries, asserting both that the pivot point stands still and that the turn reached the angle it asked for. `TestAnAssembledArm` puts a member on a deck with a gripper on its far joint and pins absolute positions through two turns, reading the wrist off each side of the joint independently, since the two classes are otherwise only ever tested apart. Full suite passes, 3238 tests.

One pre-existing wart worth naming so it is not read as new here: `Resource.deserialize` resolves the subclass and then calls its constructor rather than dispatching to a custom `deserialize`, so a `MechanicalGripper` cannot be rebuilt through the generic path at the root of a tree. `TipSpot` and `NimbusDeck` behave the same way on `main`. Nested grippers round-trip correctly, since a parent deserializes children through `child_cls.deserialize`.

This also touches the same lines of `rotate` as #1247, so whichever lands second carries a small merge.


rickwierenga:
I merged #1247

rickwierenga (approved):

PR #1250, opened 2026-09-09
Add EventBus setup/stop coverage for VSpin and Access2

j0nch:
Modern VSpin and Access2 setup/stop calls were missing semantic EventBus lifecycles. This adds `centrifuge.setup` / `centrifuge.stop` to VSpin and public `Access2.setup()` / `Access2.stop()` wrappers emitting `centrifuge_loader.setup` / `centrifuge_loader.stop`.

Each operation reports the frontend device identity and emits `started` followed by `completed` or `failed`. Access2 delegates to its existing driver methods, which remain directly callable and undecorated to avoid duplicate events. Hardware behavior is preserved. These namespaces use device-only context rather than the legacy `machine.*` schema and its required `backend` field.

Tests cover device identity, lifecycle correlation, exception preservation, and identical Access2 setup/stop transcripts through the frontend and direct driver APIs. Documentation updates cover the operation reference, canonical schema registry, frontend coverage table, and preferred Access2 frontend usage.

Validation:
- 160 tests and 103 subtests passed across Agilent VSpin/Access2 and EventBus suites.
- All eight new lifecycle cases fail against the original source and pass with this patch.
- Ruff lint/format and type checks passed.
- Tested on a physical workcell; the new lifecycle events were confirmed to look correct.
- Documentation checks encountered existing stale API references; the no-API check also cannot resolve the excluded VSpin API label.

### DIFF
--- a/docs/contributor_guide/event-schemas.md
+++ b/docs/contributor_guide/event-schemas.md
@@ -161,6 +161,8 @@ These are state-transition records rather than semantic operation lifecycles.
 
 | Operation | Fields | Notes |
 | --- | --- | --- |
+| `centrifuge_loader.setup` | `device` | Modern `Access2.setup()` initializes and homes the loader through its driver. |
+| `centrifuge_loader.stop` | `device` | Modern `Access2.stop()` closes the loader transport. |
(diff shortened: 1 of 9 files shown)


rickwierenga:
@j0nch please fix type

PR #1271, opened 2026-09-17
Make `Tip` a `HeadTool` resource, build `StandingTipRack`, introduce Hamilton NTRs

BioCam:
A tip is not a resource on `main`. It lives only inside its spot's `TipTracker`, so nothing in the tree says where a tip is, the backends each re-derive its position from the spot, and a channel carrying a tip has no geometry below it. The CO-RE grip tool, which a channel picks up exactly like a tip, has no model at all. `StandingTipRack` is a stub, and the only nested tip rack, a `NestedTipRack` for the 50 µL NTR, has its tip spots where the instrument doesn't put them.

## This PR

**`Tip` is a `HeadTool` resource**
- `HeadTool(Resource)`: what a channel carries. It has `fitting_depth`, `collar_height`, `pick_up_location` (the centre of the top of the opening the channel enters) and `kind()`, which is equal for tools a backend can use interchangeably and keys the firmware tip-type tables.
- `Tip` is a `HeadTool`. `HamiltonTip` takes its diameter and `collar_height` from `tip_size`.
- `HamiltonCoreGripperTool` / `hamilton_core_gripper_tool` model the grip tool, with `total_length` = 30 mm to its grip line, as the firmware counts it.
- A tip is a child of its `TipSpot`: centred, with its pick-up location `collar_height` above the spot. A mounted tool is a child of its `TipMountingShaft`: centred, with its pick-up location `fitting_depth` above the shaft's end. `Resource.comparable_children()` keeps a carried tool out of equality, so a spot with a tip still equals one without.
- The STAR simulator answers tip presence and lowest point from the shafts.

**`StandingTipRack`**
- A tip rack that stands on its own rather than sinking into a holder, with an optional `stacking_z_height`. `NestedTipRack` is deprecated.
- `ResourceStack` nests standing tip racks by their `stacking_z_height`, as it does plates.
- `LiquidHandler` refuses to pick up tips from, or drop tips to, a rack with a lid or another rack on top of it (`TipRack._available`).

**Hamilton NTRs**
- Hamilton racks come from two bodies, and every catalogue rack is a body plus its tip:
  - `hamilton_96_tiprack_standard`: renamed from `hamilton_tiprack_standard`, with a deprecated alias. Its spots are the 7.2 mm holes, with the same centres.
  - `hamilton_96_tiprack_ntr`: an SLAS footprint with SLAS tip positions and spots at the collar seat.
- `hamilton_96_tiprack_10uL_NTR` and `_300uL_NTR` are new. `_50uL_NTR`'s spots move from 13.5 to 55 mm.
- `hamilton_tip_carrier_L5_ntr_a00` (182074) replaces `TIP_CAR_NTR_A00`, which remains as a deprecated alias, and keeps Hamilton's sites. `hamilton_mfx_module_tiprackholder_ntr` (191425) is new.

Behaviour:
- **Unchanged:** positions and firmware for embedded racks.
- **Changed:**
  - a serialized tip includes its size and `pick_up_location`;
  - racked tips appear in `get_all_children()`.
- **Firmware:** the Vantage discard now takes the tip end from NTR spots, as it does from embedded ones.

Tests:
- `test_nested_tip_racks_ground_truth`: for the 10, 50 and 300 µL NTRs, on both the MFX NTR4 module and the NTR carrier, the channel and 96-head pick-up and drop commands match what Hamilton's own software sends.
- `TipSpot` and shaft placement, grip tool geometry, `kind()`, stacked-rack nesting and the covered-rack guard.

Full suite passes (3410 passed, 4 skipped). `ruff format`, `ruff check --select I`, `ruff check` and `typos` are clean. `mypy` shows only the existing LI-COR test errors on `main`.

This touches `TipSpot.__init__` next to #1251, so whichever lands second carries a small merge.


BioCam:
Screenshot 2026-09-17 at 01 53 35

modelling physical reality

pylabrobot/legacy/liquid_handling/liquid_handler_tests.py
```diff
@@ -1116,22 +1117,22 @@ async def test_discard_tips(self):
...
-            tip=tips[1],
+            tip=tips[2],
             offset=offsets[2],
           ),
           Drop(
             self.deck.get_trash_area(),
-            tip=tips[0],
+            tip=tips[3],
             offset=offsets[3],
           ),
```
rickwierenga, line 1137: this seems wrong? it is dangerously changing the behavior of legacy

pylabrobot/resources/hamilton/core_gripper_tools.py
```diff
@@ -0,0 +1,88 @@
...
+      pick_up_location=pick_up_location,
+    )
+    self._total_length = total_length
+    # What the machine has to be told about this tool, in the same tip type table a tip goes in:
+    # a grip tool holds no liquid, but the table demands at least 1.0 uL, and the value plays no
+    # part in picking the tool up.
+    self.has_filter = False
+    self.maximal_volume = 1.0
+    self.tip_size = TipSize.UNDEFINED
+    self.pickup_method = TipPickupMethod.OUT_OF_RACK
```
rickwierenga, line 56: why does it have a volume?

pylabrobot/resources/resource.py
```diff
@@ -253,6 +253,10 @@ def name(self, name: str):
       "when it is created. Create the resource with the name you want instead."
     )
 
+  def comparable_children(self) -> List["Resource"]:
+    """The children compared in equality: all of them, unless a subclass holds some as state."""
+    return self.children
+
   def __eq__(self, other):
```
rickwierenga, line 260: this is ugly. why do we need it?

pylabrobot/resources/resource_stack.py
```diff
@@ -104,15 +105,22 @@ def _actual_resource_height(resource: Resource) -> float:
...
     if (
-      self.direction == "z"
-      and isinstance(upper, Plate)
+      isinstance(upper, Plate)
       and upper.stacking_z_height is not None
       and isinstance(lower, Plate)
       and lower.lid is None
+    ) or (
+      isinstance(upper, StandingTipRack)
+      and upper.stacking_z_height is not None
```
rickwierenga, line 122: In the future I would like a `NestedTipRackStack(ResourceStack)` subclass that has a counting feature, but that is for a follow up

pylabrobot/resources/tip.py
```diff
@@ -30,72 +32,60 @@ def __init__(
...
-      return NotImplemented
-
-    return (
-      self.has_filter == other.has_filter
-      and self.total_tip_length == other.total_tip_length
-      and self.nominal_volume == other.nominal_volume
-      and self.maximal_volume == other.maximal_volume
-      and self.fitting_depth == other.fitting_depth
-      and self._collar_height == other._collar_height
-    )
```
rickwierenga, line 91: why remove this?

pylabrobot/resources/head_tool.py
```diff
@@ -0,0 +1,124 @@
...
+  def serialize(self) -> dict:
+    """Serialize the tool, without its location, which its holder determines."""
+    data = {
+      **super().serialize(),
+      "fitting_depth": self.fitting_depth,
+      "collar_height": self._collar_height,
+      "pick_up_location": serialize(self.pick_up_location),
+    }
+    for held_by_the_holder in ("location", "parent_name"):
+      data.pop(held_by_the_holder, None)
```
rickwierenga, line 123: why is this?

pylabrobot/resources/head_tool.py
```diff
@@ -0,0 +1,124 @@
...
+    """Return collar_height, raising if it is None."""
+    if self._collar_height is None:
+      raise ValueError(f"collar_height is not defined for this tool: {self!r}")
+    return self._collar_height
+
+  @property
+  def has_collar_height(self) -> bool:
+    """Whether this tool states the height of its collar."""
+    return self._collar_height is not None
+
```
rickwierenga, line 109: why is this like this? why not just check .collar_height is None

pylabrobot/resources/head_tool.py
```diff
@@ -0,0 +1,124 @@
...
+        "once it is set."
+      )
+    self._name = name
+    self._is_named = True
+
+  @property
+  def is_named(self) -> bool:
+    """Whether this tool has a name."""
+    return self._is_named
+
```
rickwierenga, line 97: shouldn't resources always have names? what is the reason they dont have it here? locations where tips spawn like tip spots should probably have a counter for making unique names

pylabrobot/resources/tip_rack.py
```diff
@@ -198,6 +206,17 @@ def __repr__(self) -> str:
       f"size_y={self._size_y}, size_z={self._size_z}, location={self.location})"
     )
 
+  @property
+  def _available(self) -> bool:
+    """Whether nothing, a lid or another rack in its stack, sits on top of this rack."""
+    # A rack's spots are assigned when it is made; anything put on the rack comes after them.
```
rickwierenga, line 212: this name is a bit confusing. how about `_can_accept_op` or sth

pylabrobot/resources/tip_rack.py
```diff
@@ -106,11 +112,13 @@ def empty(self) -> None:
...
   def serialize(self) -> dict:
-    """Serialize the tip spot."""
-    return {
+    """Serialize the tip spot. Its tip is state, not a serialized child."""
+    data = {
       **super().serialize(),
       "prototype_tip": self.make_tip().serialize(),
     }
+    data.pop("children", None)
+    return data
```
rickwierenga, line 121: why exclude children?

pylabrobot/legacy/liquid_handling/liquid_handler_tests.py
BioCam: good catch stopping on this one - the order itself wasn't a behaviour change (on main all four tips compare equal, so `tips[3] … tips[0]` asserted nothing either way), but you're right that it has no business being in this PR.

what dragged it in was `Tip` losing its value `__eq__` when it became a resource. i went looking and nothing actually needs that - `kind()` (new in this PR: a head tool's serialized form with the names stripped, so two tips of one model are one kind) already answers every "same model?" question the backends ask. so `__eq__`/`__hash__` are back and the test hunk is reverted entirely - this file is now untouched apart from the stacked-racks test. tip equality deserves its own PR rather than holding this one up. full suite green.

happy with that?

pylabrobot/resources/n_channel_pipettes.py
```diff
@@ -126,16 +132,11 @@ def release_tip(self) -> Resource:
...
-    which is what has to clear the deck.
-
-    Returns:
-      The offset from the channel to the bottom of what it carries.
-    """
+    """Where the bottom of what this shaft carries is, relative to the shaft's end."""
     tip = self.tip
-    return Coordinate.zero() if tip is None else Coordinate(0.0, 0.0, -tip.get_absolute_size_z())
+    if tip is None:
+      return Coordinate.zero()
```
rickwierenga: should return None

rickwierenga:
I’ve narrowed this PR to CO-RE grippers and supporting shaft changes; could you open a separate PR for the rack work preserved on [pr-1271-racks](https://github.com/PyLabRobot/pylabrobot/tree/pr-1271-racks)?

PR #1273, opened 2026-09-17
fix(resources): serialize `Container.no_go_zones` so `copy()` and `deserialize()` work

### LINKED ISSUE #1272: `Container` with `no_go_zones` cannot be copied or deserialized
`copy()` and `Container.deserialize(c.serialize())` raise for any `Container` that has `no_go_zones`. That includes the Hamilton troughs from the resource library.

```python
from pylabrobot.resources import Container, Coordinate
from pylabrobot.resources.hamilton import hamilton_1_trough_200mL_Vb

trough = hamilton_1_trough_200mL_Vb(name="trough")
print("no_go_zones:", trough.no_go_zones)
try:
  trough.copy()
  print("copy(): ok")
except TypeError as e:
  print("copy(): TypeError:", e)

c = Container(name="c", size_x=10, size_y=10, size_z=10,
              no_go_zones=[(Coordinate(0, 4, 0), Coordinate(10, 6, 10))])
try:
  Container.deserialize(c.serialize())
  print("deserialize(serialize()): ok")
except TypeError as e:
  print("deserialize(serialize()): TypeError:", e)
```

Output on `main` (e9c1c73b) and on 0.2.2 from PyPI:

```
no_go_zones: [(Coordinate(x=0, y=60, z=8.0), Coordinate(x=37.0, y=61.7, z=60.0))]
copy(): TypeError: no_go_zones[0] must be a tuple of Coordinate instances, got <class 'dict'>, <class 'dict'>.
deserialize(serialize()): TypeError: no_go_zones[0] must be a tuple of Coordinate instances, got <class 'dict'>, <class 'dict'>.
```

`Container.serialize()` writes each zone as a Python tuple of two serialized coordinates:

```python
"no_go_zones": [(flb.serialize(), brt.serialize()) for flb, brt in self.no_go_zones],
```

`serializer.deserialize()` only recurses into lists and dicts, so the tuple is passed through as is and the two dicts inside never become `Coordinate`s again. It only works if the data goes through JSON in between (save to file and load), because JSON turns the tuples into lists. `copy()` deserializes the dict directly, so it fails.

Using `serialize(self.no_go_zones)` fixes it, since the serializer already turns tuples into lists. I'll open a PR.

haeganm:
Fixes #1272

**Problem**

`Container.serialize()` writes each no-go zone as a Python tuple of two serialized coordinates. `deserialize()` only recurses into lists and dicts, so the tuple passes through and the coordinates stay dicts. `copy()` and `Container.deserialize(c.serialize())` then raise `TypeError: no_go_zones[0] must be a tuple of Coordinate instances` for every container with no-go zones, including the Hamilton troughs. Loading from a JSON file works only because JSON turns the tuples into lists.

**Fix**

Use `serialize(self.no_go_zones)`, which emits lists. `_validate_no_go_zones` already converts each zone back to a tuple, so `no_go_zones` is the same after a round trip. Files saved through JSON already contain lists, so they load as before.

**Tests**

Updated the expected value in `test_no_go_zones_serialize` and added `test_no_go_zones_deserialize` and `test_no_go_zones_copy`. All three fail on `main` and pass here. `make lint`, `make format-check` and `mypy` are clean for the changed files.


rickwierenga:
thanks!

PR #1274, opened 2026-09-17
Add the raised CO-RE II and CO-RE I 96 tip racks

BioCam:
Adds `hamilton_96_tiprack_raised_core_ii` (Hamilton cat. no. 6608647) and `hamilton_96_tiprack_raised_core_i`.

Both are `EmbeddedTipRack`s that sink into a tip carrier site or an MFX tip rack holder, as a standard rack does, and differ from it in where they hold their tips:

| | standard rack | raised rack |
| --- | --- | --- |
| footprint | 122.4 x 82.6 | 122.0 x 78.5 |
| height | 20.0 | 25.5 (CO-RE II) |
| sinks into its holder | 6.0 | 9.5 |
| frame | 10 mm wall around the tips | none |
| tips held | 1.5 mm above the holder | 16.0 mm above it |

The raised rack is a thick body that does not flex, where a standard rack is a thin floor stiffened by its wall, and it holds its tips at its top face rather than down in wells. Each takes the tip it holds - 10, 50, 300 or 1000 uL - as `make_tip`, since one body carries any of them.

`_core_i` is the generation before `_core_ii`, holding its tips 3.5 mm lower. The two look alike, and using one definition for the other part drives the tips 3.5 mm into it, which is why the generation is in the name. Only the CO-RE II body has been measured.

Tests: tip spot positions for both on a `TIP_CAR_480BC_A00` on a `STARDeck`.


BioCam:
Screenshot 2026-09-17 at 21 29 09

Issue #1218, opened 2026-08-25
Add i.prep 2 device driver


jacobveon:
Add native PyLabRobot support for the Veon Scientific i.prep 2 liquid handler.

PR #1278, opened 2026-09-18
`pylabrobot.io.HTTP`: build the request lock where there is a loop to bind it to

jacobveon:
Part of #1218.

`HTTP.__init__` constructs an `asyncio.Lock`. A transport is built in ordinary synchronous code — a device object is made long before anything is set up — and an asyncio primitive binds to a loop when it is constructed. On Python 3.9 that means it looks for a loop and raises when there is none:

```
RuntimeError: There is no current event loop in thread 'MainThread'.
```

There is none as soon as anything has run and closed one, which is why a test suite runs into this and a script usually does not. Building an `HTTP` in a synchronous test that runs after any `IsolatedAsyncioTestCase` fails at construction. `pyproject.toml` has `requires-python = ">=3.9"` and CI runs a 3.9 job, so this is inside the supported range.

It goes unnoticed today only because `http_tests.py` happens to build its transports inside async tests.

## The change

The lock is built by `setup()`, alongside the `ThreadPoolExecutor` it orders work onto — the first point at which a loop is certainly running, and already where the other half of the transport's machinery comes from. Requests reach it through `_require_lock()`, so a request before setup raises the same "call setup() first" error the missing executor already raised, rather than failing on the absent lock.

Worth doing beyond 3.9: a lock built in one loop and used in another is wrong on every version, and a constructor is exactly where that happens.

## Testing

Two tests, both of which fail without the change on 3.9 and pass with it — verified by reverting only `http.py` and keeping the tests:

```
FAILED http_tests.py::HTTPConstructionTests::test_can_be_built_without_a_running_loop
FAILED http_tests.py::HTTPConstructionTests::test_a_request_before_setup_says_to_set_up
```

`pylabrobot/io/` passes on 3.9 and 3.10 (114 passed, 16 skipped on each); the full suite passes on 3.13. `make format-check`, `make lint` and `make typecheck` are clean.


rickwierenga (approved):

rickwierenga:
thanks!

